PR #3, opened 2021-10-12
Some comments

bifurcation:


draft-patton-cfrg-vdaf.md
```diff
@@ -308,6 +339,19 @@ aggregator 1       aggregator 2             aggregator SHARES
...
+RLB: As above, this diagram makes it look like the aggregators only get the outputs.
+-->
+
+<!--
+RLB: The way you've split these things out, the definition of VDAF is really
+just the definition fo the validation stage of the overall process.  I would
+suggest defining DAF and VDAF as each having all the required parts, just with
+DAF having a null / trivial validation stage.  This is because the relevant
+algorithms (sharding, validation, aggregation, unsharding) all need to be
+compatible with one another, so you need to define them together.
```
cjpatton, line 352: DAFs and VDAFs are meant to be distinct primitives. They differ in two important ways: VDAFs require the aggregators to interact; and DAFs don't provide any guarantee about the validity of the output. DAFs and VDAFs aren't necessarily "compatible" with one another. It's true that that there may be some overlap between specific constructions, but this isn't necessarily the case.

To be clear, VDAFs are the interesting primitive here. The main reason I defined DAFs is because they are conceptually simpler and provide a nice warm up to VDAFs. Perhaps we should just drop DAFs altogether?

draft-patton-cfrg-vdaf.md
```diff
@@ -176,6 +176,37 @@ is usually handled by raising an exception.
...
+
+* Roles: 
+  * Submitter - Knows raw metrics
+  * Aggregator - Trusted not to collued
+  * Collector - Receives aggregate at the end of the protocol (may ==
+    Aggregator)
+* Measurement process (covering both DAF and VDAF):
+  0. Raw measurement - Submitter has a measurement to submit
+  1. Sharding - Submitter computes input shares from raw measurement value
+  2. Validation - Aggregators determine that the input shares are valid
```
cjpatton, line 192: Validation applies to VDAFs, not DAFs.

draft-patton-cfrg-vdaf.md
```diff
@@ -176,6 +176,37 @@ is usually handled by raising an exception.
...
+  3. Aggregation - Aggregators compute agregate shares from input shares
+  4. Unsharding - Aggregators / Collector commpute aggregate output from
+     aggregate shares
+
+Matching the below to this framework, I see a few issues:
+
+a. The diagram indicates that the output shares are sent to aggregators
+b. The meaning of the `param` input is unclear.  Is that the working aggregate?
```
cjpatton, line 200: The aggregation parameter is what maps an input to an output. For heavy hitters for example, the aggregation parameter is the set of candidate prefixes. I'll make this more clear.

draft-patton-cfrg-vdaf.md
```diff
@@ -176,6 +176,37 @@ is usually handled by raising an exception.
...
+c. You're missing the unsharding stage entirely
+
+The latter point is important, because the sharding and aggregation algorithms
+need to be compatible.  You kind of cover it in {{aggregatability}}, but that
+should be core part of the description -- a DAF is useless unless it can be
+unsharded.
```
cjpatton, line 206: Right, the way aggregability is defined currently is that the set of output shares is an algebraic group, and the output is recovered by adding up the output shares. Would defining an explicit "unsharding" algorithm be more clear, in your opinion?

draft-patton-cfrg-vdaf.md
```diff
@@ -333,6 +377,8 @@ Syntactically, a VDAF is made up of the following algorithms:
   parameters are generated once and reused across multiple VDAF evaluations. The
   verification parameter MUST NOT be revealed to the clients.
 
+<!-- RLB: Who holds the private key?  One aggregator?  All aggregators? -->
```
cjpatton, line 380: This is something that's a bit in-flux right now. I think where we're going to land is that the client gets a "public key" (the "public parameter") and each aggregator gets its own "secret key" (the "verification parameter").

draft-patton-cfrg-vdaf.md
```diff
@@ -515,6 +561,9 @@ Associated constants:
 
 ## Construction
 
+<!-- RLB: This section should be less code-centric.  At the very least, you
+should provide some prose that describes what the code is doing. -->
```
cjpatton, line 565: That's the plan! This is still very much WIP.

PR #4, opened 2021-10-20
WIP: Define VDAF for heavy hitters

schoppmp:


draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
...
+
+[TODO Clarify which of these can be assumed and which need to be explicitly defined]
+
+### Descriptions of finite fields
+We assume a structure for describing DPF output groups, named `ValueType`.
+For verifiable heavy hitters, we need an IDPF that at each level has two field 
+elements as outputs, and so the value type would be pairs of field elements.
+The IDPF key generation and evaluation functions will allow one `ValueType` parameter
+per bit in the IDPF domain, which allows us to use different types at the output
+layer, required for extractability. See [Section 4.3](https://eprint.iacr.org/2021/017.pdf) in [BBCGGI21] for details, and see [here](https://github.com/google/distributed_point_functions/blob/master/dpf/distributed_point_function.proto#L25) for an example implementation of `ValueType`. We use `ValueType` to refer to a description of a given group, and `Value` to refer to an actual element of the group.
```
cjpatton, line 718: * Replace this reference with `[BBCGGI21, Section 4.3]`
* When this gets compiled, the markdown reference won't appear in the text version. The reference should be added to the  list of informative references at the top of this file.

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
 * Output shares are secret shares of a vector of field elements, each
   corresponding to a counter for one of the candidate prefixes.
 
+## Dependencies
+
+[TODO Clarify which of these can be assumed and which need to be explicitly defined]
+
+### Descriptions of finite fields
```
cjpatton, line 712: Add a newline between headings and the first line.

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
...
+For verifiable heavy hitters, we need an IDPF that at each level has two field 
+elements as outputs, and so the value type would be pairs of field elements.
+The IDPF key generation and evaluation functions will allow one `ValueType` parameter
+per bit in the IDPF domain, which allows us to use different types at the output
+layer, required for extractability. See [Section 4.3](https://eprint.iacr.org/2021/017.pdf) in [BBCGGI21] for details, and see [here](https://github.com/google/distributed_point_functions/blob/master/dpf/distributed_point_function.proto#L25) for an example implementation of `ValueType`. We use `ValueType` to refer to a description of a given group, and `Value` to refer to an actual element of the group.
+
+### Incremental Distributed Point Functions (IDPFs)
+An IDPF is defined over a domain of size `2**d`. The client can specify an index `alpha` and values `beta`, one for each level `l` in `[d]`. The key generation generates two DPF keys that individually hide `alpha` and `beta`. When locally evaluated at any point `x` in `2**l` at level `l`, the IDPF returns shares of `beta[l]`, if `x` is the `l`-bit prefix of `alpha`, and shares of zero otherwise.
```
cjpatton, line 721: ?
```suggestion
An IDPF is defined over a domain of size `2**d`. The client can specify an index `alpha` and values `beta`, one for each level `l` in `[d]`. The key generation generates two IDPF keys that individually hide `alpha` and `beta`. When locally evaluated at any point `x` in `2**l` at level `l`, the IDPF returns shares of `beta[l]`, if `x` is the `l`-bit prefix of `alpha`, and shares of zero otherwise.
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
...
+If elements of `beta` are tuples, the additive sharing is done element-wise.
+
+- `idpf_gen(value_types: Vec[ValueType], alpha: int, beta: Vec[Value]) -> (Bytes, Bytes)`: Takes as input the value types describing the IDPF output domains, as well as the index and values for the IDPF. Returns two serialized DPF keys, one for each helper server.
```
cjpatton, line 724:
```suggestion
- `idpf_gen(value_types: Vec[ValueType], alpha: int, beta: Vec[Value]) -> (Bytes, Bytes)`: Takes as input the value types describing the IDPF output domains, as well as the index and values for the IDPF. Returns two serialized DPF keys, one for each aggregator.
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
...
+- `idpf_eval_next(state: State, dpf_key: Bytes, value_type: ValueType, x: int) -> (new_state: State, value: Value)`: Evaluates the given key at index `x`, at the next prefix length that hasn't beeen evaluated yet for `state`, returning a secret-shared value of the given type.
```
cjpatton, line 725: Should `x` be unsigned? How about `Unsigned` as the type hint, for consistency with the section above?

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
...
+
+### Malicious sketching for IDPFS
```
cjpatton, line 727:
```suggestion
### Malicious sketching for IDPFs
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,147 @@ TODO
...
+      b: b - correlation_shares_0[l]
+      c: c - correlation_shares_0[l]
+      A: A - correlation_shares_0[l]
+      B: B - correlation_shares_0[l]
+    }]
+  key_0, key_1 = idpf_gen(public_param.value_types, input, beta)
+  return ((key_0, correlation_shares_0), (key_1, correlation_shares_1))
+~~~
+
+[TODO: Not sure what to do with `nonce`]
```
cjpatton, line 819: The nonce should be used to generate a seed for the verification randomness vector `r`. For example, in prio3 we have
```python
def vdaf_start(k_query_init, _, nonce, r_input_share):  
  ...
  k_query_rand = get_key(k_query_init, byte(255) + nonce)
  ...
```

where `k_query_init` is the verification parameter.

cjpatton (changes requested):
Good start! I left some mostly editorial comments to start. Some high level editorial things: 
* `make` fails due to trailing whitespace. Make sure you can run `make` locally and that it builds successfully. (If you've never built an Internet-Draft before, let me know and I can help you get started!)
* Please wrap lines at 80 characters.

As for the design:
1. The purpose of the nonce is to allow the aggregators to derive fresh randomness per VDAF evaluation. This randomness could be derived by applying a PRF to the nonce using the key generated for the verification parameters.
2. I'm curious why you allow the caller to specify the group parameters for each level of the tree. I would think it's better if the scheme *prescribed* these parameters. In other words, a concrete IDPF would specify the field parameters of each level of the tree.

draft-patton-cfrg-vdaf.md
schoppmp: I removed the second reference, since the value type implementation is not really important if we assume that the concrete scheme fixes the types for each level.

draft-patton-cfrg-vdaf.md
schoppmp: Done.

draft-patton-cfrg-vdaf.md
schoppmp: Done.

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+
+### Descriptions of finite fields
+
+We assume a structure for describing IDPF output groups, named `ValueType`.
+For verifiable heavy hitters, we need an IDPF that at each level has two field
+elements as outputs, and so the value type would be pairs of field elements.
+The IDPF key generation and evaluation functions will allow a different `ValueType`
+parameter per bit in the IDPF domain, which allows us to use different types at
+the output layer, required for extractability. See [BBCGGI21, Section 4.3] for
+details. We assume a function `get_value_type(n: Unsigned) -> ValueType` that
```
cjpatton, line 721:
```suggestion
details. An instantiation of this VDAF specifies a function `get_value_type(n: Unsigned) -> ValueType` that
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+returns the type for bit index `n`. We use `ValueType` to refer to a description
+of a given group, and`Value` to refer to an actual element of the group.
```
cjpatton, line 723:
```suggestion
of a given group, and `Value` to refer to an actual element of the group.
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+
+### Incremental Distributed Point Functions (IDPFs)
+
+An IDPF is defined over a domain of size `2**d`. The client can specify an
```
cjpatton, line 727: I know we're supposed to be doing everything in Python, but let's just write this as
```suggestion
An IDPF is defined over a domain of size `2^d`. The client can specify an
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+index `alpha` and values `beta`, one for each level `l` in `[d]`. The key
+generation generates two IDPF keys that individually hide `alpha` and `beta`.
+When locally evaluated at any point `x` in `2**l` at level `l`, the IDPF
+returns shares of `beta[l]`, if `x` is the `l`-bit prefix of `alpha`, and
```
cjpatton, line 731: Would it be more accurate to say
```suggestion
When locally evaluated at any point `x` in `2**l` at level `l`, the IDPF
key returns a share of `beta[l]`, if `x` is the `l`-bit prefix of `alpha`, and
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+shares of zero otherwise. If elements of `beta` are tuples, the additive
+sharing is done element-wise.
+
+- `idpf_gen(alpha: Unsigned, beta: Vec[Value]) -> (Bytes, Bytes)`:
```
cjpatton, line 735: I've been naming the output parameters. How about `-> (idpf_key1: Bytes, idpf_key2: Bytes)` or something?

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+Takes as input the index and values for the IDPF. Returns two serialized IDPF
+keys, one for each helper server.
```
cjpatton, line 737:
```suggestion
keys, one for each aggregator.
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+- `idpf_eval_next(state: State, dpf_key: Bytes, x: Unsigned) `
```
cjpatton, line 738:
```suggestion
- `idpf_eval_next(state: State, idpf_key: Bytes, x: Unsigned) `
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+`-> (new_state: State, value: Value)`:
+Evaluates the given key at index `x`, at the next prefix length that hasn't
+been evaluated yet for `state`, returning a secret-shared value.
+
+### Malicious sketching for IDPFs
+
+~~~
+def verify_sketch_round_1(dpf_output: Vec[Value], correlation_share, joint_randomness: Bytes, level: Unsigned):
+  # Expand common randomness and compute verification message
+  r = expand_rand(joint_randomness, len(dpf_output), level)
```
cjpatton, line 748: Above I use "joint rand" to refer to randomness shared by the prover and verifier for PCPs. Can we call this something else, like "verify_rand"?

cjpatton, line 748: Here and elsewhere
```suggestion
  r = expand_rand(joint_randomness, len(idpf_output), level)
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+  r_squared = [v*v for v in r]
+  ver_msg_1 = inner_product(
+    [value[0] for value in dpf_values,
```
cjpatton, line 751: Here and below
```suggestion
    [value[0] for value in dpf_output,
```

draft-patton-cfrg-vdaf.md
```diff
@@ -705,6 +705,159 @@ TODO
...
+  return ((key_0, correlation_shares_0), (key_1, correlation_shares_1))
+~~~
+
+~~~
+def vdaf_start(verify_param, public_param, nonce, input_share):
+  key, correlation_share = input_share
+
+  # Expand left and right children of the DPF tree.
+  state_left, value_left = idpf_eval_next(None, key, 0)
+  state_right, value_right = idpf_eval_next(None, key, 1)
```
cjpatton, line 837: I'm having a hard time understanding what's going on here. It seeems like you want the IDPF api to be stateful, but the VDAF api is not. So how does the state of the IPDF get passed through to subsequent VDAF evaluations?

Unless I'm misunderstanding, I think the answer is that, at least as far as this spec is concerned, the IDPF api needs to be stateless.

PR #14, opened 2021-12-28
Add public preprocessing of messages

### LINKED ISSUE #11: Syntax: Public preprocessing during the preparation phase
For both prio3 and hits, each round of the preparation phase consists of a step of "preprocessing" of the previous round's messages. This preprocessing step is identical for all of the aggregators. In hits, for example, each aggregator begins each round by summing up the messages from the previous round. From Figure 15 in https://cjpatton.github.io/vdaf/draft-patton-cfrg-vdaf.html#name-preparation-3:
```
    ...
    elif self.step == "sketch round 1" and len(inbound) == 2:
      verifier_1 = Field[l].deocde_vec(inbound[0]) + \             # ADD UP INPUT MESSAGES
                   Field[l].deocde_vec(inbound[1])

      verifier_share_2 = [
        (verifier_1[0] * verifier_1[0] \
         - verifier_1[1] \
         - verifier_1[2]) * self.party_id \
        + A_share * verifer_1[0] \
        + B_share
      ]

      self.step = "sketch round 2"
      return Field[l].encode_vec(verifier_share_2)

    elif self.step == "sketch round 2" and len(inbound) == 2:
      verifier_2 = Field[l].decode_vec(inbound[0]) + \             # ADD UP INPUT MESSAGES
                   Field[l].decode_vec(inbound[1])

      if verifier_2 != 0: raise ERR_INVALID
      return Field[l].encode_vec(self.output_share)
   ...
```
The essence of what's happening here is that, in each round, the aggregators are exchanging additive shares of a single message. prio3 is similar, although each message consists of an XOR-share of as byte string and an additive share of the FLP verifier. From Figure 10:
```
   ...
   elif self.step == "waiting" and len(inbound) == SHARES:
      k_joint_rand = zeros(KEY_SIZE)
      verifier = vec_zeros(VERIFIER_LEN)
      for r_share in inbound:
        (k_joint_rand_share,
         verifier_share) = decode_verifier_share(r_share)

        k_joint_rand ^= k_joint_rand_share                           # ADD UP xor SHARES
        verifer += verifier_share                                    # ADD UP ADDITIVE SHARES

      if k_joint_rand != self.k_joint_rand: raise ERR_INVALID
      if not flp_decide(verifier): raise ERR_INVALID
      return Field.encode_vec(self.output_share)
   ...
```
What we're likely to do in the PPM protocol is have one aggregator -- the Leader -- collect the verifier messages at each round and relay them to the other aggregators. We could save on communication cost if we have the Leader perform this public preprocessing step itself and disseminate the result.

To make this optimization generic, I'd like to propose that we lift the public processing step to an explicit part of the syntax. In particular, we'd change the preparation phase as follows (from Figure 6):
```
   # Each aggregator initializes its preparation state.
    prep_states = []
    for j in range(SHARES):
      prep_states.append(PrepState(
          verify_params[j], agg_param, nonce, input_shares[j]))

    # Aggregators recover their output shares.
-   inbound = []
+   inbound = ""
    for i in range(ROUNDS+1):
      outbound = []
      for j in range(SHARES):
        outbound.append(prep_states[j].next(inbound))
      # This is where we would send messages over the network
      # in a distributed VDAF computation.
-     inbound = outbound
+     inbound = prep_preprocess(j, outbound)

    # The final outputs of validation are the output shares
    # for this input.
    output_shares.append(outbound)
```

I don't think this syntactic change would decrease generality, but I also don't want it to be a nuisance. Maybe it makes sense to not call this out explicitly?

cjpatton:
Closes #11.

This makes explicit a feature of prio3 and hits that allows applications
to save on bandwidth. At the same time, it does not decrease generality
since this preprocessing step might simply be a no-op.

This change also replaces reference to `vec_zeros()` with
`Field.zeros()`.


draft-patton-cfrg-vdaf.md
```diff
@@ -490,13 +494,18 @@ share or an error."}
...
   output share. An exception is raised if a valid output share could not be
-  recovered. The input of this algorithm is the sequence of inbound messages
-  from the previous round or, if this is the first round, an empty vector.
+  recovered. The input of this algorithm is the preprocessed inbound message or,
+  if this is the first round, `None`.
+
+* `prep_preprocess(agg_param, messages: Vec[Bytes]) -> preprocessed_message` is
+  a deterministic procedure used to preprocess the sequence of inbound messages
+  from the previous round into the input for the preparation-state update
+  algorithm.
```
cjpatton, line 508: Can we think of a better verb than "preprocess"? While accurate, it seems too generic.

draft-patton-cfrg-vdaf.md
```diff
@@ -490,13 +494,18 @@ share or an error."}
   {{run-vdaf}}), and one of the input shares generated by the client
   (`input_share`). Its outputs is the Aggregator's initial preparation state.
 
-* `PrepState.next(inbound: Vec[Bytes]) -> outbound` is the deterministic
+* `PrepState.next(inbound: Optional[Bytes]) -> outbound` is the deterministic
```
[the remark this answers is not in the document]
cjpatton: `Bytes` is itself a vector of bytes.

draft-patton-cfrg-vdaf.md
schoppmp, line 508: Can you elaborate a bit what the `prep_preprocess` step does and in what cases it is useful? This doesn't become clear from the description currently.

schoppmp: I agree, in particular the double `pre` prefix is a bit confusing. Maybe `combine_verifier_shares` or something like that?

tgeoghegan:
> It seems like this PR is just splitting `next` into two steps that both happen inside the same computer. Why is this useful?
> 
> In general, I think the design philosophy should be that the inputs and outputs from these abstract functions should be to the network, i.e., the boundaries that matter are the ones that are visible from the outside. How the processing is arranged inside the computer, between network-observable points, is up to the individual implementation.

The idea is that the leader can do the `prep_preprocess` step once and broadcast the result to however many helpers there are, whereas previously, every aggregator would have had to broadcast prep messages to every other aggregator (extra network communication) and then every aggregator would have to run the prep message assembly (extra CPU time). The reduction in network communication is a much more significant win than the reduction of redundant work, but the it's hard to see that communication graph in this specification, because it's all abstracted away behind the `inputs = outputs` step, which in a real deployment would be a big, messy graph with `n^2` edges in it.

cjpatton:
With the current syntax:  In a Leader/Helper communication model, the Leader would request each Helper's next prepare message in one round, then broadcast the entire round of messages to each Helper in the next round.

With the changed syntax: Instead of broadcasting all of the messages, the Leader can perform this precomputation step, which for both `prio3` and `poplar1` reduces the amount of data broadcast by a factor of `SHARES`.

draft-patton-cfrg-vdaf.md
cjpatton: I like the term `verifier_share`, but it's somewhat specific to schemes like `prio3` and `poplar1` where the output share are first recovered and then verified. At this stage I'd like to leave the door open for things like Prio+ that are based instead on oblivious transfer. (Output shares aren't known until the prep phase is complete.)

cjpatton: I went with "prepare-message share", which is overly wordy but says what we want. The method is now called `prep_share_to_prep`, which is somewhat consistent with our naming conventions else where. 

Other suggestions for alternatives are welcome! (Naming things is hard.)

cjpatton: I've tacked on a commit that will hopefully make this more clear.

cjpatton:
I'm going to make an executive decision and merge this by Wednesday, 2022/01/19 unless anyone flags objections to this change.

draft-patton-cfrg-vdaf.md
```diff
@@ -513,12 +528,12 @@ schemes: recovery of output shares from the input shares, and a multi-party
...
 recovers an output share if it is deemed valid (at least, based on the
-Aggregator's view of the protocol). Another way to draw this boundary woulds be
+Aggregator's view of the protocol). Another way to draw this boundary would be
 to have the Aggregators recover output shares first, then verify that they are
 valid. However, this would allow the possibility of misusing the API by, say,
-aggregating an invalid output share. Moreover, in some protocols, like Prio+
-[AGJOP21] it is necessary for the Aggregators to interact in order to recover
-output shares at all.
+aggregating an invalid output share. Moreover, in protocols like Prio+ [AGJOP21]
+based on oblivious transffer, it is necessary for the Aggregators to interact in
```
schoppmp, line 535: `transffer` -> `transfer`

schoppmp (approved):

draft-patton-cfrg-vdaf.md
cjpatton: Done

bifurcation:
I'm OK with this, as long as we agree that the degenerate case where `prep_shares_to_prep` just concatenates the shares is in-bounds.

cjpatton:
> I'm OK with this, as long as we agree that the degenerate case where `prep_shares_to_prep` just concatenates the shares is in-bounds (which puts us back in the broadcast posture).

Yeah I agree. It would be premature to rule this out.

cjpatton:
Squashed. Thanks all for your feedback!

PR #16, opened 2022-01-04
Rename "Heavy Hitters" to "Poplar"

cjpatton:
The authors of [BBCGGI21] have recently amended their paper to give the
protocol a proper name.

### DIFF
--- a/draft-patton-cfrg-vdaf.md
+++ b/draft-patton-cfrg-vdaf.md
@@ -241,25 +241,25 @@ from the literature.
   framework as the original Prio protocol, but incorporates techniques
   introduced in [BBCGGI19] that result in significant performance gains.
 
-* More recently, Boneh et al. [BBCGGI21] described a protocol for solving the
-  `t`-heavy-hitters problem in a privacy-preserving manner. Here each client
-  holds a bit-string of length `n`, and the goal of the aggregation servers is
-  to compute the set of inputs that occur at least `t` times. The core primitive
-  used in their protocol is a generalization of a Distributed Point Function
-  (DPF) [GI14] that allows the servers to "query" their DPF shares on any
-  bit-string of length shorter than or equal to `n`. As a result of this query,
-  each of the servers has an additive share of a bit indicating whether the
-  string is a prefix of the client's input. The protocol also specifies a
-  multi-party computation for verifying that at most one string among a set of
-  candidates is a prefix of the client's input.
+* More recently, Boneh et al. [BBCGGI21] described a protocol called Poplar for
+  solving the `t`-heavy-hitters problem in a privacy-preserving manner. Here
+  each client holds a bit-string of length `n`, and the goal of the aggregation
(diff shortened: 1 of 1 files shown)


draft-patton-cfrg-vdaf.md
```diff
@@ -241,25 +241,25 @@ from the literature.
...
+  servers is to compute the set of inputs that occur at least `t` times. The
+  core primitive used in their protocol is a generalization of a Distributed
+  Point Function (DPF) [GI14] that allows the servers to "query" their DPF
+  shares on any bit-string of length shorter than or equal to `n`. As a result
+  of this query, each of the servers has an additive share of a bit indicating
+  whether the string is a prefix of the client's input. The protocol also
+  specifies a multi-party computation for verifying that at most one string
+  among a set of candidates is a prefix of the client's input.
+
+  In {{pops}} we describe a VDAF called `pops` that implements this
```
bifurcation, line 256: I would just call this `poplar`, since we just use `prio` for Prio.

bifurcation (approved):

draft-patton-cfrg-vdaf.md
tgeoghegan: In fact we use `prio3` for Prio, so it's not even like there's a 4-letter code convention.

cjpatton: It's tricky because the full-blown heavy hitters protocol is called "Poplar". I wanted a distinct name for the underlying VDAF. "pops" is the name that @henrycg and I discussed over email.

cjpatton: We could go with something more verbose, like "poplar-vdaf".

tgeoghegan: I prefer `poplar` but I'm happy to defer to you as the author here.

tgeoghegan (approved):

draft-patton-cfrg-vdaf.md
cjpatton: I'd like to give folks a chance to mull on it, so we''ll keep this open for now.

schoppmp: I'm not sure about `pops`, at first it looks like it's an acronym, but it isn't. I don't think using `poplar` for both the end-to-end protocol and the VDAF is too much of an issue. Or @cjpatton do you think there will be instantiations of poplar-the-protocol that use a different VDAF? And in that case, couldn't that VDAF be called `poplar2` (the same way the current Prio VDAF is `prio3`?
Alternatively, I think using `poplar-vdaf` (and also changing `prio3` to `prio3-vdaf` or just `prio-vdaf`) is also fine.

cjpatton: Cool, one more point of agreement :) As discussed in https://github.com/abetterinternet/ppm-specification/pull/177#discussion_r778413433 we're going with `poplar1`, for the same reason you point out.

cjpatton:
s/pops/poplar1

PR #25, opened 2022-02-08
Specify a method for deriving pseudorandom field elements

### LINKED ISSUE #13: Specify method for generating field elements (was "Proposal: use hash to field to generate field elements from a seed")

and
      len.
>

cjpatton:
Closes #13.


cjpatton:
Re-opening this PR with a new base.

PR #28, opened 2022-02-14
Specify an instantiation of `Prg`

cjpatton:
Specifies a construction based on AES128. While pretty straight-forward,
this scheme uses CMAC, which is not widely used. We may want to provide
additional options in the future.

### DIFF
--- a/draft-patton-cfrg-vdaf.md
+++ b/draft-patton-cfrg-vdaf.md
@@ -906,7 +906,44 @@ def expand_into_vec(Prg, Field,
 
 ### Constructions {#prg-constructions}
 
-> TODO
+#### PrgAes128
+
+> OPEN ISSUE Phillipp points out that a fixed-key mode of AES may be more
+> performant (https://eprint.iacr.org/2019/074.pdf). See
+> https://github.com/cjpatton/vdaf/issues/32 for details.
+
+Our first construction, `PrgBlockCipher(Aes128)`, converts a blockcipher, namely
+AES-128, into a PRG. Seed expansion involves two steps. In the first step, CMAC
+{{!RFC4493}} is applied to the seed and info string to get a fresh key. In the
+second step, the fresh key is used in CTR-mode to produce a key stream for
+generating the output. A fixed initialization vector (IV) is used.
+
+~~~
+class PrgAes128:
+
+    SEED_SIZE: Unsigned = 16
+
+    def __init__(self, seed, info):
+        self.length_consumed = 0
+
+        # Use CMAC as a pseuodorandom function to derive a key.
+        self.key = AES128-CMAC(seed, info)
+
+    def next(self, length):
+        self.length_consumed += length
+
+        # CTR-mode encryption of the all-zero string of the desired
+        # length and using a fixed, all-zero IV.
+        #
+        # TODO Make this implementation more efficient. In particular,
+        # don't regenerate the already consumed stream.
+        stream = AES128-CTR(key, zeros(16), zeros(self.length_consumed))
+        return stream[-length:]
+~~~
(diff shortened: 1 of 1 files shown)


draft-patton-cfrg-vdaf.md
```diff
@@ -871,7 +871,47 @@ def expand_into_vec(Prg, Field,
...
 ### Constructions {#prg-constructions}
 
-> TODO
+#### PrgBlockCipher(Aes128)
+
+Our first construction, `PrgBlockCipher(Aes128)`, converts a blockcipher, namely
+AES-128, into a PRG. Seed expansion involves two steps. In the first step, CMAC
+{{!RFC4493}} is applied to the seed and info string to get a fresh key. In the
+second step, the fresh key is used in CTR-mode to produce a key stream for
+generating the ouutput. A fixed initialization vector (IV) is used.
```
armfazh, line 880:
```suggestion
generating the output. A fixed initialization vector (IV) is used.
```

draft-patton-cfrg-vdaf.md
```diff
@@ -871,7 +871,47 @@ def expand_into_vec(Prg, Field,
...
+
+~~~
+SEED_SIZE=16
+
+def expand(_, seed, info, length):
+    key = AES128-CMAC(seed, info)
+
+    # CTR-mode encryption of the all-zero string of the desired length and using
+    a fixed, all-zero IV.
```
armfazh, line 889:
```suggestion
    # a fixed, all-zero IV.
```

armfazh (approved):

schoppmp (commented):
This will likely be called often, but with small `length`s, right? In that case I believe constructions based on [fixed-key AES hashing](https://eprint.iacr.org/2019/074.pdf) should be more efficient, since they avoid repeating the AES key schedule. In the paper there is a tweakable version (using two AES calls) that can be used to derive new seeds for different `info` strings.

cjpatton:
> This will likely be called often, but with small `length`s, right? In that case I believe constructions based on [fixed-key AES hashing](https://eprint.iacr.org/2019/074.pdf) should be more efficient, since they avoid repeating the AES key schedule. In the paper there is a tweakable version (using two AES calls) that can be used to derive new seeds for different `info` strings.

Great question. There are two key schedules:
* One for CMAC. The same key is used for each blockcipher call, including for generating the subkey K1 (or K2, depending on the length of the message.) See https://datatracker.ietf.org/doc/html/rfc4493#section-2.2.
* One for CTR.

Eventually the cumulative cost of AES costs will outstrip the cost of the initial setup of these two key schedules, so the question is whether we reach this point in our code. There are at least two cases to consider: 
1. In #31, each `length` passed to `prg.next()` is small, but we call `prg.next()` many times. Hence the computational cost will be dominated in `AES128-CTR()`.
2. In prio3, we use `Prg.derive()` as a PRF. Here the input is long (though not as long as in (1.)), but the output is short. Hence the computational cost will be dominated by `AES128-CMAC()`.

My hunch is that we cross this threshold fairly quickly for (1.), but I'm not so sure about (2.). I think this is something worth benchmarking in any case.

I suggest we put a pin in this for now. I'll mark this is an "OPEN ISSUE" in the text and we can open an issue for tracking benchmarking of various PRGs.

cjpatton:
I've added a pointer to https://github.com/cjpatton/vdaf/issues/32 to the text.

schoppmp:
> I suggest we put a pin in this for now. I'll mark this is an "OPEN ISSUE" in the text and we can open an issue for tracking benchmarking of various PRGs.

Sounds good, thanks!

schoppmp (approved):

PR #29, opened 2022-02-14
Use "pad-then-reduce" method for generating field elements

### LINKED ISSUE #13: Specify method for generating field elements (was "Proposal: use hash to field to generate field elements from a seed")

cjpatton:
Closes #13 by taking @armfazh' suggestion.

Specifies an algorithm for mapping a pseudorandom byte string to a
vector of field elements. This method is based on the "hash_to_field"
algorithm from draft-irtf-cfrg-hash-to-curve-13.

### DIFF
--- a/draft-patton-cfrg-vdaf.md
+++ b/draft-patton-cfrg-vdaf.md
@@ -764,7 +764,7 @@ parameters:
   as a byte string.
 
 * `EXPANDED_SIZE: Unsigned` is the number of pseudorandom bytes that are sampled
-  when per field element when expanding a seed into a sequence of field elements
+  per field element when expanding a seed into a sequence of field elements
   (see {{prg}}).
 
   > CP It might be helpful to specify valid ranges for `EXPANDED_SIZE`. Perhaps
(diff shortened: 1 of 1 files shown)


draft-patton-cfrg-vdaf.md
```diff
@@ -856,7 +864,9 @@ A concrete `Prg` implements the following class method:
 
 Each `Prg` has two derived class methods. The first is used to derive a fresh
 seed from an existing one. The second is used to compute a sequence of
-pseudorandom field elements froma pseudorandom byte string.
+pseudorandom field elements froma pseudorandom byte string. This method used is
```
armfazh, line 867: ... The second one generates a sequence of field elements from a pseudorandom byte string. Such a generation is based on ...

draft-patton-cfrg-vdaf.md
```diff
@@ -763,6 +763,14 @@ parameters:
 * `ENCODED_SIZE: Unsigned` is the number of bytes used to encode a field element
   as a byte string.
 
+* `EXPANDED_SIZE: Unsigned` is the number of pseudorandom bytes that are sampled
+  when per field element when expanding a seed into a sequence of field elements
```
armfazh, line 767:
```suggestion
  per field element when expanding a seed into a sequence of field elements
```

draft-patton-cfrg-vdaf.md
cjpatton: Done

cjpatton:
Closed in favor of #31.

PR #31, opened 2022-02-25
Use "rejection sampling" for generating field elements

### LINKED ISSUE #13: Specify method for generating field elements (was "Proposal: use hash to field to generate field elements from a seed")

cjpatton:
Closes #13 by doing "rejection sampling". This an alternative to #29 and #30.

### DIFF
--- a/draft-patton-cfrg-vdaf.md
+++ b/draft-patton-cfrg-vdaf.md
@@ -856,37 +856,53 @@ FFT-friendly fields also define the following parameter:
 
 ## Pseudorandom Generators {#prg}
 
+> CP TODO Update the reference implementation to match.
+
 A pseudorandom generator (PRG) is used to expand a short, (pseudo)random seed
 into a long string of pseudorandom bits. A PRG suitable for this document
 implements the interface specified in this section. Concrete constructions are
 described in {{prg-constructions}}.
 
-PRGs are defined by a class `Prg` with a single parameter:
+PRGs are defined by a class `Prg` with the follwing associated parameter:
 
 * `SEED_SIZE: Unsigned` is the size of the input seed.
 
 A concrete `Prg` implements the following class method:
 
-* `Prg.expand(seed: Bytes, info: Bytes, length: Unsigned) -> Bytes` derives a
-  byte string from the given seed and info string. The length of the output MUST
-  be `length`.
+* `Prg(seed: Bytes, info: Bytes, length: Unsigned)` constructs an instance of
+  `Prg` from the given seed and info string. The seed MUST be of length
+  `SEED_SIZE` and MUST be generated securely (i.e., it is either the output of
+  `gen_rand` or a previous invocation of the PRG). The info string is used for
+  domain separation.
+
+* `prg.next(length: Unsigned)` returns the next `length` bytes of output of the
+  PRG. If the seed was securely generated, the output can be treated as
+  pseudorandom.
 
 Each `Prg` has two derived class methods. The first is used to derive a fresh
 seed from an existing one. The second is used to compute a sequence of
-pseudorandom field elements.
+pseudorandom field elements. For each method, the seed MUST be of length
+`SEED_SIZE` and MUST be generated securely (i.e., it is either the output of
+`gen_rand` or a previous invocation of the PRG).
 
 ~~~
 # Derive a new seed.
 def derive(Prg, seed: bytes, info: bytes) -> bytes:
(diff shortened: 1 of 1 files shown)


draft-patton-cfrg-vdaf.md
```diff
@@ -839,20 +839,24 @@ def vec_add(left, right):
...
 A concrete `Prg` implements the following class method:
 
-* `Prg.expand(seed: Bytes, info: Bytes, length: Unsigned) -> Bytes` derives a
-  byte string from the given seed and info string. The length of the output MUST
-  be `length`.
+* `Prg(seed: Bytes, info: Bytes, length: Unsigned)` constructs an instance of
+  `Prg` from the given seed and info string.
+
+* `prg.next(length: Unsigned)` returns the next `length` bytes of output of the
+  PRG.
```
schoppmp, line 859: Please add some info about what we expect from the output, e.g.
`
The bytes returned are uniformly pseudorandom when the PRG was constructed with a uniformly random seed.
`

draft-patton-cfrg-vdaf.md
```diff
@@ -861,15 +865,21 @@ pseudorandom field elements.
 ~~~
 # Derive a new seed.
```
schoppmp, line 866: Can you add some information for why this is needed? Also please add that the seed MUST have high entropy.

draft-patton-cfrg-vdaf.md
```diff
@@ -861,15 +865,21 @@ pseudorandom field elements.
...
 def derive(Prg, seed: bytes, info: bytes) -> bytes:
-    return Prg.expand(seed, info, Prg.SEED_SIZE)
+    prg = Prg(seed, info)
+    return prg.next(Prg.SEED_SIZE)
 
 # Expand a seed into a vector of Field elements.
```
schoppmp, line 871: Same as above, please make sure that this is always used in a safe way by requiring that the seed MUST have high entropy and be only used once.

draft-patton-cfrg-vdaf.md
cjpatton: Done. I also stipulated that the random seed might be an output of a previous call to the PRG.

draft-patton-cfrg-vdaf.md
cjpatton: The intended usage is explained above:
> Each `Prg` has two derived class methods. The first is used to derive a fresh seed from an existing one.

If you want to see this get used, search for `Prg.derive` in the prio3 section.

High entropy ins't going to be sufficient for all PRGs, in particular for those constructed from a pseudorandom permutation (e.g., AES). I stipulated the same requirement for securely generating seeds as above.

draft-patton-cfrg-vdaf.md
cjpatton: It depends on the PRG construction, but it should be secure to reuse the same seed with different "info" strings. That is, the intent of the "info" string is to provide domain separation.

Here I stipulated the same requirements for generating the seed as above.

draft-patton-cfrg-vdaf.md
schoppmp: The current documentation sounds good, thanks!

draft-patton-cfrg-vdaf.md
```diff
@@ -839,37 +839,52 @@ def vec_add(left, right):
...
-  byte string from the given seed and info string. The length of the output MUST
-  be `length`.
+* `Prg(seed: Bytes, info: Bytes, length: Unsigned)` constructs an instance of
+  `Prg` from the given seed and info string. The seed MUST be of length
+  `SEED_SIZE` and MUST be generated securely (i.e., it is either the output of
+  `gen_rand` or a previous invocation of the PRG).
+
+* `prg.next(length: Unsigned)` returns the next `length` bytes of output of the
+  PRG. If the seed was securely generated, the output can be treated as
+  pseudorandom.
```
schoppmp, line 862: Maybe add a note about `info` and domain separation?

draft-patton-cfrg-vdaf.md
schoppmp: Okay, let's maybe add a short sentence on this to the documentation (see comment above).

schoppmp (approved):

draft-patton-cfrg-vdaf.md
cjpatton: Done

PR #41, opened 2022-03-28
Small fix to prime selection script

divergentdave:
This fixes a small bug in the prime searching script. Previously, it would skip candidate q values, since it decremented q by i, rather than check each `2^(n-k)-i`. In the interest of completeness, we now check each possible value.


cjpatton (approved):
Thanks for checking :) Note that there is still time to pick new primes if we so choose.

Issue #47, opened 2022-04-26
Define a method for merging multiple aggregate shares into an aggregate share


tgeoghegan:
A VDAF must define a few [associated types](https://github.com/cjpatton/vdaf/blob/15c6abadc46d760a2b39a5708492a195fb4ee2e9/draft-patton-cfrg-vdaf.md?plain=1#L400):

```
| `OutShare`    | Type of each output share     |
| `AggShare`    | Type of each aggregate share  |
| `AggResult`   | Type of the aggregate result  |
```

VDAF defines a method for combining multiple output shares into an aggregate share: [`Vdaf.out_shares_to_agg_share(agg_param: AggParam, output_shares: Vec[OutShare]) -> agg_share: AggShare`](https://github.com/cjpatton/vdaf/blob/15c6abadc46d760a2b39a5708492a195fb4ee2e9/draft-patton-cfrg-vdaf.md?plain=1#L590). That's implemented in libprio as [`prio::vdaf::Aggregator::aggregate`](https://github.com/divviup/libprio-rs/blob/1bd54185d01110ff4486b8d71ed8f17f1ea77b78/src/vdaf.rs#L201).

VDAF also defines a method for combining multiple aggregate shares into an aggregate result: [`Vdaf.agg_shares_to_result(agg_param: AggParam, agg_shares: Vec[AggShare]) -> AggResult`](https://github.com/cjpatton/vdaf/blob/15c6abadc46d760a2b39a5708492a195fb4ee2e9/draft-patton-cfrg-vdaf.md?plain=1#L627). That's implemented in libprio as [`prio::vdaf::Collector::unshard`](https://github.com/divviup/libprio-rs/blob/1bd54185d01110ff4486b8d71ed8f17f1ea77b78/src/vdaf.rs#L214).

libprio's trait [`prio::vdaf::Aggregatable`](https://github.com/divviup/libprio-rs/blob/1bd54185d01110ff4486b8d71ed8f17f1ea77b78/src/vdaf.rs#L236) is meant to correspond to a VDAF `AggShare` and also provides the method `Aggregatable:merge(&self, other: &Self) -> Result<(), VdafError>`. This method is extremely useful if not required for PPM implementations which will want to implement a map-reduce strategy for constructing aggregate shares. VDAF should define a method that corresponds to libprio's `merge`, and perhaps require that the combination of VDAF output shares into aggregate shares should not depend on order of the output shares.

[...]

PR #53, opened 2022-04-30
Define DAFs

### LINKED ISSUE #20: What to do about DAFs (VDAFs w/o verifiability)
The syntax is compatible with schemes that forego verifiability altogether, but calling such a scheme a "VDAF" would be misleading. It's probably a good idea to make this distinction explicit, i.e., define "DAFs" as 0-round VDAFs that provide no verifiability. Should we make room for this in this document, or does it make sense to kick this to a different document?

Note that this question came up previously in the context of PPM: https://github.com/abetterinternet/ppm-specification/issues/45.

cjpatton:
Closes #20.

This defines syntax for DAFs. They are presented as a simpler alternative to VDAFs that may be suited for some applications. They are defined before VDAFs by way of a "gentle introduction". Text in the VDAF section that overlaps significantly with the new DAF text has been removed.

Reviewer note: This PR is broken up into several commits. It may be best to review commit-by-commit.

2022-05-06: Ready for review.
2022-04-29: WIP, not yet ready for review.


draft-irtf-cfrg-vdaf.md
```diff
@@ -342,70 +359,284 @@ measurement process is as follows:
...
+                                                  out_shares_j)
+        agg_shares.append(agg_share_j)
+
+    # Collector unshards the aggregate result.
+    agg_result = Daf.agg_shares_to_result(agg_param, agg_shares)
+    return agg_result
+~~~
+{: #run-daf title="Execution of a DAF."}
+
+The inputs to this proedure are the same as the aggregation function computed by
```
divergentdave, line 584: nit: spelling
```suggestion
The inputs to this procedure are the same as the aggregation function computed by
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -342,70 +359,284 @@ measurement process is as follows:
...
 
-A concrete VDAF specifies the algorithms involved in evaluating an aggregation
-function across a batch of inputs. This section specifies the interfaces of
-these algorithms as they would be exposed to applications.
+Like DAFs described in the previous section, a VDAF scheme is used to compute a
+particular aggregation function over a set of Client-generated measurements.
+Evaluation of a VDAF involves the same four stages as for DAFs: Sharding,
+Preparation, Aggregation, and Unsharding. However, the Preparation stage will
+require interaction among the Aggregators in order to facilitate verifiability
+of the computation's correctness. Accommadating this interaction will require
```
divergentdave, line 603: nit: spelling
```suggestion
of the computation's correctness. Accommodating this interaction will require
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -428,38 +659,20 @@ generated by the following algorithm:
 
 ## Sharding {#sec-vdaf-shard}
 
-In order to protect the privacy of its measurements, a VDAF Client splits its
-measurements into "input shares".  The `measurement_to_input_shares` method is
-used for this purpose.
+Sharding is identical to DAFs (cf. {{sec-daf-shard}}) except that the
+input--distribution algorithm takes in the public parameter generated during the
```
divergentdave, line 663: nit: typo
```suggestion
input-distribution algorithm takes in the public parameter generated during the
```

poc/daf.sage
```diff
@@ -0,0 +1,184 @@
...
+import json
+
+
+# A DAF
+class Daf:
+
+    # The number of Aggregators.
+    SHARES: Unsigned = None
+
+    # The meeasurement type.
```
divergentdave, line 15:
```suggestion
    # The measurement type.
```

poc/daf.sage
```diff
@@ -0,0 +1,184 @@
...
+    # Merge a list of output shares into an aggregate share. This is called by
+    # an Aggregator after recovering a batch of output shares.
+    @classmethod
+    def out_shares_to_agg_share(Daf,
+                                agg_param: AggParam,
+                                out_shares: Vec[Prep.OutShare]) -> AggShare:
+        raise Error("not implemented")
+
+    # Unshard the aggregate shares and compute the aggregate result. This is
+    # called by the ccollector.
```
divergentdave, line 55: nit: typo
```suggestion
    # called by the collector.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -342,70 +359,284 @@ measurement process is as follows:
...
+to the aggregation function, but in general it is used to represent the set of
+"queries" that can be made on the measurement set
+
+Execution of a DAF has four distinct stages:
+
+* Sharding - Each Client generates input shares from its measurement and
+  distributes them among the Aggregators.
+* Preparation - Each Aggregator converts each input share into an output share
+  compatible with the aggregation function. This computation involves the
+  aggregation parameter. In general, there may be a one-to-many mapping from
```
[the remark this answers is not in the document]
divergentdave: Ah, got it, thanks! New text looks good.

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+By way of a gentle introduction to VDAFs, this section describes a simpler class
+of schemes called Distributed Aggregation Functions (DAFs). Unlike VDAFs, DAFs
+do not provide verifiability of the computation. Clients must therefore be
+trusted to compute their input shares correctly. Because of this fact, the use
+of a DAF is NOT RECOMMENDED for most applications. See {{security}} for
+additional discussion.
+
+A DAF scheme is used to compute a particular "aggregation function" over a set
+of measurements generated by Clients. An aggregation function may include an
+optional "aggregation parameter", which is selected by the Collector and
```
tgeoghegan, line 391: "may" and "optional" are redundant.

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+disseminated among the Aggregators. The semantics of this parameter is specific
```
tgeoghegan, line 392: "The semantics of this parameter *are* specific"

(GitHub won't let me use the suggestion feature when I review commit-by-commit)

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+to the aggregation function, but in general it is used to represent the set of
+"queries" that can be made on the measurement set
```
tgeoghegan, line 394: Missing period at the end of the line

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+* Preparation - Each Aggregator converts each input share into an output share
+  compatible with the aggregation function. This computation involves the
+  aggregation parameter. In general, each aggregation parameter may result in a
+  different an output share.
+* Aggregation - Each Aggregator combines a sequence of output shares into its
+  aggregate share and sends the aggregate share to the Collector.
+* Unsharding - The Collector combines the aggregate shares into the aggregate
+  result.
+
+Sharding and Preparation are done once per measurement input.  Aggregation and
```
tgeoghegan, line 409: I think this sentence makes sense if we read "input" as a verb and not a noun, but it's a bit confusing given that those two words have specific meanings throughout VDAF and DAP. Maybe rephrase to only use one or the other word?

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+Unsharding are done over a batch of measurements (more precisely, over the
```
tgeoghegan, line 410: "Unsharding *is* done"

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+recovered output shares).
+
+A concrete DAF specifies an algorithm for the computation needed in each of
+these stages. The interface of each algorithm is defined in the remainder of
+this section. In addition, a concrete DAF defines the associated constants and
+tyeps enumerated in the following table.
```
tgeoghegan, line 416: s/tyeps/types

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+
+| Parameter          | Description              |
+|:-------------------|:-------------------------|
+| `SHARES`      | Number of input shares into which each measurement is sharded |
+| `Measurement` | Type of each measurement      |
+| `AggParam`    | Type of aggregation parameter |
+| `OutShare`    | Type of each output share     |
```
tgeoghegan, line 423: Reviewing this table of types and the existing one for VDAFs, it seems odd that we take the care of defining an `OutShare` type but then represent input shares as `Bytes` (e.g., `Vdaf.measurement_to_input_shares(public_param: PublicParam, input: Measurement) -> Vec[Bytes]`). Why not define `InShare`?

This wasn't introduced by this PR so feel free to punt with an issue or a TODO (or to disagree with me altogether)

draft-irtf-cfrg-vdaf.md
```diff
@@ -379,54 +379,264 @@ exception that Clients submitting measurements may be anonymous.
...
+
+    Collector
+    =========
+~~~~
+{: #unshard-flow title="Computation of the final aggregate result from aggregate
+shares."}
+
+> QUESTION Maybe the aggregation algorithms should be randomized in order to
+> allow the Aggregators (or the Collector) to add noise for differential
+> privacy. (See the security considerations of {{?I-D.draft-gpew-priv-ppm}}.) Or
```
tgeoghegan, line 545: update this reference to `draft-ietf-ppm-dap` (and presumably elsewhere in the document where `draft-gpew-priv-ppm` occurs)

poc/daf.sage
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        # Each Client shards its input into shares.
+        input_shares = Daf.measurement_to_input_shares(measurement)
+
+        for input_share in input_shares:
+            prep_test_vector["input_shares"].append(input_share.hex())
+
+        # Each Aggregator prepares its input share for aggregation.
+        outbound = []
```
tgeoghegan, line 96: I think this is adapted from the `run_vdaf` test/illustration, where `outbound` is used to simulate the leader gathering prepare messages from helpers and then broadcasting combined prepare messages back out. That step isn't needed in a DAF (right?), so I wonder if this could be rewritten to make it more clear that each aggregator is independent and doesn't interact with any other during `prepare`? That is, eliminate any value like `outbound` that is shared among aggregators (in this case, each aggregator is represented by one iteration of the `for j in range(Daf.SHARES)` loop).

draft-irtf-cfrg-vdaf.md
```diff
@@ -603,83 +795,32 @@ privacy).
...
-measurements in the batch."}
-
-For simplicity, we have written this algorithm and the unsharding algorithm
-below in "one-shot" form, where all shares for a batch are provided at the same
-time.  Some VDAFs may also support a "streaming" form, where shares are
-processed one at a time.
+The data flow for this stage is illustrated in {{aggregate-flow}}. Here again,
+we have the aggregation algorithm in a "one-shot" form, where all shares for a
+batch are provided at the same time. VDAFs typically also support a "streaming"
+form, where shares are processed one at a time.
```
tgeoghegan, line 808: I'm a bit confused about "one-shot" vs. "streaming". The distinction that I'm aware of between VDAFs is whether the aggregation parameter exists or not. If it does, then aggregation can't start until the collector proves the aggregation parameter. If it does not, then aggregators can aggregate shares immediately upon upload.

But I don't think that has anything to do with whether aggregation can be "streamed". Let's say we're running `poplar1`, and my aggregator receives 50 shares, and then the collector sends me `agg_param`. Now I can aggregate together those 50 shares, but I'm likely to do so by preparing them individually, independently from each other, summing each output share into an aggregate, which sounds like "streaming" to me. Further, my aggregator may continue to receive shares between when it receives `agg_param` (in a DAP `CollectReq`) and when it delivers the aggregate share (which will happen when the collector hits the DAP collect job URI), and I don't think anything stops the aggregator from continuing to incorporate shares into the aggregate -- or _stream_ them.

I think I know what you mean by "one-shot" vs. "streaming", but I don't think these are the right words to express it, since the issue here is when the aggregator can expect to have the aggregation parameter and not whether input shares can be aggregated incrementally.

draft-irtf-cfrg-vdaf.md
tgeoghegan: Oh, I think I see now that `Bytes` connotes that the output of the method is a message that can be transmitted on the wire to another protocol participant. I think that's inconsistent though: `AggShare` also has to be encoded to be transmitted from aggregators to the collector, but it gets a rich type definition here. My vote would be to use nice type aliases everywhere and then extend the definition of DAF/VDAF to include methods like `Daf.encode_input_share(InShare) -> Bytes` and `Daf.decode_input_share(Bytes) -> InShare`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -257,12 +257,28 @@ from the literature.
...
 
+Finally, perhaps the most complex aspect of schemes like Prio3 and Poplar1 is
+the process by which the client-generated measurements are prepared for
+aggregation. Because these constructions are based on secret sharing, the
+servers will be required to exchange some amount of information in order to
+verify the measurement is valid and can be aggregated. Depending on the
+construction, this process may require multiple round trips over the network.
+
+There are applications in which this verification step may not be necessary,
+e.g., when the client's software is run a trusted execution environmnet. To
```
schoppmp, line 268: environment

draft-irtf-cfrg-vdaf.md
divergentdave: Both `prio3` and `poplar1` support "streaming" aggregation in the sense used here, as both there aggregation operations are effectively vector addition in a vector space over a field. The distinction is that `out_shares_to_agg_share` as given here takes in all output shares at once, and returns an aggregate share, rather than processing output shares only one at a time, updating an accumulator, and returning an aggregate share at the end. For a strawman counterexample, consider a VDAF where `out_shares_to_agg_share` takes the median of numbers from each output share. This would be a non-streaming VDAF, as there's no efficient way to incrementally compute an aggregate share. Instead, the only way to compute the exact median is to keep the multiset of all output shares around until all output shares are received.

draft-irtf-cfrg-vdaf.md
schoppmp: +1, I would prefer a concrete type for the input shares.

schoppmp (changes requested):

draft-irtf-cfrg-vdaf.md
```diff
@@ -342,70 +359,284 @@ measurement process is as follows:
...
-function across a batch of inputs. This section specifies the interfaces of
-these algorithms as they would be exposed to applications.
+Like DAFs described in the previous section, a VDAF scheme is used to compute a
+particular aggregation function over a set of Client-generated measurements.
+Evaluation of a VDAF involves the same four stages as for DAFs: Sharding,
+Preparation, Aggregation, and Unsharding. However, the Preparation stage will
+require interaction among the Aggregators in order to facilitate verifiability
+of the computation's correctness. Accommodating this interaction will require
+syntactic changes, as well as a new Setup phase for distributing long-lived
+verification parameters.
```
schoppmp, line 605: > However, the Preparation stage will require interaction among the Aggregators in order to facilitate verifiability of the computation's correctness.

I think it makes sense to split up the preparation (as in, turning an input share into an aggregation share), and the verification.
The reason is that the outer protocol using a VDAF might want to "opportunistically" aggregate shares, in particular for Poplar, where there are many rounds of Preparation, Aggregation, and Unsharding. This would save communication rounds, since the verification on one layer (of the Poplar tree) can be done in parallel with computing the aggregation shares on the next layer. In case a share turns out to be invalid, the corresponding aggregation shares can just be subtracted from the aggregated result.

poc/daf.sage
cjpatton: Done

draft-irtf-cfrg-vdaf.md
tgeoghegan: Hmm, TIL the semantics of the word "semantics". Well, maybe the IETF has a style guide that someone can cite from to settle this. For now I say go with whichever you prefer.

draft-irtf-cfrg-vdaf.md
tgeoghegan: No, you're right, this is correct as-is.

draft-irtf-cfrg-vdaf.md
tgeoghegan: Maybe this is too big of a nomenclature issue to tackle in this particular PR

draft-irtf-cfrg-vdaf.md
cjpatton: To be clear:
* Prepare takes input shares to output shares; and
* Aggregation takes output shares to aggregate shares.

> [...] the verification on one layer (of the Poplar tree) can be done in parallel with computing the aggregation shares on the next layer. In case a share turns out to be invalid, the corresponding aggregation shares can just be subtracted from the aggregated result.

Very Interesting idea. The current text points to two reasons for not breaking Prepare abstraction this way:
1. Not every VDAF is compatible with this change. E.g., in Prio+ the output shares are only computed after multiple rounds of interaction.
2. Releasing unverified output shares early may lead to them being misused.

Thus, a change here would be significant and would need to be thought through. I'd suggest we discuss it in a new issue.

draft-irtf-cfrg-vdaf.md
cjpatton: It's arguably orthogonal given that "one-shot" is the current terminology (i.e., not introduced by this PR). How about adding a comment to #47 about fixing this?

draft-irtf-cfrg-vdaf.md
schoppmp: > Not every VDAF is compatible with this change. E.g., in Prio+ the output shares are only computed after multiple rounds of interaction.
I'm not asking for every VDAF to support this opportunistic aggregation, but instead to *allow* it in places where it's supported by both the VDAF and the outer aggregation protocol.

I opened https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/61 to continue this discussion.

schoppmp (approved):
I can't see the diff between my last review and the current state. LGTM assuming there are no substantial changes.

cjpatton:
Thanks @schoppmp, no substantial changes, if you want to inspect the diff you can do
```
git diff f606049..cjpatton/daf
```

Note that this inccludes changes after rebasing on `main`.

bifurcation (approved):

PR #56, opened 2022-05-06
gadget_poly_len should not include the wire seed arity

jbr:
Currently `gadget_poly_len` is the same as the total gadget chunk length, which is off by the wire seed length/arity. The tests passed because all of the tested circuits have only a single gadget and the wire seed is before the polynomial, so attempting to read off the end of the proof truncated back to the expected length. If there were two gadgets, this would read part of the second gadget's wire seed into the polynomial of the first


cjpatton (changes requested):
Thanks again @jbr! I merged a test that exercises this bug: https://github.com/cfrg/draft-irtf-cfrg-vdaf/commit/c1548a7c37a4e11eef9a7ba37c5af3c1bad6134b. Two asks:
1. Please rebase on `main` and make sure the fix is sufficient.
2. Please revise the commit message to use a verb, e..g, "poc: Fix bug in `FlpGeneric` triggered by multiple gadgets". Also, please add to the commit message a note about what the bug is.

jbr:
Confirmed manually that the tests failed as follows before this pr:
```
❯ make test
Parsing flp_generic.sage
sage common.sage
sage field.sage
sage flp.sage
sage flp_generic.sage
TestMultiGadget: test 0 failed: proof evaluation resulted in False; want True
sage prg.sage
sage vdaf.sage
sage vdaf_prio3.sage
```

and with this pr they pass:
```
❯ make test
Parsing flp_generic.sage
sage common.sage
sage field.sage
sage flp.sage
sage flp_generic.sage
sage prg.sage
sage vdaf.sage
sage vdaf_prio3.sage
```

Would you be interested in a PR that runs the POC tests in GitHub actions as CI?

cjpatton (approved):

cjpatton:
Sweet, thanks again.

PR #57, opened 2022-05-09
poc: Add VerifyParam type annotation to Prio3

jbr:
I'm not certain that this is correct, but it seems like `VerifyParam` isn't the `Vdaf` default of `None` for `Prio3`


cjpatton (changes requested):
Yes that's correct. Before merging, would you mind changing the commit message by replacing "VdafPrio3" with "Prio3"? "VdafPrio3" isn't a proper name of something.

By the way, I filed https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/59 to investigate where else way may have missed this.

cjpatton (approved):
❤️

Issue #84, opened 2022-06-09
Specify Poplar1 VDAF


tgeoghegan:
An upcoming draft-irtf-cfrg-vdaf should fully specify Poplar1.

PR #85, opened 2022-06-14
poc: Define syntax for IDFPs

cjpatton:
Partially addresses #84.


poc/idpf.sage
```diff
@@ -0,0 +1,57 @@
...
+    SHARES: Unsigned = None
+
+    # Bit length of valid input values (i.e., values of `alpha`).
+    BITS: Unsigned = None
+
+    # The finite field used to reprresent the inner nodes of the IDPF tree.
+    FieldInner: field.Field = None
+
+    # The finite field used to represent the leaf nodes of the IDPF tree.
+    FieldLeaf: field.Field = None
```
cjpatton, line 20: @schoppmp we had previously talked about allowing each level of the hierarchy to have its own field. I decided to just define two fields, one for the leaves and another for the inner nodes. That's sufficient to admit the construction on the Poplar paper. Do you think more flexibility will be desirable?

Removing this flexibility seems useful for keeping things simple.

poc/idpf.sage
```diff
@@ -0,0 +1,57 @@
...
+    # Evaluate an IDPF key at a given level of the tree and with the given set
+    # of prefixes.
+    #
+    # An error is raised if any element of `prefixes` is larger than or equal to
+    # `2^level` or if `level` greater than or equal to `BITS`.
+    @classmethod
+    def eval_inner(Idpf,
+                   key: Idpf.Key,
+                   level: Unsigned,
+                   prefixes: Vec[Unsigned]) -> Vec[Idpf.FieldInner]:
```
schoppmp, line 46: How do you plan to implement this? I'm asking because it might make sense to keep intermediate state for the prefixes (i.e., the inner nodes in the DPF evaluation tree). Alternatively, you can always evaluate from the root, but then the total runtime for evaluating all levels sequentially will be quadratic in the number of levels.

divergentdave (commented):
Here are some typo fixes:

poc/idpf.sage
```diff
@@ -0,0 +1,57 @@
...
+
+from __future__ import annotations
+from sagelib.common import Error, Unsigned
+import sagelib.field as field
+import sagelib.prg as prg
+
+# An Incremntal Distributed Point Function (IDPF).
+class Idpf:
+    # Number of keys generated by the IPDF-key generation algorithm.
+    SHARES: Unsigned = None
```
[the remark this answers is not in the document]
cjpatton: Yes. Of course for the Poplar IDPF, `SHARES == 2`.

poc/idpf.sage
cjpatton: I'm planning on translating Figure 11 more or less directly, including carving out an explicit `EvalNext()` function. In the spec we will certainly want to note that implementations will typically make use of this stateful functionality.

schoppmp: Does that mean that `eval_inner` with this signature will usually not be called, and instead people will call `eval_next`? Then why specify `eval_inner`?

schoppmp: > Perhaps it would make sense to just unify eval_inner() and eval_leaf() into one function that returns either field type, as in the paper. Thinking about the eventual Poplar1 spec, I think this would be a bit awkward.

I think it's fine to have two functions for the two fields (in C++ I used a single templated function, I guess generics would work in Rust as well).

> Under the hood, both eval_inner() and eval_leaf() will call eval_next() some number of times.

My point is that if you want to have a sequence of calls `eval_inner(key, 1, ...)`,  `eval_inner(key, 2, ...)`, ..., `eval_leaf(key, ...)`, the number of calls you'd have to make to `eval_next` would be quadratic in the number of levels if you don't save state between the `eval_inner` calls and pass it to the next call.

cjpatton: Right, but the question is whether to address this in the syntax or  leave to implementations. It sounds like you'd prefer the former?

I'm good either way. One thing to keep in mind is that the VDAF syntax isn't stateful, so `Poplar1.prep_init()` will take as input a set of candidate prefixes and will end up with the quadratic explosion. Thus we will need to make a note there that `Poplar1` SHOULD be implemented in a way that avoids this.

cjpatton: Remember that for the reference implementation we're after the clearest implementation, not necessarily the most efficient.

schoppmp: > Thus we will need to make a note there that Poplar1 SHOULD be implemented in a way that avoids this.

Can you elaborate on how you'd implement this optimized version?

I was thinking of passing the state into and out of the VDAF interface by appropriately defined input/output shares. In fact, in the C++ DPF library, I use an [`EvaluationContext`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.proto#L156=) object as the input to [`EvaluateNext`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.h#L307=)/[`EvaluateUntil`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.h#L302=), which wraps a [`DpfKey`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.proto#L129=) and additionally keeps state between iterations.

> Remember that for the reference implementation we're after the clearest implementation, not necessarily the most efficient.

Agreed, let's focus on simplicity and clarity over speed in the reference implementation. I just want to make sure that we don't rule out these efficient implementations via the syntax.

cjpatton: > Can you elaborate on how you'd implement this optimized version?
> 
> I was thinking of passing the state into and out of the VDAF interface by appropriately defined input/output shares. In fact, in the C++ DPF library, I use an [`EvaluationContext`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.proto#L156=) object as the input to [`EvaluateNext`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.h#L307=)/[`EvaluateUntil`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.h#L302=), which wraps a [`DpfKey`](https://github.com/google/distributed_point_functions/blob/0f7f758d2c67092c05a3b93ff4ae9ad00cca1487/dpf/distributed_point_function.proto#L129=) and additionally keeps state between iterations.

Yeah this is more or less what I would do in an efficient implementation of `Poplar1`. The [current API for libprio-rs](https://github.com/divviup/libprio-rs/blob/main/src/vdaf.rs#L164-L174) is too strict and will need to be changed in order to allow `prepare_init()` to to update this `EvaluationContext`.

cjpatton:
FYI, I pushed some minor changes to the syntax (https://github.com/cfrg/draft-irtf-cfrg-vdaf/commit/04f9850da721e90e71905aa67ba70ff0274af6af) that resulted from having implemented the Poplar IDPF.

poc/idpf.sage
schoppmp: Okay, as long as this can still be implemented while being standard-compliant, I'm fine with it.

PR #86, opened 2022-06-16
poc: Add `OUT_LEN` parameter to IDPF syntax

cjpatton:
Based on #85 (merge that first).

Specifies that the output type is a vector over a finite field. This
allows IDPFs to be specified that are more naturally integrated with the
Poplar secure sketching scheme where one of the programmed outputs is
`1` and the other is a random "authenticator" value.


poc/idpf.sage
```diff
@@ -11,9 +11,13 @@ class Idpf:
     # Number of keys generated by the IPDF-key generation algorithm.
     SHARES: Unsigned = None
 
-    # Bit length of valid input values (i.e., values of `alpha`).
+    # Bit length of valid input values (i.e., the length of `alpha` in bits).
     BITS: Unsigned = None
 
+    # The length of each output vector (i.e., the length of `beta_leaf` and each
+    # element of `beta_inner`).
+    OUT_LEN: Unsigned = None
```
schoppmp, line 19: Doesn't the output length depend on the number of prefixes? Or is this to support tuples as values?

cjpatton: This is to support tuples as values. I guess the comments need to be clearer ...

schoppmp: Maybe call it `VALUE_LEN` then, since it's the length of one output value, not the entire output?

schoppmp (approved):

poc/idpf.sage
cjpatton: Oh that's better.

cjpatton: Done.

PR #87, opened 2022-06-16
poc: Add `VALUE_LEN` parameter to IDPF syntax

cjpatton:
Specifies that the output type is a vector over a finite field. This
allows IDPFs to be defined that are more naturally integrated with the
Poplar secure sketching scheme.


cjpatton:
Replaces https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/86.

Issue #92, opened 2022-06-26
Pass the number of reports to agg_shares_to_result()


divergentdave:
Inspired by the discussion on divviup/libprio-rs#256, I propose that the number of reports should be added as an argument to `agg_shares_to_result()` for DAFs and VDAFs. This would be useful in Prio for any AFE where the encoding does not preserve the additive identity. (or doesn't map some other identity to the additive identity) In such cases, then the unsharding algorithm could cancel out whatever constant term got added by every client's input. Any application using VDAF should already know how many reports contributed to one aggregate, we just need to expose it back down to the VDAF layer.

Note that this would also enable computing an average "inside the VDAF" rather than computing a sum and leaving it to the application to compute the average from that. This is more trivial than remapping [-1, 1] to [0, 2^n], as in the example that motivated this.

While it would be technically possible for an application to reverse the [-1, 1] to [0, 2^n] mapping, if given just a raw sum from Prio and a priori knowledge of the number of reports (by dividing by 2^(n-1) and subtracting num_reports), I think it would be more user-friendly if we made it possible for the unsharding step to do so. The postprocessing necessary for the averaging use case is easy enough to write out by hand, this fixed-point with offset encoding is slightly more difficult, but generally, AFE encodings could include multiple such encoded numbers, or even nonlinear functions composed with affine remappings. (The Prio paper includes examples of calculating a product or geometric mean. This would be done by taking a logarithm of the input, and this logarithm would most likely need to be encoded as a fixed point number with an offset.) For this reason, I think we should provide this number of reports to `agg_shares_to_result()`, so that unsharding can be "as powerful as" sharding, and applications won't have to concern themselves as much with the encoding details of particular more complicated VDAFs.

cc @MxmUrw

divergentdave:
Sure, I can write one up later this week

schoppmp:
Sounds good, no objection from my side.

cjpatton:
Closed by #95.

PR #95, opened 2022-06-29
Pass num_measurements when computing AggResult, add decode methods

divergentdave:
This implements the change proposed by #92, passing `num_measurements` to `agg_shares_to_result()`. For Prio3, I decided to thread this through both the FLP and validity circuit layers, in new `decode()` methods. With this change, `encode()`, `truncate()`, and `decode()` now give us a more complete API for Prio's AFEs. To test this, I wrote a quick Average type that makes use of the new argument. I also took advantage of the new flexibility in `Prio3.AggResult` to change the output types of Prio3Aes128Count and Prio3Aes128Sum to be `Unsigned`, as it's more intuitive and ergonomic. (If we don't like that last bit, it's separated out in the commit history already)


poc/vdaf_prio3.sage
```diff
@@ -371,28 +371,42 @@ class Prio3Aes128Histogram(Prio3):
...
+    A Prio3 instantiation to test use of num_measurements in the Valid
+    class's decode() method.
+    '''
+
+    @classmethod
+    def with_bits(cls, bits: Unsigned):
+        new_cls = deepcopy(cls).with_prg(prg.PrgAes128)
+        new_cls.Flp = flp_generic.FlpGeneric \
+            .with_valid(flp_generic.Average.with_bits(bits))
+        return new_cls
```
cjpatton, line 385: Does this actually need to be re-implemented?

cjpatton (commented):
Thanks @divergentdave! I massively appreciate your thoroughness and attention to detail here. Adding test code to exercise the new logic is particularly appreciated.

I will give this a thorough review next Thursday, 7/7, when I'm back from PTO. In the meantime, a few high-level things:
* I'm a bit weary of the reference implementation defining the VDAF `Prio3Aes128Average` without also defining it in the draft itself. That said it's definitely useful to exercise the API changes that are introduced in this change. We should either add it it to the draft or rename it to something like `TestPrio3Aes128Average` (to suggest that it's there for testing, but may eventually go in the draft).
* Don't worry about updating the change log. We'll wait to do so until we're ready to cut the next draft.

poc/vdaf_prio3.sage
divergentdave: I re-implemented it to swap out `flp_generic.Sum` for `flp_generic.Average`. I could possibly define some new members on `Prio3Aes128Sum`, and and override a member for the validity circuit instead, but I figure the separate factory classmethod is less invasive.

poc/daf.sage
```diff
@@ -138,7 +141,7 @@ class TestDaf(Daf):
             reduce(lambda x, y: [x[0] + y[0]], out_shares))
 
     @classmethod
-    def agg_shares_to_result(cls, _agg_param, agg_shares):
+    def agg_shares_to_result(cls, _agg_param, agg_shares, num_measurements):
```
cjpatton, line 144: nit: Since this parameter gets ignored here, prefix it with `_`.

poc/flp_generic.sage
```diff
@@ -518,6 +541,10 @@ class Histogram(Valid):
     def truncate(Histogram, inp):
         return inp
 
+    @classmethod
+    def decode(Histogram, output, _num_measurements):
+        return [x.as_unsigned() for x in output]
```
cjpatton, line 546: A more descriptive name than `x` would be useful. How about `bucket_count`?

poc/flp_generic.sage
```diff
@@ -607,6 +634,18 @@ def test_flp_generic(cls, test_cases):
             print('{}: test {} failed: proof evaluation resulted in {}; want {}'.format(
                 cls.Valid.__name__, i, decision, expected_decision))
 
+
+class TestAverage(Sum):
+    # Associated types
+    AggResult = float
```
cjpatton, line 640: I'm not sure we should have `float` as the result type, since the measurements are `Unsigned` and thus have a different precisison. What about defining this aggregation function as `floor(average(measurements))` and letting `AggResult = Unsigned`.

cjpatton (changes requested):
Excellent work! Just a few minor comments and this'll be ready to merge.

cjpatton (approved):

PR #97, opened 2022-07-09
Complete spec of Poplar1

### LINKED ISSUE #44: Poplar1: `k_verify_rand` should be derived from agg param
Currently only the nonce is used to derive the shared verification randomness. This should also depend on the aggregation parameter so that the verification randomness is unique for each aggregation parameter.

### LINKED ISSUE #84: Specify Poplar1 VDAF

cjpatton:
Closes #84.
Closes #44.

Completes the specification of Poplar1, including the reference
implementation and test vectors.

* poc: Add additional tests for IdpfPoplar.
* poc: Modify Idpf syntax so that the size of the keys is defined. This
  is needed for decoding the keysfrom the Poplar1 input share.
* poc: Fix a minor bug in IdpfPoplar that caused `eval()` to throw an
  error when it shouldn't.
* poc: Have `eval()` check that the prefixes are unique. This is needed
  for Poplar1, which checks for one-hotness of the output shares.
* poc: Fix various typos.
* poc: Implement Poplar1 and add code for generating test vectors.
* Add new field `Field255`, as required for IdpfPoplar.
* Refactor Prg interface to align with reference implementation.
* Re-generate test vectors and add test vectors for Poplar1Aes128.


divergentdave (approved):
Looks good, I left remarks with a variety of small cleanups. I'm excited!

draft-irtf-cfrg-vdaf.md
```diff
@@ -2275,281 +2281,654 @@ the protocol works as follows.
...
-Each `beta[l]` is a vector of elements of a finite field. We distinguish two types of
-fields: one for inner nodes (denoted `Idpf.FieldInner`), and one for leaf nodes (`Idpf.FieldLeaf`).
+A concrete IDPF defines the types and constants enumerated in {{idpf-param}}. In
+the remainder we write `Idpf.Vec` as shorthand for the type
+`Union[Vec[Vec[Idpf.FieldInner]], Vec[Vec[Idpf.FieldLeaf]]]`. (This type denotes
+either a vector of inner node field elements or leaf node field elements.) The
+scheme is comprised of the following algorithms:
 
-An IDPF is comprised of the following algorithms :
+* `Idpf.gen(alpha: Unsigned, beta_inner: Vec[Vec[Idpf.FieldInner]], beta_leaf:
```
bifurcation, line 2340: `alpha` is the index being sharded here, right?  If so, it would be clearer to call it a bit string, not an integer, since that's how it is referred to above.  (Maybe internally to the function it is converted to an integer, but that doesn't really matter as long as all participants put the bits in the same order.)

Where do the `beta` values come from?  If they're computed from `alpha`, it seems like there's a missing step for that.  If they're constant / defined by the IDPF, then it seems like they can be treated as internal to the function, not arguments.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2275,281 +2281,654 @@ the protocol works as follows.
...
 
 Finally, an implementation note. The interface for IDPFs specified here is
 stateless, in the sense that there is no state carried between IDPF evaluations.
 This is to align the IDPF syntax with the VDAF abstraction boundary, which does
 not include shared state across across VDAF evaluations. In practice, of course,
 it will often be beneficial to expose a stateful API for IDPFs and carry the
-state across evaluations.
+state across evaluations. See {{idpf-poplar}} for details.
+
+<br>
```
bifurcation, line 2382:
```suggestion
```

bifurcation (approved):
Couple of high-level notes, otherwise LGTM

poc/test_vec/01/Prio3Aes128Count_0.json
```diff
@@ -1,8 +1,6 @@
 {
     "agg_param": null,
-    "agg_result": [
-        1
-    ],
+    "agg_result": 1,
```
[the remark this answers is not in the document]
cjpatton: Right now the plan is for the test vectors to match a tag of the repo. For example, test_vec/01 will matches `draft-irtf-cfrg-vdaf-01`. Moreover, I plan to remove test vectors from old versions from the repo as changes are made. (See #100). Thus these "in-between" changes to the test vectors don't matter much for interop.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2275,281 +2281,652 @@ the protocol works as follows.
...
+  Aggregator's unique identifier, the public share distributed to all of the
+  Aggregators, the Aggregator's IDPF key, the "level" at which to evaluate the
+  IDPF, and the sequence of candidate prefixes. It returns the share of the
+  value corresponding to each candidate prefix.
 
-A concrete IDPF specifies a single associated constant:
+  The type the output depends on the value of `level`: If `level < Idpf.BITS-1`,
+  the output is the value for an inner node, which has type
+  `Vec[Idpf.FieldInner]`; otherwise, if `level == BITS-1`, then the output is
+  the value for a leaf node, which has type `Vec[Idpf.FieldLeaf]`.
```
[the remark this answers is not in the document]
cjpatton: Yes, nice catch.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2275,281 +2281,654 @@ the protocol works as follows.
...
+    public_share = IdpfPoplar.encode_public_share(correction_words)
+    return (public_share, init_seed)
+~~~
+{: #idpf-poplar-gen title="IDPF-key generation algorithm of IdpfPoplar."}
+
+### Key Evaluation
+
+> TODO Describe in prose how IDPF-key evaluation algorithm works.
+
+The description of the IDPF-evaluation generation algorithm makes use
```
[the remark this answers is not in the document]
cjpatton: ❤️

draft-irtf-cfrg-vdaf.md
cjpatton: Done, here and elsewhere.

draft-irtf-cfrg-vdaf.md
cjpatton: > `alpha` is the index being sharded here, right? If so, it would be clearer to call it a bit string, not an integer, since that's how it is referred to above. (Maybe internally to the function it is converted to an integer, but that doesn't really matter as long as all participants put the bits in the same order.)

@schoppmp and I discussed this a bit and decided that `Unsigned` was the best representation of the input. A couple of alternatives:
* `Bytes` - This is the most intuitive, however it has the problem of not being able to represent a bit strings whose length is not a multiple of 8. Of course we could consider requiring input strings be a multiple 8. This would probably be fine for most applications.
* Define some sort of `BitString` type, perhaps even making use of ASN.1. We decided this would be a bit too complicated.
 
> Where do the `beta` values come from? If they're computed from `alpha`, it seems like there's a missing step for that. If they're constant / defined by the IDPF, then it seems like they can be treated as internal to the function, not arguments.

The `beta` values are arbitrary and are picked by the caller. In `Poplar1`, they are chosen in a particular way.

bifurcation: On the first point: I'm not overly picky here.  Maybe you could just say somewhere that the Unsigned value is the input bit string of length `BITS`, interpreted as a `BITS`-bit big-endian integer (or whatever, doesn't really matter).

On the second point: If they're (a) generated unilaterally by the caller and (b) not used elsewhere, they should get generated internally to `gen`, not passed in as arguments.

cjpatton: > On the first point: I'm not overly picky here. Maybe you could just say somewhere that the Unsigned value is the input bit string of length `BITS`, interpreted as a `BITS`-bit big-endian integer (or whatever, doesn't really matter).

Yeah we should provide some guidance here, but at this point I'm not sure what to recommend here. Really any encoding is fine, so long as the application is consistent.


> On the second point: If they're (a) generated unilaterally by the caller and (b) not used elsewhere, they should get generated internally to `gen`, not passed in as arguments.

Well, one could imagine different applications for IDPFs that call for different values to be "programmed". For Poplar1, the programmed values are pairs `(1, k)` where `k` is a random field element used for verifying the output shares. When used in a plain DAF, there would be no need for the random field element. One could also imagine an application in which "weighted" inputs are used, i.e., rather than always output `1`, output some integer in range `[0, c)` chosen by the user.

Another thing to note is that the current IDPF syntax more or less matches the paper.

Issue #102, opened 2022-07-28
Review: Henry Corrigan-Gibbs


cjpatton:
@henrycg reviewed draft 02 and sent his feedback privately via an email. 

- [ ] In the introduction, point out that VDAF can provide DP in addition to MPC-style security. (See also #94.)
- [ ] Section 6.1.2: "Large subgroup": Say exactly how large the group needs to be. (2^20-ish is going to be sufficient, I think.)
- [ ] Section 7.2.2: There is an attack on the soundness of Prio3 stemming from a bug in [the paper](https://eprint.iacr.org/2019/188). The paper has been updated with a patch.
- [ ] Section 7.2.2: For Fiat-Shamir to be secure we may need a larger or field or to re-run the proof check multiple times. More investigation is needed.
- [ ] Section 8: We should flag early on in this section that, when used multiple times, Poplar1 has leakage (via prefix hit counts) that other VDAFs don't necessarily suffer from in the same way.
- [ ] SEction 8: Discuss hashing construction from Appendix B.

cjpatton:
All tasks for VDAF-03 are complete. What remains are editorial changes and additional security considerations for Poplar1.

cjpatton:
Remaining work is purely editorial; let's try to land by 05.

Issue #32, opened 2022-02-25
PRG: Fixed-key AES


cjpatton:
Our initial PRG construction is based on AES: https://github.com/cjpatton/vdaf/pull/28. @schoppmp points out that there may alternative AES-based constructions that are better suited to our application. In addition, there may be alternative primitives, like an XOF hash function, that are well suited for VDAFs. This issue is for tracking this discussion.

cjpatton:
Here's the paper Phillipp referenced: https://eprint.iacr.org/2019/074.pdf

martinthomson:
The text you added to the draft suggests that fixed-key AES (as it is called in the paper) might be something to move to.  The paper suggests that you don't get correlation robustness in that case, so maybe that isn't the best choice.

A PRG based on AES does seem to be a sensible choice here from a performance perspective if nothing else, and the idea of using a counter seems reasonable.  Maybe consider using $\mathsf{MMO}^\pi$ or $\widehat{\mathsf{MMO}}_\sigma^\pi$ from that paper instead.  The added cost is modest, but it would appear that the properties they provide are superior.

martinthomson:
Oh, I should add https://eprint.iacr.org/2022/080 describes another variant that adds a tweak.  The tweak looks like $H(m,t) = \pi(m\otimes t)\xor (m\otimes t)$, which is a little slower overall, but probably still in line with what you are looking for.  The only trick there is in deciding the tweak.

wangxiao1254:
I was contacted by Phillipp @schoppmp about this issue and we had some discussion. Briefly, letting each client pick a uniform AES key seems the best option. The underlying reasoning would be based on https://eprint.iacr.org/2019/1168.pdf, MiTCCR, in the ideal cipher model. The loss of concrete security would be minimal.

cjpatton:
Thanks @wangxiao1254 for your feedback and contribution to the design of VDAF!

PR #104, opened 2022-08-15
Define codepoints for (V)DAFs

cjpatton:
Partially addresses #102.

Each (V)DAF defines a 32-bit codepoint used to uniquely identify it. The
current scheme is as follows:

* The first 3 nibbles are used to identify the VDAF "class", e.g.,
  Prio3Aes128 or Poplar1Aes128.
* The remaining 5 nibbles are reserved for particular "instances",
  e.g., Prio3Aes128Count, Prio3Aes128Sum, etc.

While at it, incorporate the codepoints into domain separation tags
(DSTs) in Prio3 and Poplar1. This is needed primarily for Prio3. (See
issue#102.)

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -472,6 +472,7 @@ types enumerated in the following table.
 
 | Parameter          | Description              |
 |:-------------------|:-------------------------|
+| `ID`          | Algorithm identifier for this DAF. |
 | `SHARES`      | Number of input shares into which each measurement is sharded |
 | `Measurement` | Type of each measurement      |
 | `AggParam`    | Type of aggregation parameter |
@@ -489,6 +490,9 @@ values must be transmitted between parties over a network.
 > OPEN ISSUE It might be cleaner to define a type for each value, then have that
 > type implement an encoding where necessary. This way each method parameter has
 > a meaningful type hint. See issue#58.
+
+Each DAF is identified by a unique, 32-bit integer `ID`. Identifiers for each
+(V)DAF specified in this document are defined in {{codepoints}}.
 
 ## Sharding {#sec-daf-shard}
 
@@ -703,6 +707,7 @@ listed in {{vdaf-param}} are defined by each concrete VDAF.
 
 | Parameter         | Description              |
 |:------------------|:-------------------------|
+| `ID`              | Algorithm identifier for this VDAF. |
 | `VERIFY_KEY_SIZE` | Size (in bytes) of the verification key ({{sec-vdaf-prepare}}) |
 | `ROUNDS`          | Number of rounds of communication during the Preparation stage ({{sec-vdaf-prepare}}) |
 | `SHARES`          | Number of input shares into which each measurement is sharded ({{sec-vdaf-shard}}) |
@@ -719,6 +724,9 @@ other quantities are given a concrete type.
 
 > OPEN ISSUE It might be cleaner to define a type for each value, then have that
 > type implement an encoding where necessary. See issue#58.
+
+Each VDAF is identified by a unique, 32-bit integer `ID`. Identifiers for each
+(V)DAF specified in this document are defined in {{codepoints}}.
 
 ## Sharding {#sec-vdaf-shard}
 
@@ -1345,7 +1353,7 @@ For some FLPs, the encoded input also includes redundant field elements that are
 useful for checking the proof, but which are not needed after the proof has been
 checked. An example is the "integer sum" data type from {{CGB17}} in which an
 integer in range `[0, 2^k)` is encoded as a vector of `k` field elements (this
-type is also defined in {{prio3-sum}}). After consuming this vector,
+type is also defined in {{prio3aes128sum}}). After consuming this vector,
 all that is needed is the integer it represents. Thus the FLP defines an
 algorithm for truncating the input to the length of the aggregated output:
 
@@ -1415,7 +1423,7 @@ are described in {{prio3-helper-functions}}.
 ~~~
 def measurement_to_input_shares(Prio3, measurement):
     # Domain separation tag for PRG info string
-    dst = VERSION + b' prio3'
+    dst = VERSION + I2OSP(Prio3.ID, 4)
     inp = Prio3.Flp.encode(measurement)
     k_joint_rand = zeros(Prio3.Prg.SEED_SIZE)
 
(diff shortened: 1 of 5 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -3020,8 +3029,21 @@ removing it after unsharding.
 
 # IANA Considerations
 
-This document makes no request of IANA.
+| Value      | Scheme               | Type | Reference |
+|:-----------|:---------------------|:-----|:----------|
+| 0x00000000 | reserved for testing | n/a  | n/a       |
```
tgeoghegan, line 3034: I think it'd be a good idea to reserve everything with the prefix `0x000` for testing, just in case someone has a test scenario that requires multiple fake (V)DAFs interacting somehow.

cjpatton: We can probably do that. Once concern I have is that 32 bits might be too large. Typically codepoints are just 8-16 bits, in which case we wouldn't be able to afford allocating lots of space.

tgeoghegan: Then reserving `0x000000XX` should suffice.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3020,8 +3029,21 @@ removing it after unsharding.
...
+| 0x00100002 | Prio3Aes128Histogram | VDAF | {{prio3aes128histogram}} |
+| 0x00200000 | Poplar1Aes128        | VDAF | {{poplar1aes128}} |
+{: #codepoints title="Unique identifers for (V)DAFs."}
 
+> TODO Add IANA considerations for the codepoints summarized in {{codepoints}}.
+
+> OPEN ISSUE Currently the scheme includes the PRG. This means that we need bits
+> of the codepoint to differentiate between PRGs. We could instead make the PRG
+> generic (e.g., Prio3Count(Aes128) instead of Prio3Aes128Count) and define a
+> separate codepoint for.
```
chris-wood, line 3046: I would leave it as-is for now. I don't know see any reason why an application would want to be opinionated about the PRG, so it makes sense to me that we pick one or maybe two variants that work well for all.

chris-wood, line 3046: I would leave it as-is for now. I don't know see any reason why an application would want to be opinionated about the PRG, so it makes sense to me that we pick one or maybe two variants that work well for all.

chris-wood (approved):
This seems good enough for now. We can always change the shape of these later on. One question I might consider asking is whether or not you want the draft version to be part of the DST, if it's not already, just in case things change over time.

cjpatton:
> This seems good enough for now. We can always change the shape of these later on. One question I might consider asking is whether or not you want the draft version to be part of the DST, if it's not already, just in case things change over time.

The version (literally `b'vdaf-02'` is the prefix of the DST :)

chris-wood:
🦆

draft-irtf-cfrg-vdaf.md
cjpatton: I think there are a couple potential reasons for this:
1. It's useful to be able to reuse existing primitives whenever possible. (For instance, @simon-friedberger found himself in the position of having to re-implement `PrgAes128` from primitives in NSS.) This suggests to me that there may eventually be a preference to build something from, say, HMAC. (HKDF is not particularly well suited here, given its relatively short output length.)
2. In #32 we're discussing an alternative AES-based PRG. One might have a preference to stick with what we have today. (I hope not, though.)

In any case, we can cross this bridge later.

draft-irtf-cfrg-vdaf.md
cjpatton: Done

draft-irtf-cfrg-vdaf.md
```diff
@@ -3020,8 +3029,24 @@ removing it after unsharding.
 
 # IANA Considerations
 
-This document makes no request of IANA.
+A codepoint for each (V)DAF in this document is defined in the table below. Note
```
bifurcation, line 3032: You probably want to structure this section a little differently.  You're providing instructions to IANA: Create X, Y, Z, registries; add entries to Q registry; etc.  So I would have a top-level summary here, and then a subsection for each registry you're creating / touching.  You should be able to basically copy/paste the prose you need from MLS:

https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-iana-considerations

draft-irtf-cfrg-vdaf.md
```diff
@@ -3020,8 +3029,24 @@ removing it after unsharding.
...
+that `0x00000000` through `0x000000FF` are reserved for testing purposes.
 
+| Value      | Scheme               | Type | Reference |
+|:-----------|:---------------------|:-----|:----------|
+| `0x000000XX` | reserved for testing | n/a  | n/a       |
```
bifurcation, line 3037: The more common way to specify test code points is as a range, so `0x00000000 - 0x000000ff` here (as in the prose above).

Also, usually a reservation like this is made not just for testing, but for "vendor specific usage" -- whatever you want to do for your own products that you don't expect to generally interop.

bifurcation (commented):
This seems fine, but could be aligned better with the usual patterns for IANA language.

I don't love the division of the algorithm identifier into high- and low-order nibbles.  Why not just have two identifiers here, say a 2-byte VDAF class and a 4-byte algorithm identifier within that class?  There crypto things we're dealing with here are so gigantic that a couple of bytes doesn't seem likely to be noticeable.

draft-irtf-cfrg-vdaf.md
cjpatton: Good catch. I'm also going to bump the range to 0x00000000 - 0x0000FFFF so that there's more room for "vendor specific usage".

draft-irtf-cfrg-vdaf.md
cjpatton: Thanks, @bifurcation! The immediate need for codepoints is to address #102. I filed #110 to address this later on.

bifurcation:
If you don't think that this variant pattern is going to be common for other VDAFs, then yeah, we shouldn't structure the code points around it.  But at the same time, you're not going to have 2^32 Prio variants, so I would suggest being a little more restrained, since (a) you're not going to have 2^20 Prio variants and (b) you can always grab more space later.  Maybe something like:

* 0x00000000 - 0x000000FF - Prio variants
* 0x00000100 - Poplar
* 0xFFFF0000 - 0xFFFFFFFF reserved for private use

(It is traditional to put reserved space at the high end, see, e.g., [TLS ciphersuites](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4).)

PR #105, opened 2022-08-16
Prio3: Align spec with revised paper

cjpatton:
Partially addresses #102.

A bug was found in [BBCGGI+19] that leads to an attack on the robustness
of Prio3. The attack is based on an observation in Appendix A of the
paper "A New Paradigm for Collision-Free Hashing: Incrementality at
Reduced Cost" from Bellare and Micciancio (Eurocrypt 1997). In short,
the attack allows a malicious Client to construct invalid input shares
for which the Aggregators would compute a `k_joint_rand` of its
choosing.

This bug was patched in Section 6.2.3 of
https://eprint.iacr.org/archive/2019/188/20220727:184451 as follows:
Instead of XORing joint randomness shares computed by the
Aggregators, `k_joint_rand` is computed by hashing the shares. This
change patches the Prio3 spec in kind.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -283,7 +283,7 @@ security considerations for VDAFs.
 
 ## Change Log
 
-(*) Indicates a change that breaks wire compatibility with the previous draft.
+(\*) Indicates a change that breaks wire compatibility with the previous draft.
 
 02:
 
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -355,6 +355,9 @@ Some common functionalities:
 * `byte(int: Unsigned) -> Bytes` returns the representation of `int` as a byte
   string. The value of `int` MUST be in `[0,256)`.
 
+* `concat(parts: Vec[Bytes]) -> Bytes` returns the concatenation of the input
+  byte strings, i.e., `parts[0] || ... || parts[len(parts)-1]`.
+
 * `xor(left: Bytes, right: Bytes) -> Bytes` returns the bitwise XOR of `left`
   and `right`. An exception is raised if the inputs are not the same length.
 
@@ -1161,6 +1164,11 @@ def expand_into_vec(Prg,
 > OPEN ISSUE Phillipp points out that a fixed-key mode of AES may be more
 > performant (https://eprint.iacr.org/2019/074.pdf). See issue#32.
 
+> TODO(issue #106) Decide if it's safe to model this construction as a random
+> oracle. `PrgAes128.derive_seed()` is used for the Fiat-Shamir heuristic in
+> Prio3 ({{prio3}}). A fixed-key is used for this step (the all-zero string). A
+> reasoanble starting point would be to model AES as an ideal cipher.
+
 Our first construction, `PrgAes128`, converts a blockcipher, namely AES-128,
 into a PRG. Seed expansion involves two steps. In the first step, CMAC
 {{!RFC4493}} is applied to the seed and info string to get a fresh key. In the
@@ -1416,22 +1424,22 @@ shares (called the "leader" shares below) are vectors of field elements. The
 other shares (called the "helper" shares) are represented instead by PRG seeds,
 which are expanded into vectors of field elements.
 
-The code refers to a pair of auxiliary functions for encoding the shares. These
-are called `encode_leader_share` and `encode_helper_share` respectively and they
-are described in {{prio3-helper-functions}}.
+The code refers to auxiliary functions for encoding the shares and deriving the
+joint randomness. In particular, `encode_leader_share` and
+`encode_helper_share` are used for encoding and `joint_rand` is used for joint
+randomness. These are defined in {{prio3-helper-functions}}.
 
 ~~~
 def measurement_to_input_shares(Prio3, measurement):
     # Domain separation tag for PRG info string
     dst = VERSION + I2OSP(Prio3.ID, 4)
     inp = Prio3.Flp.encode(measurement)
-    k_joint_rand = zeros(Prio3.Prg.SEED_SIZE)
 
     # Generate input shares.
     leader_input_share = inp
     k_helper_input_shares = []
     k_helper_blinds = []
-    k_helper_hints = []
+    k_joint_rand_parts = []
     for j in range(Prio3.SHARES-1):
         k_blind = gen_rand(Prio3.Prg.SEED_SIZE)
         k_share = gen_rand(Prio3.Prg.SEED_SIZE)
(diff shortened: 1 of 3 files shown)


poc/vdaf_prio3.sage
```diff
@@ -205,6 +215,35 @@ class Prio3(Vdaf):
...
+    # #102.)
+    #
+    # For instance, consider hashing the XOR of the joint randomness parts:
+    #
+    #   k_joint_rand = Prio3.Prg.derive_seed(
+    #       xor(k_joint_rand_parts)
+    #       dst + byte(255))
+    #
+    # This allows us to set the hint to the XOR of the other Aggregator's parts,
+    # as we did in VDAF-02.
```
divergentdave, line 239: I don't think this could work, as it would fall prey to the same attack. The attacker would still be able to choose an output value for the XOR operation, and then hash it before deriving the joint randomness elements it generates. Then, they could do the same Gaussian reduction to select shares that would result in the chosen XOR output. In effect, this change would just change the random oracle from expand_into_vec to derive_seed composed with expand_into_vec. We need to get rid of the linearity of the XOR operation entirely, as that's what gives the attacker an advantage.

One idea for an alternative scheme, with lower overhead: The client includes the joint randomness seed in each report, computed as H(...H(H(H(share 1 || blinding nonce 1) || share 2 || blinding nonce 2) || share 3 || blinding nonce 3) ... || share n || blinding nonce n). The client also sends hash i-1 and hash i from this chain of hashes (if i-1 is in bounds) as part of each report i. Then, the aggregators use the joint randomness seed in their calculation, confirm that hash i == H(hash i-1 || share i || blinding nonce n) based on the view of its report share, and broadcast hash i-1 and hash i from its report share, along with the joint randomness seed they used. During preprocessing, the leader verifies that each aggregator reported that its check of the next hash evaluation was valid on its report share, verifies that all consecutive aggregators agree on their views of successive hash values, and that each aggregator's view of the joint randomness seed was equal to the final hash value. This would only require sending 2 or 3 seeds in each direction, rather than n-1. (This could probably be optimized a bit more, but it follows the guideline of hashing everything to produce the public coin)

cjpatton: This seems like it would definitely improve communication overhead for large n, but in most applications I think we're still expecting n to be small, say 2 or 3.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1161,6 +1164,11 @@ def expand_into_vec(Prg,
 > OPEN ISSUE Phillipp points out that a fixed-key mode of AES may be more
 > performant (https://eprint.iacr.org/2019/074.pdf). See issue#32.
 
+> TODO(issue #106) Decide if it's safe to model this construction as a random
+> oracle. `PrgAes128.derive_seed()` is used for the Fiat-Shamir heuristic in
+> Prio3 ({{prio3}}). A fixed-key is used for this step (the all-zero string). A
+> reasoanble starting point would be to model AES as an ideal cipher.
```
cjpatton, line 1170: Filed https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/106.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1557,12 +1566,15 @@ def prep_init(Prio3, verify_key, agg_id, _agg_param,
...
-    joint_rand, k_joint_rand, k_joint_rand_share = [], None, None
+    joint_rand, k_joint_rand, k_joint_rand_part = [], None, None
     if Prio3.Flp.JOINT_RAND_LEN > 0:
         encoded = Prio3.Flp.Field.encode_vec(input_share)
-        k_joint_rand_share = Prio3.Prg.derive_seed(
+        k_joint_rand_part = Prio3.Prg.derive_seed(
             k_blind, dst + byte(agg_id) + encoded)
-        k_joint_rand = xor(k_hint, k_joint_rand_share)
+        k_joint_rand_parts = k_hint[:agg_id] + \
+                             [k_joint_rand_part] + \
```
tgeoghegan, line 1575: What do the enclosing `[]` denote on this line? `k_joint_rand_part` is already a `Bytes` (`def derive_seed(Prg, seed: Bytes, info: Bytes) -> bytes`), so can't it already be concatenated with the elements of `k_hint`?

draft-irtf-cfrg-vdaf.md
```diff
@@ -1643,6 +1658,12 @@ def agg_shares_to_result(Prio3, _agg_param, agg_shares,
 ### Auxiliary Functions {#prio3-helper-functions}
 
 ~~~
+def joint_rand(Prio3, k_joint_rand_parts):
```
tgeoghegan, line 1661: This isn't really an issue with this PR, but is there a reason these function declarations can't have type annotations on their arguments, like we do for `derive_seed` or `expand_into_vec` in the PRG section? It would be a big help to readers trying to understand if `k_joint_rand_parts` is a `Bytes` (all parts already concatenated) or a `[Bytes]` (each element is a part) or some other thing.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1497,11 +1505,15 @@ def measurement_to_input_shares(Prio3, measurement):
         k_leader_hint,
     ))
     for j in range(Prio3.SHARES-1):
+        # Each Helper's hint consists of the joint randomness part of the
+        # other Aggregators.
+        k_helper_hint = k_joint_rand_parts[:j+1] + \
+                        k_joint_rand_parts[j+2:]
```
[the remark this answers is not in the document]
tgeoghegan: Oh, no, never mind, Python's slice operator doesn't care:
```
Python 3.10.6 (main, Aug  2 2022, 00:00:00) [GCC 12.1.1 20220507 (Red Hat 12.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> the_list = [1, 2, 3]
>>> the_list[100:]
[]
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1444,22 +1452,19 @@ def measurement_to_input_shares(Prio3, measurement):
...
                                      helper_input_share)
         encoded = Prio3.Flp.Field.encode_vec(helper_input_share)
-        k_hint = Prio3.Prg.derive_seed(
+        k_joint_rand_part = Prio3.Prg.derive_seed(
             k_blind, dst + byte(j+1) + encoded)
-        k_joint_rand = xor(k_joint_rand, k_hint)
         k_helper_input_shares.append(k_share)
         k_helper_blinds.append(k_blind)
-        k_helper_hints.append(k_hint)
+        k_joint_rand_parts.append(k_joint_rand_part)
```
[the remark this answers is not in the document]
divergentdave: The list's insert method shifts all later list elements to higher indices, so we won't lose the first helper's contribution.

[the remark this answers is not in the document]
tgeoghegan: Oh, once again, I have misunderstood Python list semantics: `list.insert(0, val)` will put the new value at 0 and push existing values to the right.

draft-irtf-cfrg-vdaf.md
divergentdave: The plus operators here are acting as list concatenation -- the first and last arguments are slices of the k_hint list, so they are both lists of Bytes. The middle argument is a list literal, consisting of just the `k_join_rand_part` Bytes object.

draft-irtf-cfrg-vdaf.md
cjpatton: There's no rhyme or reason for this. I'm happy to add type hints here.

poc/vdaf_prio3.sage
```diff
@@ -3,8 +3,8 @@
 from copy import deepcopy
 from typing import Tuple
 from sagelib.common import ERR_DECODE, ERR_INPUT, ERR_VERIFY, I2OSP, VERSION, \
-                           TEST_VECTOR, Bytes, Unsigned, Vec, byte, gen_rand, \
-                           vec_add, vec_sub, xor, zeros
+                           TEST_VECTOR, Bytes, Unsigned, Vec, byte, concat, \
+                           gen_rand, vec_add, vec_sub, xor, zeros
```
divergentdave, line 7: Nit: this import is unused now
```suggestion
                           gen_rand, vec_add, vec_sub, zeros
```

divergentdave (approved):

poc/vdaf_prio3.sage
cjpatton: Done!

PR #117, opened 2022-09-14
Add nonces to VDAF sharding syntax & Prio3

### LINKED ISSUE #119: Review: Hannah Davis: Potential for joint randomness repetition in Prio3
Prio3 uses a variant of the Fiat-Shamir transform to derive random coins for its underlying proof system from the inputs to each aggregator. However, this derivation doesn't include the nonce, which means that a malicious client can submit the same set of input shares multiple times under different nonces and they will be processed with the same joint randomness. This introduces a threat to robustness. As a toy example, there is a degenerate FLP that is sound with L bits of security when each set of input shares is submitted at most once, but is vulnerable to an 2^(L/2)-query attack if they can be submitted under many nonces.

hannahdaviscrypto:
Closes #119.

I came across a concern about the robustness of Prio3.The construction uses a PRG to derive joint randomness from the input shares that is used by the client and all aggregators. If a client uploads the same set of input shares twice under distinct nonces, the aggregators will derive the same joint randomness in both preparation phases.Intuitively, this violates the expectation that the joint randomness is pseudorandom, or at least not completely controllable by an attacker. 
More formally, there is a degenerate FLP that is sound  with L bits of security when each set of input shares is submitted at most once, but is vulnerable to an 2^(L/2)-query attack if they can be submitted under many nonces. 

This PR mitigates the attack by including the nonce in the joint randomness derivation so that if input shares are submitted under two different nonces, they will go through preparation with independent joint randomness. It requires changing VDAF syntax so that Shard takes the nonce as input. There shouldn't be any change to communication as a result, and the operational cost to Prio3 is at worst 1 additional AES call per party.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -754,16 +754,18 @@ Each VDAF is identified by a unique, 32-bit integer `ID`. Identifiers for each
 
 ## Sharding {#sec-vdaf-shard}
 
-Sharding is syntactically identical to DAFs (cf. {{sec-daf-shard}}):
+Sharding transforms a measurement into input shares as it does in DAFs
+(cf. {{sec-daf-shard}}); in addition, it takes a nonce as input and
+produces a public share:
 
-* `Vdaf.measurement_to_input_shares(measurement: Measurement) -> (Bytes,
-  Vec[Bytes])` is the randomized input-distribution algorithm run by each
-  Client. It consumes the measurement and produces a public share, distributed
-  to each of Aggregators, and the corresponding sequence of input shares, one
-  for each Aggregator. Depending on the VDAF, the input shares may encode
-  additional information used to verify the recovered output shares (e.g., the
-  "proof shares" in Prio3 {{prio3}}). The length of the output vector MUST be
-  `SHARES`.
+* `Vdaf.measurement_to_input_shares(measurement: Measurement, nonce: Bytes) ->
+  (Bytes, Vec[Bytes])` is the randomized input-distribution algorithm run by
+  each Client. It consumes the measurement and the nonce and produces a public
+  share, distributed to each of Aggregators, and the corresponding sequence of
+  input shares, one for each Aggregator. Depending on the VDAF, the input shares
+  may encode additional information used to verify the recovered output shares
+  (e.g., the"proof shares" in Prio3 {{prio3}}). The length of the output vector
+  MUST be `SHARES`.
 
 ## Preparation {#sec-vdaf-prepare}
 
@@ -939,7 +941,7 @@ def run_vdaf(Vdaf,
     for (nonce, measurement) in zip(nonces, measurements):
         # Each Client shards its measurement into input shares.
         (public_share, input_shares) = \
-            Vdaf.measurement_to_input_shares(measurement)
+            Vdaf.measurement_to_input_shares(measurement, nonce)
 
         # Each Aggregator initializes its preparation state.
         prep_states = []
@@ -1440,7 +1442,7 @@ The input-distribution algorithm involves the following steps:
 
 1. Encode the Client's raw measurement as an input for the FLP
 1. Shard the input into a sequence of input shares
-1. Derive the joint randomness from the input shares
+1. Derive the joint randomness from the input shares and nonce
 1. Run the FLP proof-generation algorithm using the derived joint randomness
 1. Shard the proof into a sequence of proof shares
 
@@ -1455,7 +1457,7 @@ joint randomness. In particular, `encode_leader_share` and
 randomness. These are defined in {{prio3-helper-functions}}.
 
 ~~~
-def measurement_to_input_shares(Prio3, measurement):
+def measurement_to_input_shares(Prio3, measurement, nonce):
     # Domain separation tag for PRG info string
     dst = VERSION + I2OSP(Prio3.ID, 4)
     inp = Prio3.Flp.encode(measurement)
@@ -1478,14 +1480,14 @@ def measurement_to_input_shares(Prio3, measurement):
                                      helper_input_share)
         encoded = Prio3.Flp.Field.encode_vec(helper_input_share)
         k_joint_rand_part = Prio3.Prg.derive_seed(
-            k_blind, dst + byte(j+1) + encoded)
+            k_blind, dst + byte(j+1) + nonce + encoded)
         k_helper_input_shares.append(k_share)
         k_helper_blinds.append(k_blind)
         k_joint_rand_parts.append(k_joint_rand_part)
     k_leader_blind = gen_rand(Prio3.Prg.SEED_SIZE)
     encoded = Prio3.Flp.Field.encode_vec(leader_input_share)
     k_leader_joint_rand_part = Prio3.Prg.derive_seed(
-        k_leader_blind, dst + byte(0) + encoded)
+        k_leader_blind, dst + byte(0) + nonce + encoded)
     k_joint_rand_parts.insert(0, k_leader_joint_rand_part)
 
     # Compute joint randomness seed.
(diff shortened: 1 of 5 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -293,6 +293,16 @@ security considerations for VDAFs.
...
 
+04:
+* Change Prio3 spec to include the nonce in joint randomness derivation. This 
+  mitigates a toy attack on robustness where a client submits the same set of 
+  input shares under two distinct nonces and exploits the use of the same joint 
+  randomness in both preparation phases. The mitigated attack has the same query 
+  complexity as a birthday attack.
+
+* Change VDAF syntax to include a nonce as an input in the sharding phase. This
+  accommodates the above change to Prio3.
```
cjpatton, line 304: You can remove the change log update; we typically wait to do this until the next draft is cut.

FWIW, I would summarize the Prio3 as follows

> Prio3: Add the nonce to the transcript for joint randomness computation. This improves the provable security bound for robustness.

draft-irtf-cfrg-vdaf.md
```diff
@@ -705,6 +715,8 @@ Preparation, Aggregation, and Unsharding. However, the Preparation stage will
 require interaction among the Aggregators in order to facilitate verifiability
 of the computation's correctness. Accommodating this interaction will require
 syntactic changes.
+One change is that in a VDAF, each client measurement is identified by a unique
+nonce provided by the environment (`nonce`, see {{run-vdaf}}).
```
cjpatton, line 719: I don't think this is all that useful here. 
```suggestion
```

cjpatton (changes requested):
Thanks Hannah! This PR will need a couple more things before it can be merged:
1. The definition of Poplar1 needs to be updated to pass `nonce` to `measurement_to_input_shares()`. It's fine that it ignores it, but we want to keep the API boundaries consistent nonetheless.
2. The reference implementation needs to be updated to match. (See the `poc/` folder.) When making a change like this, typically the first step is to update the reference implementation, make sure the tests pass, then copy the updated definitions into the draft.

poc/common.sage
```diff
@@ -8,10 +8,10 @@ import struct
 
 # If set, then test vectors will be generated. A fixed source of randomness is
 # used for `gen_rand()`.
-TEST_VECTOR = False
+TEST_VECTOR = True
```
cjpatton, line 11: Let's revert this, as `TEST_VECTOR == True` is only needed for generating test vectors.

poc/common.sage
```diff
@@ -8,10 +8,10 @@ import struct
...
 
 # Document version, reved with each draft that contains breaking changes.
-DRAFT = '03'
+DRAFT = '04'
```
cjpatton, line 14: Let's revert this. Bumping the version is usually a chore saved until just before cutting the draft.

poc/test_vec/04/Poplar1Aes128_0.json
```diff
@@ -0,0 +1,54 @@
+{
```
cjpatton, line 1: There might be other breaking changes for VDAF-04, so we'll wait to update the test vectors until just before cutting.

poc/vdaf.sage
```diff
@@ -44,7 +44,8 @@ class Vdaf:
     # one for each Aggregator. This method is run by the Client.
     @classmethod
     def measurement_to_input_shares(Vdaf,
-                                    measurement: Measurement) -> (Bytes,
+                                    measurement: Measurement,
+                                    nonce: Bytes) -> (Bytes,
                                                                   Vec[Bytes]):
```
cjpatton, line 49: The spacing here is a bit odd. 
```suggestion
                                    nonce: Bytes) -> (Bytes, Vec[Bytes]):
```

poc/vdaf_poplar1.sage
```diff
@@ -34,7 +34,7 @@ class Poplar1(Vdaf):
     AggResult = Vec[Unsigned]
 
     @classmethod
-    def measurement_to_input_shares(Poplar1, measurement):
+    def measurement_to_input_shares(Poplar1, measurement, nonce):
```
cjpatton, line 37: nit: In Python (i.e., Sage :D) it's conventional to prefix ignored parameters with an underscore.
```suggestion
    def measurement_to_input_shares(Poplar1, measurement, _nonce):
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -754,11 +754,14 @@ Each VDAF is identified by a unique, 32-bit integer `ID`. Identifiers for each
 
 ## Sharding {#sec-vdaf-shard}
 
-Sharding is syntactically identical to DAFs (cf. {{sec-daf-shard}}):
-
-* `Vdaf.measurement_to_input_shares(measurement: Measurement) -> (Bytes,
-  Vec[Bytes])` is the randomized input-distribution algorithm run by each
-  Client. It consumes the measurement and produces a public share, distributed
+Sharding transforms a measurement into input shares as it does in DAFs 
+(cf. {{sec-daf-shard}}); in VDAFs it additionally takes a nonce as input and
```
cjpatton, line 758:
```suggestion
(cf. {{sec-daf-shard}}); in addition, it takes a nonce as input and 
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -2510,7 +2513,7 @@ follows. Function `encode_input_shares` is defined in {{poplar1-helper-functions
 
 
 ~~~
-def measurement_to_input_shares(Poplar1, measurement):
+def measurement_to_input_shares(Poplar1, measurement, nonce):
```
cjpatton, line 2516:
```suggestion
def measurement_to_input_shares(Poplar1, measurement, _nonce):
```

poc/common.sage
cjpatton, line 11: Let's revert this. We only want to set `TEST_VECTOR == True` when we want to generate the test vectors; otherwise, the tests use truly random bits, which is what we want.

draft-irtf-cfrg-vdaf.md
```diff
@@ -754,11 +754,14 @@ Each VDAF is identified by a unique, 32-bit integer `ID`. Identifiers for each
...
-* `Vdaf.measurement_to_input_shares(measurement: Measurement) -> (Bytes,
-  Vec[Bytes])` is the randomized input-distribution algorithm run by each
-  Client. It consumes the measurement and produces a public share, distributed
+Sharding transforms a measurement into input shares as it does in DAFs 
+(cf. {{sec-daf-shard}}); in VDAFs it additionally takes a nonce as input and 
+produces a public share:
+
+* `Vdaf.measurement_to_input_shares(measurement: Measurement, nonce: Bytes) -> 
+  (Bytes, Vec[Bytes])` is the randomized input-distribution algorithm run by each
+  Client. It consumes the measurement and the nonce and
```
cjpatton, line 763: * `make` fails if a line ends with trailing white space:
  ```
  $ make
  757:Sharding transforms a measurement into input shares as it does in DAFs
  758:(cf. {{sec-daf-shard}}); in VDAFs it additionally takes a nonce as input and
  761:* `Vdaf.measurement_to_input_shares(measurement: Measurement, nonce: Bytes) ->
  763:  Client. It consumes the measurement and the nonce and
  draft-irtf-cfrg-vdaf.md contains trailing whitespace
  *** Run 'make fix-lint' to automatically fix some errors
  make: *** [lint-whitespace] Error 1
  ```

* Also, lines should break at 80 characters.

```suggestion
  Client. It consumes the measurement and the nonce and
```

cjpatton (changes requested):

poc/common.sage
hannahdaviscrypto: Done

cjpatton:
Looking good! However, CI is still complaining about spurious whitespace: https://github.com/cfrg/draft-irtf-cfrg-vdaf/actions/runs/3063497529/jobs/4945718835

cjpatton:
Looking good! One minor nit, then we'll need to "squash" the commits on this branch into a single one. This can be done with an "interactive rebase":

```
$ git rebase main -i
```

This will pull up your editor and give you instructions for how to carry out the rebase. You will "pick" the first commit, then "squash" each subsequent commit into the first. You will have a chance to edit the commit title and message: Make sure the title is succinct (the title of this PR works perfectly fine) and that the message is complete. By "complete" I mean that it should be evident to the reader what has changed and why.

Also, this PR includes an orthogonal change to poc/README.md that fixes some typos. It's great to pick up these kinds of changes in PRs. However, it's important that the commit history reflects all changes, even minimal ones. There are two ways we can address this:
1. (ideal) Pull the change out into a separate commit
2. (perfectly fine, especially for a spec) Note in the commit message that the commit also includes minor editorial changes.

Finally, once the rebase is complete, you will have to "force-push" your branch. Be careful because this rewrites history. (git provides a backup if needed.)

```
$ git push -f
```

cjpatton (approved):
Nice work!

PR #120, opened 2022-09-28
Improved verification scheme for Poplar1

hannahdaviscrypto:


poc/idpf_poplar.sage
```diff
@@ -227,7 +231,7 @@ class IdpfPoplar(Idpf):
 if __name__ == '__main__':
     cls = IdpfPoplar \
                 .with_prg(prg.PrgAes128) \
-                .with_value_len(2)
+                .with_value_len(3)
```
cjpatton, line 234: Following my suggestion above, here we would have something like

```python
    cls = IdpfPoplar \
                .with_prg(prg.PrgAes128) \
                .with_inner_value_type(Field64, Field64, Field2) \
                .with_outer_value_type(Field255, Field255, Field2)
```

poc/vdaf_poplar1.sage
```diff
@@ -20,184 +21,183 @@ class Poplar1(Vdaf):
...
     OutShare = Union[Vec[Vec[Idpf.FieldInner]],
                      Vec[Vec[Idpf.FieldLeaf]]]
     AggResult = Vec[Unsigned]
 
     @classmethod
     def measurement_to_input_shares(Poplar1, measurement):
         dst = VERSION + I2OSP(Poplar1.ID, 4)
-        prg = Poplar1.Idpf.Prg(
-            gen_rand(Poplar1.Idpf.Prg.SEED_SIZE), dst + byte(255))
+        corr_seed = gen_rand(Poplar1.Idpf.Prg.SEED_SIZE)
```
cjpatton, line 40: Perhaps we should rename `corr_seed`? `auth_seed`, maybe?

poc/vdaf_poplar1.sage
```diff
@@ -20,184 +21,183 @@ class Poplar1(Vdaf):
...
-            if level < Poplar1.Idpf.BITS - 1:
-                corr_inner[0] += corr0
-                corr_inner[1] += corr1
+            k = auth[level]
+            print('randomness\t:', k)
+            (leader_data_share, leader_auth_share, leader_indicator_share) = \
+                Poplar1.Idpf.eval(0, public_share, keys[0], level, [prefix])[0]
+            helper_auth_share = leader_auth_share -k
+            helper_data_share = leader_data_share - Field(1)
+            w = Poplar1.hash1(level, prefix, \
```
cjpatton, line 79: Isn't the correction word `w` supposed to be the XOR of the hashes, not the concatenation?

poc/vdaf_poplar1.sage
```diff
@@ -20,184 +21,183 @@ class Poplar1(Vdaf):
...
-        sketch_share = [a_share, b_share, c_share]
+
+        # Evaluate the IDPF key at the given set of prefixes.
+        value = Poplar1.Idpf.eval(agg_id,
+                                  idpf_public_share,
+                                  key,
+                                  level,
+                                  prefixes)
+
+        #Leader expands randomness
```
cjpatton, line 117:
```suggestion
        # Leader expands randomness
```

poc/vdaf_poplar1.sage
```diff
@@ -20,184 +21,183 @@ class Poplar1(Vdaf):
...
+            (data_share, auth_share, indicator_share) = value[i]
+            z = Field(1) if agg_id == 0 else Field(-1)
+            corrected_x = Poplar1.hash1(level,
+                                        prefixes[i],
+                                        [z * data_share, z * auth_share])
+            result = Poplar1.correct(corrected_x,
+                                     indicator_share,
+                                     correction_word)
+            corr_results.append(result)
+            data_sum -=  z * data_share
```
cjpatton, line 144:
```suggestion
            data_sum -= z * data_share
```

poc/vdaf_poplar1.sage
```diff
@@ -20,184 +21,183 @@ class Poplar1(Vdaf):
...
+        if status == b'init':
+            (level, agg_id, out_share, check_bound, corr_check, bound_msg) = \
+                                                                    prep_state
+            new_prep = (b'finish',(agg_id, check_bound, out_share))
+            msg = corr_check + bound_msg
+            return (new_prep, msg)
+        if status == b'finish':
+            (agg_id, check_bound, out_share) = prep_state
+            # Helper (only) verifies boundedness
+            if (agg_id == 0):
```
cjpatton, line 178: Unnecessary paranetheses (here and below)
```suggestion
            if agg_id == 0:
```

cjpatton (commented):
This looks to be mostly correct to me. A couple of high level things:

1. Looks like this only supports checking that the non-zero element of the vector is in range [0, 1). In fact, the new scheme supports [0, k) for any small k. I think we should incorporate that generalization here. (Remember that the hash sequence sent to the peer needs to be shifted by a random amount.)

2. Tests are failing on my machine:
```
$ make && sage vdaf_poplar1.sage
make: Nothing to be done for `pyfiles'.
randomness	: 14474163779119285356
randomness	: 23710317869799874109156653167373759793599482990239257754186445210723389039029
auth_sum:	 14474163779119285356
auth_sum:	 0
inbound:	 95eec14105e696c84525532a65df37fd533d9ce30896df1db60bc754fb26d8de
check_bound:	 95eec14105e696c84525532a65df37fd
Traceback (most recent call last):
  File "/Users/chris/github.com/cfrg/draft-irtf-cfrg-vdaf/poc/vdaf_poplar1.sage.py", line 350, in <module>
    test_vdaf(Poplar1Aes128.with_bits(_sage_const_2 ),
  File "/Users/chris/github.com/cfrg/draft-irtf-cfrg-vdaf/poc/sagelib/vdaf.py", line 373, in test_vdaf
    agg_result = run_vdaf(Vdaf,
  File "/Users/chris/github.com/cfrg/draft-irtf-cfrg-vdaf/poc/sagelib/vdaf.py", line 173, in run_vdaf
    out = Vdaf.prep_next(prep_states[j], inbound)
  File "/Users/chris/github.com/cfrg/draft-irtf-cfrg-vdaf/poc/vdaf_poplar1.sage.py", line 173, in prep_next
    raise ERR_VERIFY #Input is one-hot but the entry is not 1
sagelib.common.Error: verification of the user's input failed
```

poc/vdaf_poplar1.sage
hannahdaviscrypto: It's not the XOR or the concatenation, it's the sum (we're working with additive shares not boolean hence the change from Mike's writeup)

poc/vdaf_poplar1.sage
```diff
@@ -11,198 +11,206 @@ from sagelib.vdaf import Vdaf, test_vdaf
...
+    Prep = Tuple[Bytes, Union[Tuple[Vec[Bytes], Unsigned, OutShare],
+                              Tuple[Unsigned, Bytes, OutShare]]]
     AggResult = Vec[Unsigned]
 
+    # Poplar1 specific parameters
+    # maximum number of times a prefix may be counted per string
+    BOUND = 1
+
     @classmethod
     def measurement_to_input_shares(Poplar1, measurement):
```
cjpatton, line 40: We likely ahve use cases where we would want to add weight to the output share

poc/vdaf_poplar1.sage
```diff
@@ -11,198 +11,206 @@ from sagelib.vdaf import Vdaf, test_vdaf
...
+        # and output share
+        # sum over all prefixes for boundedness verification
         out_share = []
-        for (i, r) in enumerate(verify_rand):
-            (data_share, auth_share) = value[i]
-            sketch_share[0] += data_share * r
-            sketch_share[1] += data_share * r^2
-            sketch_share[2] += auth_share * r
+        data_sum = Field(0)
+        auth_sum = Field(0)
```
cjpatton, line 130: Is the Field type is not necessarily the same for data and auth?

poc/vdaf_poplar1.sage
```diff
@@ -312,6 +324,38 @@ class Poplar1(Vdaf):
...
         return 'bits'
 
+    @classmethod
+    def correct(Poplar1, corrected_x, b, correction_word):
+        if b == Field2(0):
+            return corrected_x
+        return correction_word - corrected_x
+
+    @classmethod
+    def hash1(Poplar1, level, x, y):
```
cjpatton, line 334: Please relabel parameter to make this easier to read
```suggestion
    def hash1(Poplar1, level, prefix, elems):
```

poc/vdaf_poplar1.sage
```diff
@@ -11,198 +11,206 @@ from sagelib.vdaf import Vdaf, test_vdaf
...
+            (agg_id, (check_bound, corr_check, bound_msg), out_share) \
+                                                                    = prep_state
+            new_prep = (b'finish',(agg_id, check_bound, out_share))
+            msg = corr_check + bound_msg
+            return (new_prep, msg)
+        if status == b'finish':
+            (agg_id, check_bound, out_share) = prep_state
+            l = Poplar1.Idpf.Prg.SEED_SIZE
+            possibilities = []
+            for i in range(Poplar1.BOUND+1):
```
cjpatton, line 186: I think the upper bound should be strict. That is, the range check should assert that the value is in range [0..k).

poc/vdaf_poplar1.sage
```diff
@@ -11,198 +11,206 @@ from sagelib.vdaf import Vdaf, test_vdaf
...
+        if agg_id == 0:
+            offset = OS2IP(gen_rand(Poplar1.BOUND)) % (Poplar1.BOUND + 1)
+            for i in range(0, offset):
+                bound_msg_part_2 += Poplar1.hash2(Field(i) + data_sum,
+                                            Field(i) * auth[level] + auth_sum,
+                                            level)
+            for i in range(offset, Poplar1.BOUND+1):
+                bound_msg_part_1 += Poplar1.hash2(Field(i) + data_sum,
+                                            Field(i) * auth[level] + auth_sum,
+                                            level)
```
cjpatton, line 161: Can we de-duplicate this code?

poc/vdaf_poplar1.sage
```diff
@@ -11,198 +11,206 @@ from sagelib.vdaf import Vdaf, test_vdaf
...
+        bound_msg = bound_msg_part_1 + bound_msg_part_2 if agg_id == 0 else b''
+
+        check_bound = b'' if agg_id == 0 \
+                    else Poplar1.hash2(data_sum, auth_sum, level)
```
cjpatton, line 165: Doesn't `hash2()` need to be a random hash function unknown to the attacker? I.e., don't we want to pass in something dervied from `verify_key` here?

poc/idpf_poplar.sage
```diff
@@ -216,18 +222,20 @@ class IdpfPoplar(Idpf):
...
-    def with_value_len(IdpfPoplar, value_len: Unsigned):
-        if value_len == 0:
-            raise ERR_INPUT # value length must be positive
+    def with_value_type(IdpfPoplar, value_inner, value_leaf):
+        if value_inner == None or value_leaf == None:
+            raise ERR_INPUT # values must both be defined
         new_cls = deepcopy(IdpfPoplar)
-        new_cls.VALUE_LEN = value_len
+        new_cls.InnerValue = value_inner
+        new_cls.LeafValue = value_leaf
```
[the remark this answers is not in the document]
hannahdaviscrypto: That doesn't seem necessary to me. The two lengths need to be the same in our Poplar1 scheme, but only in the sense that they both need to equal 4 and it would seem too specific to assert that. I don't see a reason, especially in the leaf-is-k-bounded and inner-vals-are-1-bounded world that we couldn't have the leaves include multiple group elements, each of which are k-bounded.

poc/vdaf_poplar1.sage
hannahdaviscrypto: It should be the same type because auth is supposed to equal delta * (data) and this operation is only defined if they're the same type.

poc/vdaf_poplar1.sage
hannahdaviscrypto: I don't think it does? Hash_2 needs to be a random oracle for Privacy, in which case the attacker is Bob who knows verify_key (and the hash isn't private anyway), and for soundness, where the attacker genuinely has no reason to know H, we only require collision resistance.

hannahdaviscrypto:
The protocol that this pull request was designed to implement has privacy issues; closing it while we develop an alternative.

Issue #106, opened 2022-08-16
Securely instantiating the PRG


cjpatton:
In Prio3, `PrgAes128.derive_seed()` is used with a fixed seed for the Fiat-Shamir heuristic. We need to decide if this is safe. It would be sufficient to prove, say, that this function is indifferentiable from a random oracle when modeling AES as an ideal cipher.

Note that we might end up picking a PRG in #32 that is already safe a safe choice here, in which case we should just replace `PrgAes128`.

cjpatton:
TODO:
- [ ] Add Prg variant based on SHA-3 (SHAKE)
- [ ] Change PrgAes128 to use some fixed-key "thing"

cjpatton:
https://eprint.iacr.org/2019/074.pdf
https://eprint.iacr.org/2022/1431.pdf

cjpatton:
Per @simon-friedberger's request, I've updated the benchmark branch of libprio with a PRG based on HKDF-SHA256. I played around with a few variants, all are roughly as performant:

| PRG | 100 | 1000 | 10,000 | 
| - | - | - | - | 
| PrgHkdfSha256 | 465.88 us | n/a | n/a |

(Note that input sizes of 1000 and 10,000 exceed the the maximum output length for HKDF-SHA256.)

The performance is quite bad compared to the other algorithms. I'm not sure if we're using a software implementation of SHA-256, but if we are, these numbers aren't too surprising given how much hashing is involved in HKDF.

cjpatton:
@divergentdave and I have been working on Poplar1 in libprio: https://github.com/divviup/libprio-rs/pull/434

We're still working on benchmarks, but so far it looks like SHA-3 is up to 40% more expensive than `PrgAes128`. This confirms @schoppmp's hypothesis that Prio is arithmetic-bound and Poplar is PRG-bound.

cjpatton:
We have merged the basic functionality for Poplar1 into libprio (🎉). I have also updated my PRG prototypes branch (https://github.com/divviup/libprio-rs/compare/cjpatton/prg-prototypes) with benchmarks with Poplar1 using the current `PrgAes128` and the alternative `PrgSha3`. The table below summarizes shard and preparation timings for both variants and for various bit lengths.

| PRG | 16 bits | 64 bits | 128 bits | 256 |
|-|-|-|-|-|
| PrgAes128 shard time | 49.4 us | 143.8 us | 270.6 us | 519.5 us |
[...]

Issue #114, opened 2022-09-12
Note: in Prio3, clients duplicate hints


hannahdaviscrypto:
In Prio3, the client generates a "hint" for each aggregator to help derive the joint randomness. These hints need to be communicated to all aggregators in the start of preparation, so the sharding algorithm encodes the entire list into each input share. If there are s aggregators, the client's message length will be quadratic in s: s hints * s input shares. This seems fine for s=2, but if applications where s > 2 are targeted, we may want to consider other encodings.

hannahdaviscrypto:
A possible alternative for s>2: VDAF already supports a "public share" that the leader broadcasts to all aggregators, and Prio3 currently leaves this empty. We could have clients encode the hints in the public share so they are only sent once by the client. If authentication of the hints is needed, the client could encode some part of its joint randomness within each input share as a confirmation, which would preserve linear communcation  in s between the client and leader.

hannahdaviscrypto:
As I step away from my internship project, these are the two outstanding proposed changes for Prio3:

1. Put the joint randomness hints into the public share instead of duplicating them in the aggregators' input shares, for the reasons described above. Additional authentication is unnecessary because in DAP, the public share is authenticated by HPKE.
2. Follow every domain separation tag in a PRG call with a 1-byte domain separation value between 249 and 255. The existing domain separation tag performs domain separation between Prio3 and external uses of its PRG. In contrast, the new value separates domains between internal uses of the PRG in different parts of the scheme,. e.g., for generating joint vs query randomness. This value performs domain separation between the different uses of the PRG.

cjpatton:
Solved by #121.

PR #121, opened 2022-10-06
[poc] Prio3 improvements

hannahdaviscrypto:
In Prio3, a single PRG is used to derive random coins for several different purposes. These are differentiated using a domain separation string "dst" and a single byte. This byte is set to the aggregator index when deriving joint randomness parts, and to 255 for joint randomness seeds and query randomness, but it is omitted for joint randomness and prover randomness. I propose to include this byte in every PRG call, and to set it to 254 for joint randomness, and 253 for query and prover randomness. This will help in the eventual modeling of the PRG as a random oracle (as in Issue [#106](https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/106).

### DIFF
--- a/poc/vdaf_prio3.sage
+++ b/poc/vdaf_prio3.sage
@@ -10,6 +10,13 @@ import sagelib.flp as flp
 import sagelib.flp_generic as flp_generic
 import sagelib.prg as prg
 
+DST_MEASUREMENT_SHARE = byte(1)
+DST_PROOF_SHARE = byte(2)
+DST_JOINT_RANDOMNESS = byte(3)
+DST_PROVE_RANDOMNESS = byte(4)
+DST_QUERY_RANDOMNESS = byte(5)
+DST_JOINT_RAND_SEED = byte(6)
+DST_JOINT_RAND_PART = byte(7)
 
 # Base class for VDAFs based on prio3.
 class Prio3(Vdaf):
@@ -35,32 +42,32 @@ class Prio3(Vdaf):
         dst = VERSION + I2OSP(Prio3.ID, 4)
         inp = Prio3.Flp.encode(measurement)
 
-        # Generate input shares.
-        leader_input_share = inp
-        k_helper_input_shares = []
+        # Generate measurement shares.
+        leader_measurement_share = inp
+        k_helper_measurement_shares = []
         k_helper_blinds = []
         k_joint_rand_parts = []
         for j in range(Prio3.SHARES-1):
             k_blind = gen_rand(Prio3.Prg.SEED_SIZE)
             k_share = gen_rand(Prio3.Prg.SEED_SIZE)
-            helper_input_share = Prio3.Prg.expand_into_vec(
+            helper_measurement_share = Prio3.Prg.expand_into_vec(
                 Prio3.Flp.Field,
                 k_share,
-                dst + byte(j+1),
+                dst + DST_MEASUREMENT_SHARE + byte(j+1),
                 Prio3.Flp.INPUT_LEN
             )
-            leader_input_share = vec_sub(leader_input_share,
-                                         helper_input_share)
-            encoded = Prio3.Flp.Field.encode_vec(helper_input_share)
+            leader_measurement_share = vec_sub(leader_measurement_share,
+                                               helper_measurement_share)
+            encoded = Prio3.Flp.Field.encode_vec(helper_measurement_share)
             k_joint_rand_part = Prio3.Prg.derive_seed(
-                k_blind, dst + byte(j+1) + nonce + encoded)
-            k_helper_input_shares.append(k_share)
+                k_blind, dst + DST_JOINT_RAND_HINT + byte(j+1) + nonce + encoded)
+            k_helper_measurement_shares.append(k_share)
             k_helper_blinds.append(k_blind)
             k_joint_rand_parts.append(k_joint_rand_part)
         k_leader_blind = gen_rand(Prio3.Prg.SEED_SIZE)
-        encoded = Prio3.Flp.Field.encode_vec(leader_input_share)
+        encoded = Prio3.Flp.Field.encode_vec(leader_measurement_share)
         k_leader_joint_rand_part = Prio3.Prg.derive_seed(
-            k_leader_blind, dst + byte(0) + nonce + encoded)
+            k_leader_blind, dst + DST_JOINT_RAND_HINT + byte(0) + nonce + encoded)
         k_joint_rand_parts.insert(0, k_leader_joint_rand_part)
 
         # Compute joint randomness seed.
@@ -70,13 +77,13 @@ class Prio3(Vdaf):
         prove_rand = Prio3.Prg.expand_into_vec(
             Prio3.Flp.Field,
             gen_rand(Prio3.Prg.SEED_SIZE),
-            dst,
+            dst + DST_PROVER_RANDOMNESS,
             Prio3.Flp.PROVE_RAND_LEN
         )
         joint_rand = Prio3.Prg.expand_into_vec(
             Prio3.Flp.Field,
             k_joint_rand,
-            dst,
+            dst + DST_JOINT_RANDOMNESS,
             Prio3.Flp.JOINT_RAND_LEN
         )
         proof = Prio3.Flp.prove(inp, prove_rand, joint_rand)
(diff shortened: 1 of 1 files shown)


hannahdaviscrypto:
Two more Prio3 changes added to this pull request: (1) a minor terminology change to distinguish between the aggregators' input shares and the measurement shares that the input shares contain, and (2) a moving joint randomness hints to the public share, which decreases total communication costs whenever there are more than two aggregators. (2) was discussed in Issue #114 along with the above domain separation change.

poc/vdaf_prio3.sage
```diff
@@ -35,32 +35,32 @@ class Prio3(Vdaf):
...
                 k_share,
-                dst + byte(j+1),
+                dst + byte(254) + byte(j+1),
                 Prio3.Flp.INPUT_LEN
             )
-            leader_input_share = vec_sub(leader_input_share,
-                                         helper_input_share)
-            encoded = Prio3.Flp.Field.encode_vec(helper_input_share)
+            leader_measurement_share = vec_sub(leader_measurement_share,
+                                         helper_measurement_share)
```
cjpatton, line 53: nit: Column alignment can improve readability
```suggestion
                                               helper_measurement_share)
```

poc/vdaf_prio3.sage
```diff
@@ -88,42 +88,38 @@ class Prio3(Vdaf):
...
 
     @classmethod
     def prep_init(Prio3, verify_key, agg_id, _agg_param,
-                  nonce, _public_share, input_share):
+                  nonce, public_share, input_share):
         # Domain separation tag for PRG info string
         dst = VERSION + I2OSP(Prio3.ID, 4)
 
-        (input_share, proof_share, k_blind, k_hint) = \
+        k_hint = Prio3.decode_public_share(public_share, agg_id)
```
cjpatton, line 121: For consistency with the sharding algorithm. Alternatively we could use `k_hint` there instead of `k_joint_rand_parts`.
```suggestion
        k_joint_rand_parts = Prio3.decode_public_share(public_share, agg_id)
```

poc/vdaf_prio3.sage
```diff
@@ -278,23 +266,41 @@ class Prio3(Vdaf):
...
+                            k_hint):
+        encoded = Bytes()
+        if Prio3.Flp.JOINT_RAND_LEN > 0:
+            encoded += concat(k_hint)
+        return encoded
+
+    @classmethod
+    def decode_public_share(Prio3, encoded, agg_id):
+        l = Prio3.Prg.SEED_SIZE
+        k_hint = None
```
cjpatton, line 294:
```suggestion
        k_joint_rand_parts = None
```

poc/vdaf_prio3.sage
```diff
@@ -278,23 +266,41 @@ class Prio3(Vdaf):
...
+        k_blind = None
         if Prio3.Flp.JOINT_RAND_LEN > 0:
             k_blind, encoded = encoded[:l], encoded[l:]
+        if len(encoded) != 0:
+            raise ERR_DECODE
+        return (input_share, proof_share, k_blind)
+
+    @classmethod
+    def encode_public_share(Prio3,
+                            k_hint):
```
cjpatton, line 285:
```suggestion
                            k_joint_rand_parts):
```

poc/vdaf_prio3.sage
```diff
@@ -278,23 +266,41 @@ class Prio3(Vdaf):
...
+    def decode_public_share(Prio3, encoded, agg_id):
+        l = Prio3.Prg.SEED_SIZE
+        k_hint = None
+        if Prio3.Flp.JOINT_RAND_LEN > 0:
             k_hint = []
-            for _ in range(Prio3.SHARES-1):
+            for i in range(Prio3.SHARES):
                 k_joint_rand_part, encoded = encoded[:l], encoded[l:]
-                k_hint.append(k_joint_rand_part)
+                if i != agg_id:
```
cjpatton, line 299: It's slightly odd to skip the the aggregator's part while decoding. I think it would be cleaner to decode all of the parts here, then overwrite it later like we do in the paper.

cjpatton (commented):
Looking great so far. One high-level comment: Using fixed constants, e.g., `251`, for the domain separators makes the spec difficult to read. I think we should replace these something that specifies what the derived bytes are used for, e.g., instead of `251` we have something like `"query randomness"`.

Here's an important question: In your opinion, should each of these tags be equal length? If so, what we could do is keep them as single byte constants, but define the constants as global variables. For instance:

```python
DST_QUERY_RANDOMNESS = byte(251)
```

What do you think?

hannahdaviscrypto:
Short answer, yes, it is best for all the tags to be the same length because they need to be prefix-free and having them be the same length makes that apparent. Also in some cases we use longer domain separation tags of the format dst + (usage byte) + (more stuff, often the aggregator's ID) and it's easier to argue separation for those if we don't have to worry about the byte alignment of the "more stuff". The global variable idea is a good one, I'll set that up.

poc/vdaf_prio3.sage
hannahdaviscrypto: Sure

poc/vdaf_prio3.sage
```diff
@@ -88,71 +95,67 @@ class Prio3(Vdaf):
...
         # Compute joint randomness.
         joint_rand, k_joint_rand, k_joint_rand_part = [], None, None
         if Prio3.Flp.JOINT_RAND_LEN > 0:
-            encoded = Prio3.Flp.Field.encode_vec(input_share)
+            encoded = Prio3.Flp.Field.encode_vec(measurement_share)
             k_joint_rand_part = Prio3.Prg.derive_seed(
-                k_blind, dst + byte(agg_id) + nonce + encoded)
-            k_joint_rand_parts = k_hint[:agg_id] + \
+                k_blind, dst + DST_JOINT_RAND_HINT + byte(agg_id) + nonce + encoded)
+            k_joint_rand_parts = k_joint_rand_parts[:agg_id] + \
```
cjpatton, line 140: I would give this variable a name that distinguishes it from `k_joint_rand_parts` parsed from the public share. Maybe something like `k_checked_joint_rand_parts`? Accordingly, maybe `k_joint_rand` should be renamed to `k_checked_joint_rand`?

poc/vdaf_prio3.sage
```diff
@@ -1,4 +1,11 @@
 # The prio3 VDAF.
```
cjpatton, line 1: This won't compile because `byte` is not in scope yet. We need to move the constants after the imports.

poc/vdaf_prio3.sage
```diff
@@ -222,79 +225,90 @@ class Prio3(Vdaf):
...
+    @classmethod
+    def encode_public_share(Prio3,
+                            k_joint_rand_parts):
+        encoded = Bytes()
+        if Prio3.Flp.JOINT_RAND_LEN > 0:
+            encoded += concat(k_joint_rand_parts)
+        return encoded
+
+    @classmethod
+    def decode_public_share(Prio3, encoded):
```
cjpatton, line 301: The control flow in this function makes it difficult to modify later on. In particular, here you are doing:
```python
thing = X
if special_case:
    consume_buffer
    thing = Y
ensure_buffer_is_empty
return thing
```

The following control flow is preferable, as it makes failures more obvious:
```python
if not special_case:
   ensure_buffer_is_empty
   return X

consume_buffer
ensure_buffer_is_empty
return Y
```

poc/vdaf_prio3.sage
```diff
@@ -1,4 +1,11 @@
 # The prio3 VDAF.
+DST_JOINT_RAND_HINT = byte(249)
```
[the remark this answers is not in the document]
cjpatton: Works for me

poc/vdaf_prio3.sage
hannahdaviscrypto: I think k_checked_joint_rand would get confusing since we already have k_joint_rand_check. Trying "corrected" instead since we're actually overwriting.

cjpatton: ah, yes. good catch

poc/vdaf_prio3.sage
```diff
@@ -222,79 +227,96 @@ class Prio3(Vdaf):
...
         proof_share = Prio3.Flp.Field.decode_vec(encoded_proof_share)
         l = Prio3.Prg.SEED_SIZE
-        k_blind, k_hint = None, None
-        if Prio3.Flp.JOINT_RAND_LEN > 0:
-            k_blind, encoded = encoded[:l], encoded[l:]
-            k_hint = []
-            for _ in range(Prio3.SHARES-1):
-                k_joint_rand_part, encoded = encoded[:l], encoded[l:]
-                k_hint.append(k_joint_rand_part)
+        if Prio3.Flp.JOINT_RAND_LEN <= 0:
```
cjpatton, line 253: While technically correct, this code makes it look as if `-23` might be a valid value for `JOINT_RAND_LEN`. (Similarly below.)
```suggestion
        if Prio3.Flp.JOINT_RAND_LEN == 0:
```

poc/vdaf_prio3.sage
```diff
@@ -222,79 +227,96 @@ class Prio3(Vdaf):
...
                                                 Prio3.Flp.INPUT_LEN)
         k_proof_share, encoded = encoded[:l], encoded[l:]
         proof_share = Prio3.Prg.expand_into_vec(Prio3.Flp.Field,
                                                 k_proof_share,
-                                                dst + byte(agg_id),
+                                                dst + DST_PROOF_SHARE 
+                                                + byte(agg_id),
                                                 Prio3.Flp.PROOF_LEN)
-        k_blind, k_hint = None, None
+        if Prio3.Flp.JOINT_RAND_LEN <= 0:
```
cjpatton, line 289:
```suggestion
        if Prio3.Flp.JOINT_RAND_LEN == 0:
```

poc/vdaf_prio3.sage
```diff
@@ -222,79 +227,96 @@ class Prio3(Vdaf):
...
-            for _ in range(Prio3.SHARES-1):
-                k_joint_rand_part, encoded = encoded[:l], encoded[l:]
-                k_hint.append(k_joint_rand_part)
+            encoded += concat(k_joint_rand_parts)
+        return encoded
+
+    @classmethod
+    def decode_public_share(Prio3, encoded):
+        l = Prio3.Prg.SEED_SIZE
+        if Prio3.Flp.JOINT_RAND_LEN <= 0:
```
cjpatton, line 309:
```suggestion
        if Prio3.Flp.JOINT_RAND_LEN == 0:
```

poc/vdaf_prio3.sage
```diff
@@ -309,10 +331,12 @@ class Prio3(Vdaf):
         l = Prio3.Flp.Field.ENCODED_SIZE * Prio3.Flp.VERIFIER_LEN
         encoded_verifier, encoded = encoded[:l], encoded[l:]
         verifier = Prio3.Flp.Field.decode_vec(encoded_verifier)
-        k_joint_rand = None
-        if Prio3.Flp.JOINT_RAND_LEN > 0:
-            l = Prio3.Prg.SEED_SIZE
-            k_joint_rand, encoded = encoded[:l], encoded[l:]
+        if Prio3.Flp.JOINT_RAND_LEN <= 0:
```
cjpatton, line 334:
```suggestion
        if Prio3.Flp.JOINT_RAND_LEN == 0:
```

poc/vdaf_prio3.sage
```diff
@@ -326,10 +350,12 @@ class Prio3(Vdaf):
 
     @classmethod
     def decode_prep_msg(Prio3, encoded):
-        k_joint_rand_check = None
-        if Prio3.Flp.JOINT_RAND_LEN > 0:
-            l = Prio3.Prg.SEED_SIZE
-            k_joint_rand_check, encoded = encoded[:l], encoded[l:]
+        if Prio3.Flp.JOINT_RAND_LEN <= 0:
```
cjpatton, line 353:
```suggestion
        if Prio3.Flp.JOINT_RAND_LEN == 0:
```

poc/vdaf_prio3.sage
```diff
@@ -10,6 +10,13 @@ import sagelib.flp as flp
 import sagelib.flp_generic as flp_generic
 import sagelib.prg as prg
 
+DST_JOINT_RANDOMNESS = byte(1)
```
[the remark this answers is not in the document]
hannahdaviscrypto: Ok. we should make a note to change the paper before it gets published then.

poc/vdaf_prio3.sage
hannahdaviscrypto: Funnily enough, i started with == and then changed it, in the hopes that if -23 showed up as an invalid value it would cause an ERR_DECODE.

cjpatton (commented):
Looks good to me! Last step: Please squash the commits and fix up the commit message. In the body of the message, make sure to list the things that are changing.

cjpatton (approved):
Approved. Thanks for making VDAFs better, @hannahdaviscrypto !

PR #128, opened 2023-01-04
fix poc: "ipdf" instead of "idpf"

jbr:
Noticed a typo in the poplar POC


cjpatton (commented):
👍 thanks! Can you check to see if this needs to be fixed in the actual spec as well (draft-irtf-cfrg-vdaf.md)?

jbr:
I don't see any other uses of "ipdf" in the repository

cjpatton (approved):

PR #129, opened 2023-01-09
poc: Prio3: Reuse the same seed for multiple functions

cjpatton:
Each input share includes a blinding factor, a PRG seed used to derive the Aggregator's part of the joint randomness. In addition, each Helper share includes two seeds, one for its measurement share and another for its proof share.

In the (not yet published) analysis of Prio3, it was found that, as long as a distinct domain separation tag is used for each of these functions, it is safe to use the same seed for deriving the input share, proof share, and joint randomness part. This improves communication cost slightly and simplifies the algorithm.

cc/ @hannahdaviscrypto

### DIFF
--- a/poc/vdaf_prio3.sage
+++ b/poc/vdaf_prio3.sage
@@ -44,30 +44,29 @@ class Prio3(Vdaf):
 
         # Generate measurement shares.
         leader_measurement_share = inp
-        k_helper_measurement_shares = []
-        k_helper_blinds = []
+        k_seeds = []
         k_joint_rand_parts = []
-        for j in range(Prio3.SHARES-1):
-            k_blind = gen_rand(Prio3.Prg.SEED_SIZE)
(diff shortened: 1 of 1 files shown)


hannahdaviscrypto:
I think we should put a pause on this for now; this change needs a little more formal security analysis before we move forward.

In a scheme that derives the secret input share and proof share from the same seed as the public joint randomness part,  a malicious aggregator has the ability to perform a brute-force search for the input share by repeatedly guessing a seed, rederiving the joint randomness hint, and comparing its result with the public share. If the seed is sufficiently high entropy, the scheme may still be considered secure and we may want to make this change regardless, but it's going to change the security analysis a little and we should address that first.

This concern likely does not extend to Doplar because the joint randomness hints there also depend on secret VIDPF keys.

cjpatton:
Filed #185 to track this.

PR #131, opened 2023-01-10
poc: IdpfPoplar: Avoid branching on contorl bits

cjpatton:
Control bits are secret, so in order to reduce the risk of leaking them via a timing channel, implementations need to avoid branching on them. This is somewhat subtle, so spell out constant-time algorithms in a few places in the IDPF implementation.

Credit to @divergentdave for pointing out this deficiency in the spec, and for devising a couple of the algorithms.


poc/idpf_poplar.sage
```diff
@@ -137,8 +158,18 @@ class IdpfPoplar(Idpf):
 
         next_ctrl = t[bit]
         (next_seed, y) = IdpfPoplar.convert(level, s[bit])
-        if next_ctrl == Field2(1):
```
divergentdave, line 140: We should make similar bit-to-mask changes for the "if prev_ctrl..." block above.

divergentdave (changes requested):
There's one more conditional we should transform, then the extracts in the spec can be updated too.

poc/idpf_poplar.sage
cjpatton: Done, good catch.

poc/idpf_poplar.sage
```diff
@@ -127,18 +131,23 @@ class IdpfPoplar(Idpf):
...
+        s[0] = xor(s[0], prev_ctrl.mask(seed_cw))
+        s[1] = xor(s[1], prev_ctrl.mask(seed_cw))
+        t[0] += ctrl_cw[0] * prev_ctrl
+        t[1] += ctrl_cw[1] * prev_ctrl
 
         next_ctrl = t[bit]
         (next_seed, y) = IdpfPoplar.convert(level, s[bit])
-        if next_ctrl == Field2(1):
-            y = vec_add(y, w_cw)
+        # Implementation note: Here wea add the correction word to the
```
divergentdave, line 144:
```suggestion
        # Implementation note: Here we add the correction word to the
```

divergentdave (approved):

poc/field.sage
```diff
@@ -102,6 +102,19 @@ class Field2(Field):
     # Operational parameters
     gf = GF(MODULUS)
 
+    # Return `inp` unmodified if `self == 1`; otherwise return the
+    # all-zero string of the same length.
+    #
+    # Implementation note: To protect the code from timing side channels,
+    # it is important to implement this algorithm in constant time.
+    def mask(self, inp: Bytes) -> Bytes:
```
cjpatton, line 110: This is sometimes called ["conditional select"](https://docs.rs/subtle/latest/subtle/trait.ConditionallySelectable.html). I'm not sure if there is a widely used term for this.

schoppmp: I think `conditional_select` is more descriptive.

poc/idpf_poplar.sage
```diff
@@ -127,18 +131,23 @@ class IdpfPoplar(Idpf):
...
         next_ctrl = t[bit]
         (next_seed, y) = IdpfPoplar.convert(level, s[bit])
-        if next_ctrl == Field2(1):
-            y = vec_add(y, w_cw)
+        # Implementation note: Here we add the correction word to the
+        # output if `next_ctrl` is set. We avoid branching on the value of
+        # the control bit in order to reduce side channel leakage.
+        mask = Field(next_ctrl.as_unsigned())
+        for i in range(len(y)):
+            y[i] += w_cw[i] * mask
```
schoppmp, line 149: In other languages, there might be a concern that the compiler "optimizes" this to `y[i] = 0 if mask == 0 else w_cw[i]` or something similar, in particular if checking for zero is cheaper than multiplication. Not sure if that's the case with Sage.

schoppmp (approved):

poc/field.sage
cjpatton: Done.

poc/idpf_poplar.sage
cjpatton: Hmm yeah. Though I think this falls into the category of "finite field arithmetic needs to be constant time", i.e., https://github.com/divviup/libprio-rs/blob/main/src/fp.rs#L81.

PR #133, opened 2023-01-19
Check unused bits when decoding IDPF public share

divergentdave:
This updates the IdpfPoplar public share decoding algorithm to check that any unused bits alongside the packed control bits are zero.


draft-irtf-cfrg-vdaf.md
```diff
@@ -3002,6 +3002,8 @@ def decode_public_share(IdpfPoplar, encoded):
         encoded_w_cw, encoded = encoded[:l], encoded[l:]
         w_cw = Field.decode_vec(encoded_w_cw)
         correction_words.append((seed_cw, ctrl_cw, w_cw))
+    if packed_ctrl != 0:
+        raise ERR_DECODE
     if len(encoded) != 0:
```
cjpatton, line 3007: More concise.
```suggestion
    if packed_ctrl != 0 or len(encoded) != 0:
```

cjpatton (changes requested):

PR #135, opened 2023-02-02
Update spec to match reference code

cjpatton:
All planned changes to Prio3 have been implemented in the reference code; propagate these changes to the document itself. Pick up some changes for Poplar1 as well.

Changes to the reference code were required in order to make it render nicely in the dcocument. However the algorithms themselves have not changed.

This change also breaks up the auxiliary functions sections into subsections: One for joint randomness computation, another for serialization, and a new subsection for enumerating the constants.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1415,19 +1415,31 @@ This section specifies `Prio3`, an implementation of the `Vdaf` interface
 ({{vdaf}}). It has two generic parameters: an `Flp` ({{flp}}) and a `Prg`
 ({{prg}}). The associated constants and types required by the `Vdaf` interface
 are defined in {{prio3-param}}. The methods required for sharding, preparation,
-aggregation, and unsharding are described in the remaining subsections.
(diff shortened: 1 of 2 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -1785,15 +1802,29 @@ def encode_prep_msg(Prio3, k_joint_rand_check):
...
+        return None
+    l = Prio3.Prg.SEED_SIZE
+    k_joint_rand_check, encoded = encoded[:l], encoded[l:]
     if len(encoded) != 0:
         raise ERR_DECODE
     return k_joint_rand_check
 ~~~
-{: #prio3-helpers title="Helper functions required for Prio3."}
+
+#### Constants
```
schoppmp, line 1816: Moving these up, next to the variables, would help with understanding the code.

schoppmp (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: Good call. They are referenced in multiple subsections, so I put these at the top of the Prio3 section.

Issue #59, opened 2022-05-10
poc: Enforce types


cjpatton:
Type annotations are used in the reference implementation in order to generate a more readable spec. Currently these type annotations aren't for correctness. For instance, in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/57, @jbr noticed that type for the verification parameter for Prio3 is not defined.

In python3 it's possible to use a static type checker, like https://mypy.readthedocs.io/en/stable/, for this purpose. At the very least, we should extend the unit tests to check that class attributes that are expected to be non-`None` have a value.

Note that, for the verification parameter in particular, a solution to https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/38 might make this a moot point. However there are many situations for which this would be useful.

divergentdave:
One other example: GEN_ORDER has a type hint that says "Unsigned", but it is actually sage.rings.integer.Integer. This matters because there's a place it is used in single-slash division, and the result is used as an exponent. If it were Unsigned, this would be raising an element to a power of a floating point number, but in reality, the power is a sage rational number, with no fractional part, which is OK.

cjpatton:
Ugh, yeah the fact that Sage is effectively "compiled" to Python3 makes this issue quite hairy. Then intended semantics of `__div__(self, other)` for `Field` is indeed `self * other.inv()`.

Still unclear as to whether Sage is really the best choice for the reference implementation/spec. @chris-wood had prepared [a rant for IETF 113](https://datatracker.ietf.org/meeting/113/materials/slides-113-cfrg-cfrg-specifications-in-theory-and-practice-00) about this very question, but we unfortunately ran out of time to give him a full hearing.

cjpatton:
I don't see an obvious way to enforce these types:
https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/6684fcca6995e080e10bf852d68f3671bc568dca/poc/vdaf_prio3.py#L37-L54

PR #136, opened 2023-02-03
Use SHA-3 to instantiate the PRG and improve domain separation

### LINKED ISSUE #106: Securely instantiating the PRG

cjpatton:
Based on #135 (merge that first).
Closes #106.

Replace `PrgAes128` with a new scheme, `PrgSha3`, that uses SHA-3 in the cSHAKE128 mode of operation. Existing analysis models this object as a random oracle, and SHA-3 is a safer way to instantiate it.

Whenever the PRG is used to map a seed and some context to a finite domain, e.g., derive a fresh seed or a vector of field elements, the analysis models this function as an independent random oracle. One consequence of this is that we need to take care to ensure domain separation for each distinct usage of PRG. In particular, we need to format the inputs to the random oracle so that they are decodable.

To make this easier, we split the current "info string" into two components: The "context string" and the "binder string". The context string encodes the document version, algorithm, and some usage information; all of which was previously packed into the prefix of the info string. The "binder string" is an optional ephemeral artifact that we want to bind ot the output, e.g., the measurement shares in Prio3.

We set cSHAKE128's "customization string" to the context string. The seed is set to the start of the "main input" of cSHAKE128; and the remainder is the binder string. This is useful because the binder string may have arbitrary length; because the seed is always fixed length, it can always be decoded from the main input.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -29,6 +29,16 @@ author:
 
 normative:
 
+  SP800-185:
+    title: "SHA-3 Derived Functions: cSHAKE, KMAC, TupleHash and ParallelHash"
+    date: December 2016
+    seriesinfo: NIST Special Publication 800-185
+
+  FIPS202:
+    title: "SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions"
+    date: August 2015
+    seriesinfo: NIST FIPS PUB 202
+
 informative:
 
   AGJOP21:
@@ -354,8 +364,8 @@ A variable with type `Bytes` is a byte string. This document defines several
 byte-string constants. When comprised of printable ASCII characters, they are
 written as Python 3 byte-string literals (e.g., `b'some constant string'`).
 
-A global constant `VERSION` is defined, which algorithms are free to use as
-desired. Its value SHALL be `b'vdaf-03'`.
+A global constant `VERSION` of type `Unsigned` is defined, which algorithms are
+free to use as desired. Its value SHALL be `3`.
 
 This document describes algorithms for multi-party computations in which the
 parties typically communicate over a network. Wherever a quantity is defined
@@ -1135,8 +1145,7 @@ The tables below define finite fields used in the remainder of this document.
 
 A pseudorandom generator (PRG) is used to expand a short, (pseudo)random seed
 into a long string of pseudorandom bits. A PRG suitable for this document
-implements the interface specified in this section. Concrete constructions are
-described in the subsections that follow.
+implements the interface specified in this section.
 
 PRGs are defined by a class `Prg` with the following associated parameter:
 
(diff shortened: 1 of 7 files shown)


simon-friedberger:
I'm not sure what exactly the distinction between the context string and binder string is.
1. Is the intent is to use the context string for domain separation and to add Fiat-Shamir transform inputs into the binder string?
2. There are instances where `binder=byte(agg_id)` or similar which look like this should actually be domain separation. Is that intentional?
4. Can we just replace "context string" by "customization string" which is also used in the cSHAKE definition?

cjpatton:
> I'm not sure what exactly the distinction between the context string and binder string is.
> 
>     1. Is the intent is to use the context string for domain separation and to add Fiat-Shamir transform inputs into the binder string?

Yes, that's the basic idea :)

Context:
1. version ("VDAF-04")
2. Algorithm (prio3count, prio3sum, poplar1,...)
3. "usage" ("joint randomness" for prio3, "verifier randomness" for poplar1, ...)

Basically all the information that is sufficient to ensure that, if a seed gets reused, the derived outputs are independent.

Binder: Some "runtime artifact", like a nonce or an input share (as for Fiat-Shamir).

>     2. There are instances where `binder=byte(agg_id)` or similar which look like this should actually be domain separation. Is that intentional?

Hmm interesting point. I was thinking agg_id as something to "bind to", but it arguably makes more sense as context. I could go either way here. Do you have a strong preference?

>     3. Can we just replace "context string" by "customization string" which is also used in the cSHAKE definition?

Yeah I'm cool with that 👍

simon-friedberger:
I think this is fine for now. I still dislike that we are hard-coding SHA-3 but we can fix that if it ever becomes really necessary.

schoppmp (approved):

cjpatton:
Rebased and squashed.

poc/prg.sage
```diff
@@ -24,8 +24,8 @@ class Prg:
 
     # Derive a new seed.
     @classmethod
-    def derive_seed(Prg, seed: Bytes, info: Bytes) -> bytes:
-        prg = Prg(seed, info)
+    def derive_seed(Prg, seed: Bytes[Prg.SEED_SIZE], custom: Bytes, binder: Bytes):
```
divergentdave, line 27: FYI `bytes` does not support a type parameter in type hints. I see that this may be useful for documentation purposes, and it's certainly not the first place we diverge from type-checkability in here, so I'm OK leaving it.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1178,50 +1188,71 @@ def next_vec(self, Field, length: Unsigned):
...
+1. The document version (i.e.,`VERSION`);
+1. The "class" of the algorithm using the output (e.g., VDAF);
+1. A unique identifier for the algorithm; and
+1. Some indication of how the output is used (e.g., for deriving the measurement
+   shares in Prio3 {{prio3}}).
 
-    def next(self, length):
-        self.length_consumed += length
+The following algorithm is used in the remainder in order to format the
+customization string:
```
divergentdave, line 1244:
```suggestion
The following algorithm is used in the remainder of this document in order to format the
customization string:
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -2725,6 +2793,23 @@ def agg_shares_to_result(Poplar1, agg_param,
 
 ### Auxiliary Functions {#poplar1-helper-functions}
 
+~~~
+def custum(Poplar1, usage):
```
divergentdave, line 2797: nit: typo
```suggestion
def custom(Poplar1, usage):
```

divergentdave (approved):

poc/vdaf_poplar1.sage
```diff
@@ -313,36 +324,36 @@ class Poplar1(Vdaf):
         return 'bits'
 
 
-class Poplar1Aes128(Poplar1):
+class Poplar1(Poplar1):
```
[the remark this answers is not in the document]
cjpatton: Oofta, I didn't even noticed. Deleted, and moved `with_bits()` to the superclass.

draft-irtf-cfrg-vdaf.md
cjpatton: Here and elsewhere

poc/prg.sage
cjpatton: Right, the primary purpose is to make the expected input values more explicit. We are abusing type hints all over the place so far. It was pointed out before that we should probably enforce type checking for the reference code: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/59.

Issue #134, opened 2023-01-27
Poplar1 improvements


cjpatton:
After doing the VDAF security analysis and implementing Poplar1 in libprio, there are a few minor changes I think we should make to Poplar1:

- [x] Don't mix field types for PRGs; use a different PRG for each field instead. This leads to some awkwardness in the implementation: https://github.com/divviup/libprio-rs/pull/429. **Closed by https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/137**
- [ ] Consider using a distinct PRG instance for deriving correlated randomness shares at each level of the tree. That way the Aggregators don't have to "fast-forward" the PRG state for inner nodes. (We should only consider this if performance doesn't degrade on the Client side.) **WON'T FIX:** This turns out to be a performance hit for the Client. The difference on the server side is negligible.
- [x] Don't incorporate aggregation parameter into verifier randomness computation; the level and nonce should suffice. **Closed by https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/138**
- [x] Relabel domain separation bits along the lines of Prio3.
- [x] Require candidate prefixes to be unique. (Mayble also require them to be in lexicographic order, to make this check easier to implement.) **Closed by https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/139**
- [x] The encoded aggregation parameter includes the number of prefixes, which is limited to <2^16. Is this enough? **Closed by https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/143**
- [x] Encoding of aggregation parameter is optional. We should emphasize a different approach, which would be more efficient for implementations that can cache per-report state: Just serialize "prefix updates". **Closed by https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/145**

PR #137, opened 2023-02-07
Poplar1: Split PRG for leaf/inner sketch offsets

cjpatton:
Based on #136 (merge that first).
Partially addresses #134.

Use a different `Prg` instance for the inner and leaf nodes when generating shares of the offsets for the sketch computation (i.e., the `(a, b, c)` triples). This change is meant to reduce implementation complexity. In particular, for performance reasons it is desirable to buffer the seed stream output by the `Prg` prior to doing rejection sampling. The buffer size is picked based the size of the field elements; changing the field size is fiddly, as this requires us to copy the unused part of the buffer to the beginning when we wrap around the end. For details, see:
https://github.com/divviup/libprio-rs/pull/429


cjpatton:
Squashed, rebased, and moved the constants table as requested by @schoppmp here: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/135#discussion_r1100296697

divergentdave (approved):
Looks good to me. I think the hit of performing an extra PRG derivation is probably worth it for simplifying analysis and flexibility for implementations.

PR #138, opened 2023-02-07
Poplar1: Don't derive verifier randomness from prefixes

cjpatton:
Based on #137 (merge that first).
Partially addresses #134.

Security requirements will require us to restrict preparation so that the sketch computed only once per level of the IDPF tree. Remove the prefixes from the verifier randomness derivation, as this will not be needed.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2689,8 +2689,8 @@ for each. The Aggregators use these and the correlation shares provided by the
 Client to verify that the sequence of `data_share` values are additive shares of
 a one-hot vector.
 
-The algorithms below make use of auxiliary functions `verify_binder()` and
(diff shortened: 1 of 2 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -2689,8 +2689,8 @@ for each. The Aggregators use these and the correlation shares provided by the
 Client to verify that the sequence of `data_share` values are additive shares of
 a one-hot vector.
 
-The algorithms below make use of auxiliary functions `verify_binder()` and
-`decode_input_share()` defined in {{poplar1-auxiliary}}.
+The algorithms below make use of auxiliary function and `decode_input_share()`
```
divergentdave, line 2692:
```suggestion
The algorithms below make use of the auxiliary function `decode_input_share()`
```

divergentdave (approved):

PR #139, opened 2023-02-07
Poplar1: Enforce prefix uniqueness

cjpatton:
Based on #138 (merge that first).
Partially addresses #134.

If the candidate prefixes are not unique, then the refined measurement may not be a one-hot vector. This would cause the Aggregators to fail sketch verification when it would have otherwise passed.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2689,6 +2689,14 @@ for each. The Aggregators use these and the correlation shares provided by the
 Client to verify that the sequence of `data_share` values are additive shares of
 a one-hot vector.
 
+Aggregators MUST ensure the candidate prefixes are all unique and appear in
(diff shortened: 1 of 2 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -2696,6 +2704,12 @@ def prep_init(Poplar1, verify_key, agg_id, agg_param,
         Poplar1.decode_input_share(input_share)
     Field = Poplar1.Idpf.current_field(level)
 
+    # Ensure that candidate prefixes are all unique and appear in
+    # lexicographic order.
+    for i in range(1,len(prefixes)):
+        if prefixes[i-1] >= prefixes[i]:
+            raise ERR_INPUT # out-of-order prefix
+
```
cjpatton, line 2712: Reviewer note: We might consider adding this check to aggregation parameter encoding as well.

divergentdave (approved):

PR #148, opened 2023-02-15
Poplar1: Bind nonce to correlated randomness derivation

cjpatton:
Append the nonce to the binder string for deriving the offsets for the sketch computation. That way if the Aggregators use a nonce to verify a report other than the one that was picked by the Client, verification will likely fail.

This change is about defense-in-depth. We don't yet have an analysis to suggest this is necessary for either privacy or robustness. However given the importance of this binding for Prio3 (specifically for the joint randomness), it seems like a sensible change.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2597,7 +2597,7 @@ Putting everything together, the input-distribution algorithm is defined as
 follows. Function `encode_input_shares` is defined in {{poplar1-auxiliary}}.
 
 ~~~
-def measurement_to_input_shares(Poplar1, measurement, _nonce):
+def measurement_to_input_shares(Poplar1, measurement, nonce):
     prg = Poplar1.Idpf.Prg(gen_rand(Poplar1.Idpf.Prg.SEED_SIZE),
                            Poplar1.custom(DST_SHARD_RAND), b'')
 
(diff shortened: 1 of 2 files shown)


divergentdave:
I don't think the analogy to Prio3 joint randomness works. The Prio3 joint randomness derivation is part of a Fiat-Shamir transformation. On the other hand, to the best of my knowledge, this PRG instance is used for compression of client-chosen randomness, not Fiat-Shamir. Moreover, the seed is already chosen by the client, so I'm not sure whether it's worth it to additionally mix in the client-generated nonce.

cjpatton:
> I don't think the analogy to Prio3 joint randomness works. The Prio3 joint randomness derivation is part of a Fiat-Shamir transformation. On the other hand, to the best of my knowledge, this PRG instance is used for compression of client-chosen randomness, not Fiat-Shamir. Moreover, the seed is already chosen by the client, so I'm not sure whether it's worth it to additionally mix in the client-generated nonce.

Yeah the analogy to Prio3 isn't quite right. Basically what I like about this change is that the Aggregators are forced to use the nonce the Client picked in order to consume the report. Without it, the Aggregators could use a different nonce. I don't know if this feature leads to an attack, but it strikes me as slightly worrisome. In any case, I think it's cheap enough to implement without any trouble (in particular for SHA3, this shouldn't add any new rounds of Keccak).

divergentdave (approved):
I think this makes sense as a defense-in-depth measure.

PR #150, opened 2023-02-16
Revise security considerations for nonces

### LINKED ISSUE #141: Security: Clients MUST choose the nonce at random
From [this post](https://mailarchive.ietf.org/arch/msg/cfrg/-errBjFRvCqi7KuAxoZwH6iY4Vc/):

> Clients MUST choose the report nonce at random. This ensures the nonce is
> unpredictable to a privacy attacker; and it reduces the chance of
> collisions, which is required for robustness.

cjpatton:
Closes #141.

New guidance:

* Clients MUST choose the nonce at random (required for privacy as defined in {{DPRS23}}).

Updated guidance:

* Currently we say "for robustness, Aggregators MUST NOT aggregate two reports with the same nonce". In reality the situation is more subtle: The robustness game of {{DPRS23}} allows the the attacker to process any report with any initial state (i.e., aggregation parameter), so long as the sequence of initial states with which that report was processed is "allowed", as determined by the specific VDAF. This sequence of initial states is determined by the nonces provided with reports. We will clarify this as part of issue 142.

  That said, report replay is important for other security properties, like differential privacy. Thus, RECOMMEND that the nonce be used for replay protection.


draft-irtf-cfrg-vdaf.md
```diff
@@ -3240,6 +3252,21 @@ the inputs.  It is possible that a future VDAF could incorporate differential
...
+measurement, the Client MUST generate this nonce using a CSPRNG. This is
+required in order to leverage security analysis for the privacy definition of
+{{DPRS23}}, which assumes the nonce is chosen at random prior to generating the
+report.
+
+Other security considerations may require the nonce to be non-repeating. For
+example, to achieve differential privacy it is necessary to avoid "over
+exposing" a measurement by including it too many times in a single batch or
+across multiple batches. It is RECOMMENDED that the nonce generated by the
+Client be used for replay protection.
```
divergentdave, line 3268: Small clarification:
```suggestion
Other security considerations may require the nonce to be non-repeating. For
example, to achieve differential privacy it is necessary to avoid "over
exposing" a measurement by including it too many times in a single batch or
across multiple batches. It is RECOMMENDED that the nonce generated by the
Client be used by Aggregators for replay protection.
```

divergentdave (approved):

PR #151, opened 2023-02-16
IDPF: Change byte order of packed control bits

### LINKED ISSUE #147: Byte ordering of encoded bit strings
There are two places where the document currently deals with encoding and decoding of bit strings, in IdpfPoplar's public share, which packs control bits, and the optional serialization of Poplar1's aggregation parameter, which contains many bit string prefixes packed together. In both cases, the encoding is phrased in terms of using `I2OSP` on an arbitrary precision integer. In the case of IDPF pubilc share correction word control bits, the integer is formed from the control bits in LSB-to-MSB order. In the case of Poplar1, the algorithm's measurement and aggregation parameter inputs are both defined as integers, ~~though it's apparent that the least significant bit is semantically first in this case as well~~. This serialization is confusing to the eye, and requires some extra effort from implementors to accommodate (typically, reversing the order of a byte array). I think we should switch the byte-endianness of these encodings to simplify implementations. It might be worthwhile to drop the arbitrary precision integer representation altogether, and instead describe the algorithms in terms of bit strings or boolean arrays.

As a concrete example, here's how twelve bits currently get encoded:

```
sage: binascii.hexlify(sagelib.idpf_poplar.IdpfPoplar.with_bits(6).encode_public_share([(b"", [Field2(0), Field2(1)], []), (b"", [Field2(1), Field2(0)], []), (b"", [Field2(0), Field2(0)], []), (b"", [Field2(1), Field2(1)], []), (b"", [Field2(1), Field2(0)], []), (b"", [Field2(0), Field2(1)], [])]))
b'09c6'
```

|Level|Control bit index|Value|Encoded byte offset|Encoded bit position (LSB=0)|
|---|---|---|---|---|
|0|0|0|1|0|
|0|1|1|1|1|
|1|0|1|1|2|
|1|1|0|1|3|
|2|0|0|1|4|
|2|1|0|1|5|
|3|0|1|1|6|
|3|1|1|1|7|
|4|0|1|0|0|
|4|1|0|0|1|
|5|0|0|0|2|
|5|1|1|0|3|

divergentdave:
This changes the byte order of the packed control bits in the IdpfPoplar public share. The order of bits within each byte is the same, but control bits for the first level will now be encoded in the first byte of the public share.

Closes #147.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3132,15 +3132,10 @@ def convert(IdpfPoplar, level, seed):
 
 def encode_public_share(IdpfPoplar, correction_words):
     encoded = Bytes()
-    l = floor((2*IdpfPoplar.BITS + 7) / 8)
-    encoded_ctrl = [int(0)] * l
-    for (level, (_, ctrl_cw, _)) \
-        in enumerate(correction_words):
-        encoded_ctrl[level // 4] |= (
-            ctrl_cw[0].as_unsigned() |
-            (ctrl_cw[1].as_unsigned() << 1)
-        ) << (level % 4 * 2)
-    encoded += Bytes(encoded_ctrl)
+    control_bits = list(itertools.chain.from_iterable(
+        cw[1] for cw in correction_words
+    ))
+    encoded += pack_bits(control_bits)
     for (level, (seed_cw, _, w_cw)) \
         in enumerate(correction_words):
         Field = IdpfPoplar.current_field(level)
@@ -3151,14 +3146,13 @@ def encode_public_share(IdpfPoplar, correction_words):
 def decode_public_share(IdpfPoplar, encoded):
     l = floor((2*IdpfPoplar.BITS + 7) / 8)
     encoded_ctrl, encoded = encoded[:l], encoded[l:]
+    control_bits = unpack_bits(encoded_ctrl, 2 * IdpfPoplar.BITS)
     correction_words = []
     for level in range(IdpfPoplar.BITS):
         Field = IdpfPoplar.current_field(level)
         ctrl_cw = (
-            Field2((encoded_ctrl[level // 4] >>
-                (level % 4 * 2)) & 1),
-            Field2((encoded_ctrl[level // 4] >>
-                (level % 4 * 2 + 1)) & 1),
+            control_bits[level * 2],
+            control_bits[level * 2 + 1],
         )
         l = IdpfPoplar.Prg.SEED_SIZE
         seed_cw, encoded = encoded[:l], encoded[l:]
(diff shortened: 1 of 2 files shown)


cjpatton (commented):
Hmm, I wonder if this would a bit easier to read if we wrote it in terms of some function for encoding a sequence of bits (i.e., `Field2` elements). Something like:
```python
bits = list(map(lambda cw: cw[1], correction_words))
encoded += pack_bits(bits)
```

where `pack_bits(bits: Vec[Field2]) -> Bytes`. Unpacking would require the expected length, e.g., `unpack_bits(packed_bits: Bytes, length: Unsigned) -> Vec[Field2]`.

poc/idpf_poplar.sage
```diff
@@ -231,6 +233,23 @@ class IdpfPoplar(Idpf):
         return new_cls
 
 
+def pack_bits(bits: Vec[Field2]) -> Bytes:
```
cjpatton, line 236: If my suggestion above is taken, then let's move these to common.sage.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3149,22 +3146,41 @@ def encode_public_share(IdpfPoplar, correction_words):
...
         correction_words.append((seed_cw, ctrl_cw, w_cw))
-    if packed_ctrl != 0 or len(encoded) != 0:
+    leftover_bits = encoded_ctrl[-1] >> (
+        ((IdpfPoplar.BITS + 3) % 4 + 1) * 2
+    )
+    if leftover_bits != 0 or len(encoded) != 0:
         raise ERR_DECODE
     return correction_words
+
+def pack_bits(bits: Vec[Field2]) -> Bytes:
```
cjpatton, line 3170: I'd suggest moving these definitions to the preliminaries section and specifying them in English. Too much code runs the risk of making the spec less accessible. Since this method is relatively simple, a prose description of these methods would be sufficient in my opinion. (An implementer could refer to the reference code.)

poc/idpf_poplar.sage
divergentdave: `unpack_bits()` uses the `Field2` constructor, so I can't directly move it to common.sage without hacks. field.sage already imports items from common.sage, so attempting to import Field2 at top level would create a circular import. Importing the field module at the top of the function would work, but I'd rather not do that. We could leave the functions where they are, or move them to field.sage, or make them deal in integers and move them to common.sage. I'll try out the last option.

cjpatton: Good catch. I think we should keep them where they are then. I would still describe them in prose.

cjpatton (commented):
Nice! One last question: What do you think about shifting the left-over bits check into `unpack_bits()`? It seems like we would always want to enforce that the end of the last byte is 0.

cjpatton (approved):

cjpatton:
Please squash before merging!

PR #157, opened 2023-02-21
poc: Ensure the customizer for PrgAes128 is decoable

cjpatton:
fyi/ @simon-friedberger: @divergentdave thought it would be sensible to add a length prefix to the customization string.

### DIFF
--- a/poc/prg.sage
+++ b/poc/prg.sage
@@ -60,6 +60,7 @@ class PrgAes128(Prg):
 
         # Use CMAC as a pseuodorandom function to derive a key.
         hasher = CMAC.new(seed, ciphermod=AES)
+        hasher.update(I2OSP(len(custom), 2))
         hasher.update(custom)
         hasher.update(binder)
         self.key = hasher.digest()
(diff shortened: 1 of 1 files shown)


divergentdave (approved):

simon-friedberger:
I don't understand the title of the PR.
This seems like a good change but doesn't it also need to be changed here: https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-6.2.2-4
I thought the sage code was non-normative.

cjpatton:
> This seems like a good change but doesn't it also need to be changed here: https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-6.2.2-4

No, that section describes how the customization string is constructed; this PR changes how we encode that string in the input to CMAC in PrgAes128.

> I thought the sage code was non-normative.

You're right. The change here is to PrgAes128, which as of the next draft is no longer "standard".

simon-friedberger:
We should add something to the spec even if we remove it again because we want to remove the AES PRG. Maybe we end up not doing that or adding some other construction that would also benefit from this.

cjpatton:
I think I'd rather remove it if we're not using it right now. Better to have less cruft in the spec if we can!

PR #161, opened 2023-02-23
Add definition of validity scopes over aggregation parameters.

### LINKED ISSUE #142: Security: Poplar1: Do not aggregate at a report at a given level more than once
This suggestion comes from [the analysis posted here](https://mailarchive.ietf.org/arch/msg/cfrg/-errBjFRvCqi7KuAxoZwH6iY4Vc/).

schoppmp:
For Poplar1, this ensures that any given input share will only be used once at a given level.

Fixes #142.


draft-irtf-cfrg-vdaf.md
```diff
@@ -590,6 +590,13 @@ The protocol in which the DAF is used MUST ensure that the Aggregator's
 identifier is equal to the integer in range `[0, SHARES)` that matches the index
 of `input_share` in the sequence of input shares output by the Client.
 
+Concrete DAFs implementations MAY also specify a *validity scope* for
```
cjpatton, line 593: nit: The I-D tools don't render this as bold-face text as you would hope. It will instead leave these asterisks as they are. I would replace with quotations.

```suggestion
Concrete DAFs implementations MAY also specify a "validity scope" for
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -2713,6 +2720,11 @@ but this does allow uniqueness to be determined more efficiently.
 The algorithms below make use of the auxiliary function `decode_input_share()`
 defined in {{poplar1-auxiliary}}.
```
cjpatton, line 2722: Given the importance of this text, I think we should lift it to a subsection somewhere. We want to make sure it's not buried

cjpatton (commented):
I think this is the right approach. It's not overly cumbersome but clear enough. That said, we should solicit feedback from @tgeoghegan, who in all likelihood will be tasked with translating this into language in the DAP spec.

A few high level things:

1. In security considerations, I think we should talk about what attacks the requirements around the validity scope are meant to prevent.
2. Although it's somewhat redundant, I think we should redefine validity scope for VDAFs as well. Primarily because the language is not generic: Instead of `Daf.prep()`, the requirement applies to `Vdaf.prep_init() / Vdaf.prep_next()`.
* Should we say explicitly that Prio3 has no validity scope? I wonder if readers will wonder whether the absence of this definition is an oversight.

tgeoghegan (commented):
I agree with Chris that this needs to cover VDAFs as well as DAFs. Additionally, I think this needs to add the `VDAF.validity_scope` function sketched out [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/410#discussion_r1117246959) because otherwise I don't think DAP will be able to enforce this requirement, since from DAP's point of view, the aggregation parameter is opaque.

schoppmp:
Sounds good, currently working on this. Do we want `VDAF.validity_scope` to be defined for any VDAF, or only for ones that actually make use of validity scopes? For Prio3, I could always return an empty byte string, or simply not define that function.

schoppmp:
To put it another way: Do we want a default `validity_scope` implementation (implemented in the Daf/Vdaf classes in the reference code), or do we want to require that to be implemented by every DAF/VDAF?

tgeoghegan:
For some context, what I want to avoid is text in DAP (and corresponding code in DAP implementations) that has to do something like `if vdaf.is_poplar1() { check_validity_scope() } else { pass }`. So I believe we need an interface that can be implemented on all DAFs and VDAFs and generically executed by higher level protocols like DAP.

So I think the way forward is:

- [V]DAFs must define an associated type `ValidityScope`. `ValidityScope` in turn must define a method `ValidityScope.conflicts_with(other: ValidityScope) -> boolean`.
- [V]DAFs must define a method `[V]DAF.validity_scope(AggParam) -> ValidityScope`.

Then, when aggregating a report, DAP implementations would check every other aggregation parameter the report has ever been aggregated with and do:

```
VDAF.validity_scope(incoming_aggregation_parameter)
    .conflicts_with(VDAF.validity_scope(existing_aggregation_parameter))
```

For Poplar1, `ValidityScope.conflicts_with` checks the level. For Prio3, it always returns true (you may never re-prepare reports in Prio3).

There's a bunch of different ways to spell this -- you could choose different verbs or put the methods in different places -- but I think that's what we need.

schoppmp:
We decided to go with something even simpler. We'll define a function `Vdaf.is_valid(agg_param: AggParam, previous_agg_params: Vec[AggParam]) -> Bool`, that already checks `agg_param` against all `previous_agg_params`. WDYT @tgeoghegan ?

schoppmp:
I think that also matches what @simon-friedberger had planned [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/410).

tgeoghegan:
Yes, I think that works nicely!

draft-irtf-cfrg-vdaf.md

draft-irtf-cfrg-vdaf.md
```diff
@@ -1714,6 +1747,17 @@ def prep_shares_to_prep(Prio3, _agg_param, prep_shares):
...
 {: #prio3-prep-state title="Preparation state for Prio3."}
 
+### Validity of Aggregation Parameters
+
+Every input share MUST only be used once, regardless of the aggregation
+parameters used.
+
+~~~
+def is_valid(agg_param, previous_agg_params):
+    return (len(previous_agg_params) == 0)
```
cjpatton, line 1757: nit (totally feel free to ignore)
```suggestion
    return len(previous_agg_params) == 0
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3302,6 +3362,18 @@ exposing" a measurement by including it too many times in a single batch or
 across multiple batches. It is RECOMMENDED that the nonce generated by the
 Client be used by the Aggregators for replay protection.
 
+## Requirements for the Aggregation Parameters
+
+As described in {{sec-daf-validity-scopes}}, DAFs and VDAFs may impose
```
cjpatton, line 3367:
```suggestion
As described in {{sec-daf-validity-scopes}} and {{sec-vdaf-validity-scopes}} respectively, DAFs and VDAFs may impose
```

cjpatton (approved):
LGTM, just a couple nits (take them or leave them).

draft-irtf-cfrg-vdaf.md
cjpatton: Here and below.

draft-irtf-cfrg-vdaf.md
```diff
@@ -934,6 +950,23 @@ privacy).
 > example, public keys are required. On the other hand, we could consider
 > defining this class of schemes as a different primitive. See issue#77.
 
+## Validity of Aggregation Parameters {#sec-vdaf-validity-scopes}
+
+Similar to DAFs (see {{sec-vdaf-validity-scopes}}), VDAFs MAY impose
```
divergentdave, line 955: nit: link points to wrong section
```suggestion
Similar to DAFs (see {{sec-daf-validity-scopes}}), VDAFs MAY impose
```

draft-irtf-cfrg-vdaf.md

draft-irtf-cfrg-vdaf.md
```diff
@@ -934,6 +950,23 @@ privacy).
...
+restrictions for input shares and aggregation parameters. Protocols using a VDAF
+MUST ensure that for each input share and aggregation parameter `agg_param`, The
```
divergentdave, line 957: nit: copyedit
```suggestion
MUST ensure that for each input share and aggregation parameter `agg_param`, the
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -2826,6 +2870,22 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
 ~~~
 {: #poplar1-prep-state title="Preparation state for Poplar1."}
 
+### Validity of Aggregation Parameters
+
+Aggregation parameters are valid for a given input share, if no aggregation
```
divergentdave, line 2875: nit: unnecessary comma
```suggestion
Aggregation parameters are valid for a given input share if no aggregation
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -2826,6 +2870,22 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+parameter with the same level has been used with the same input share before.
+The whole preparation phase MUST NOT be run more than once for a given
+combination of input share and level.
+
+~~~
+def is_valid(agg_param, previous_agg_params):
+    (level, _) = agg_param
+    return len(filter(lambda (other_level, _): other_level == level, \
+                     previous_agg_params)) == 0
```
divergentdave, line 2884: This may be more idiomatic and clearer.
```suggestion
    return all(
        level != other_level
        for (other_level, _) in previous_agg_params
    )
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3302,6 +3362,19 @@ exposing" a measurement by including it too many times in a single batch or
...
 
+## Requirements for the Aggregation Parameters
+
+As described in {{sec-daf-validity-scopes}} and {{sec-vdaf-validity-scopes}}
+respectively, DAFs and VDAFs may impose restrictions on the re-use of input
+shares. This is to ensure that correlated randomness provided by the Client
+through the input share is not used more than once, which might compromise
+confidentiality of the Client's measurements.
+
+Protocols that make use of VDAFs therefore MUST ensure to call `Vdaf.is_valid`
```
divergentdave, line 3373: nit: rephrasing
```suggestion
Protocols that make use of VDAFs therefore MUST call `Vdaf.is_valid`
```

divergentdave (approved):

draft-irtf-cfrg-vdaf.md

PR #132, opened 2023-01-10
poc: Generate test vectors for IdpfPoplar

cjpatton:
For a given input `alpha`, generate the expected public and key shares.

PR #162, opened 2023-02-23
Add IDPF test vector

divergentdave:
This is a refresh of #132, rebased on top of #160. I changed it to conditionally override one of the IDPF keys to be `b"\x02" * length`, so that we get a more interesting public share, without runs of zeros. I also checked in the resulting test vector. This has knock-on effects on the Poplar1 test vector, naturally.


poc/idpf_poplar.sage
```diff
@@ -34,6 +35,8 @@ class IdpfPoplar(Idpf):
             gen_rand(IdpfPoplar.Prg.SEED_SIZE),
             gen_rand(IdpfPoplar.Prg.SEED_SIZE),
         ]
+        if TEST_VECTOR:
```
cjpatton, line 38: To remind us to delete these lines before copy-pasting into the spec.
```suggestion
        # REMOVE ME
        if TEST_VECTOR:
```

cjpatton (changes requested):
This seems mostly reasonable. My only concern is that the exception in the test vector generation code contradicts this statement in the spec:

> To make the tests deterministic, `gen_rand()` was replaced with a function that returns the requested number of octets.

We need to figure out a way to spell out how the random values are picked.

It may not be strictly necessary to fix this for 04, but it would certainly be nice to have by 05.

divergentdave:
Hmm, I suppose if we want to specify which random bytes go where, then making it an explicit input is the simplest way to do so. Sounds good to me.

cjpatton:
Oops, actually we will want to re-open this once I've addressed #173.

Issue #127, opened 2023-01-02
Specify nonce length


tgeoghegan:
VDAF specifies a sharding algorithm [`VDAF.measurement_to_input_shares(measurement: Measurement, nonce: Bytes) -> (Bytes, Vec[Bytes])`](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-03#name-sharding-2). In practice, `nonce` is always 16 bytes (see the test vectors and the poc in this repo, as well as the implementation in `libprio-rs`), but this isn't spelled out anywhere in VDAF. The same way they specify the verify key length, VDAFs will need to specify the expected nonce length.

PR #171, opened 2023-02-27
Specify nonce sizes for Prio3 and Poplar1

tgeoghegan:
Set `NONCE_SIZE` in the Prio3 and Poplar1 constructions, as required by the VDAF definition. The nonces are 16 bytes long, matching the existing Sage reference implementations.

Related to #127


cjpatton (approved):
Nice catch! We had changed this in the reference code but neglected to specify it in the doc itself.

PR #174, opened 2023-02-28
Poplar1: Hoist sketch check to prep_shares_to_prep

### LINKED ISSUE #169: Poplar1: Hoist decision from prep_next to prep_shares_to_prep
Currently, the final sketch validity check is done in `prep_next()` on round 2. This check only depends on the prepare message, so we could hoist it into `prep_shares_to_prep()`, as we previously did for Prio3. If we made this change, and dropped the optimization that substitutes the empty string instead of zero vectors, this would simplify implementations by making prepare message lengths predictable from knowledge of any aggregator's prepare state.

divergentdave:
This hoists the final sketch verification check from `prep_next()` to `prep_shares_to_prep()`. As a result, the round 2 prepare message will always be the empty string, if it doesn't raise an error. (Previously, it could have been the empty string or a 32-byte encoded field element)

Closes #169.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2897,14 +2897,10 @@ def prep_next(Poplar1, prep_state, opt_sketch):
                 Field.encode_vec(sketch_share))
 
     elif step == b'sketch round 2' and opt_sketch != None:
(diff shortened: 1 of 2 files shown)


cjpatton (approved):
👍

PR #175, opened 2023-02-28
Use littlen-endian byte-order for fields

### LINKED ISSUE #90: Field element encoding: Accommodations for `GF(2^255-19)`
As of #88, we'll use `GF(2^255-19)` for the leaf node field for `IdpfPoplar`. This field is a natural choice because it is close to the security level cited by [the paper](https://eprint.iacr.org/2021/017) and there are already a number of implementations of it already. (E.g., https://github.com/cloudflare/circl/tree/master/math/fp25519.) However, this choice surfaces an issue related to field element encoding that needs discussion.

First, the manner in which field elements are represented as byte strings matters for two reasons:
1. In `Prio3`, vectors of field elements are encoded and used for key derivation. In general, it's likely that the VDAF draft will have cryptographic computations that depend on encoded field elements.
2. So far we have opted to be prescriptive about the encoding of any value that would be written to the network by some application. For example, input shares and prep messages all have a fully-specified encoding. For additional discussion, see #58.

@armfazh [points out](https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/88#discussion_r899660677) that implementations of `GF(2^255-19)` typically operate directly on little-endian byte-strings (https://github.com/cloudflare/circl/blob/master/math/fp25519/fp.go#L14). The current draft uses `I2OSP()` from RFC 3447 for encoding field elements, which is big-endian. This is makes it inconvenient to use these implementations for `IdpfPoplar`, as they will be required to re-order the byte strings in order to encode them properly. This is unlikely to be a performance issue, but is potentially a source of bugs.

A related question raised by this issue is whether little-endian is somehow "better" for optimized finite field arithmetic. Perhaps using `I2OSP()` is an anti-pattern, given current trends in cryptographic code design?

Here are a few ways we might resolve this issue:

1. Keep the current big-endian encoding for all fields. In order to be compatible with the draft, an implementation of `IdpfPoplar` that uses an existing implementation of `GF(2^255-19)` may need to adjust the encoding.
2. Use little-endian for all fields. This would make it convenient to use existing implementations of `GF(2^255-19)`. Of course, this would require changing the other fields defined in the draft. However, these fields are new and there are no optimized implementations of them yet. (For what it's worth, the choice to use `I2SOP()` was somewhat arbitrary.  We noticed that it was used for other drafts, like [hash-to-curve](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/), and it seemed like a good idea to crib on prior work.)
3. Allow each field to specify its own encoding. The potential downside here is code complexity. (It's nice when there is only one way to do a thing.)
4. If we decide not to be prescriptive about encoding (cf. #58), then this is issue is somewhat less relevant. However we would still need to specify how to encode field elements that are used as inputs to other cryptographic operations. This may or may not end up mattering for `GF(2^255-19)`.

cjpatton:
Closes #90 

Add methods to_le_bytes/from_le_bytes for encoding/decoding unsigned
integers as little-endian bytes and use these for field element
serialization. This change is meant to improve alignment of the spec
with implementations, which largely prefer little endian to big endian
for representing field elements.

For consistency, rename I2OSP/OS2IP to to_be_bytes/from_be_bytes.

While at it, check in David's script for generating the PRG rejection
test case. This is not strictly needed, since we could just reverse the
order of the bytes in the test seed, but it will be useful later on.

### DIFF
--- a/poc/common.sage
+++ b/poc/common.sage
@@ -75,38 +75,30 @@ def vec_add(left, right):
 def vec_neg(vec):
     return list(map(lambda x: -x, vec))
 
-# As defined in {{!RFC3447}}, Section 4.1.
-#
-# TODO This was copy-pasted from the hash-to-curve reference implementation.
-# Instead of copy-pasting it, add hash-to-curve as a git submodule and import
-# it, similar to what the voprf draft does.
-def I2OSP(val, length):
+# Convert unsigned integer `val` in range `[0, 2^(8*length))` to a little-endian
+# byte string.
+def to_le_bytes(val, length):
     val = int(val)
     if val < 0 or val >= (1 << (8 * length)):
-        raise ValueError('bad I2OSP call: val=%d length=%d' % (val, length))
-    ret = [0] * length
-    val_ = val
-    for idx in reversed(range(0, length)):
-        ret[idx] = val_ & 0xff
-        val_ = val_ >> 8
-    ret = struct.pack('=' + 'B' * length, *ret)
-    assert OS2IP(ret, True) == val
-    return ret
+        raise ValueError('bad to_le_bytes call: val=%d length=%d' % (val, length))
+    return val.to_bytes(length, byteorder='little')
 
+# Parse an unsigned integer from a little-endian byte string.
+def from_le_bytes(encoded):
+    return int.from_bytes(encoded, byteorder='little')
 
-# As defined in {{!RFC3447}}, Section 4.2.
-#
-# TODO This was copy-pasted from the hash-to-curve reference implementation.
-# Instead of copy-pasting it, add hash-to-curve as a git submodule and import
-# it, similar to what the voprf draft does.
-def OS2IP(octets, skip_assert=False):
-    ret = 0
-    for octet in struct.unpack('=' + 'B' * len(octets), octets):
-        ret = ret << 8
-        ret += octet
-    if not skip_assert:
-        assert octets == I2OSP(ret, len(octets))
-    return ret
+# Convert unsigned integer `val` in range `[0, 2^(8*length))` to a big-endian
+# byte string.
+def to_be_bytes(val, length):
+    val = int(val)
+    if val < 0 or val >= (1 << (8 * length)):
+        raise ValueError('bad to_be_bytes call: val=%d length=%d' % (val, length))
+    return val.to_bytes(length, byteorder='big')
+
+# Parse an unsigned integer from a big-endian byte string.
+def from_be_bytes(encoded):
+    return int.from_bytes(encoded, byteorder='big')
+
 
 # Return the concatenated byte strings.
 def concat(parts: Vec[Bytes]) -> Bytes:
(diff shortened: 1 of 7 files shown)


poc/common.sage
```diff
@@ -75,38 +75,17 @@ def vec_add(left, right):
 def vec_neg(vec):
     return list(map(lambda x: -x, vec))
```
cjpatton, line 77: Reviewer note: I plan to propagate these changes to the draft before merging this PR. For now, the PR is marked as draft until we've decided whether to reverse the IDPF index bit order (see below).

poc/vdaf_poplar1.sage
```diff
@@ -389,24 +389,24 @@ if __name__ == '__main__':
...
             127,
-            [OS2IP(b'0123456789abcdef')],
+            [int.from_bytes(b'0123456789abcdef', byteorder='big')],
         ),
         [
-            OS2IP(b'0123456789abcdef'),
+            int.from_bytes(b'0123456789abcdef', byteorder='big'),
         ],
         [1],
     )
```
cjpatton, line 398: Reviewer note: For this test to pass it is necessary to interpret the input measurements and candidate prefixes as big-endian integers. This is because of the order in which we traverse the bits of the IDPF index in the IDPF key-generation algorithm.

If we reversed the order, then we change `int.from_bytes(b'00000000', byteorder='big')` to `from_le_bytes(b'00000000')` and so on.

poc/vdaf_poplar1.sage
```diff
@@ -310,24 +310,24 @@ class Poplar1(Vdaf):
...
-        encoded += I2OSP(level, 2)
-        encoded += I2OSP(len(prefixes), 4)
+        encoded += to_le_bytes(level, 2)
+        encoded += to_le_bytes(len(prefixes), 4)
         packed = 0
         for (i, prefix) in enumerate(prefixes):
             packed |= prefix << ((level+1) * i)
         l = floor(((level+1) * len(prefixes) + 7) / 8)
-        encoded += I2OSP(packed, l)
+        encoded += to_le_bytes(packed, l)
```
divergentdave, line 319: I think we should avoid switching the endianness of this, because it would result in encoding candidate prefixes with a confusing combination of bit and byte orders. For example, taking a test input from below:

```
# With PR
sage: sagelib.vdaf_poplar1.Poplar1().with_bits(128).encode_agg_param(127, [int.from_bytes(b'0123456789abcdef', byteorder='big')])
b'\x7f\x00\x01\x00\x00\x00fedcba9876543210'

# With `main`
sage: sagelib.vdaf_poplar1.Poplar1().with_bits(128).encode_agg_param(127, [int.from_bytes(b'0123456789abcdef', byteorder='big')])
b'\x00\x7f\x00\x00\x00\x010123456789abcdef'
```

cjpatton: ACK, I'll add back `to_be_bytes()` and revert this. Thanks for the careful feedback here.

poc/vdaf_poplar1.sage
cjpatton: SGTM, I'll refocus and ping you when I'm ready for a full review.

cjpatton:
Ready for review. @divergentdave note that I have changed the scope of the wire changes to just using little endian for finite fields. Given that there are places in IDPF land where we want big endian, it makes sense to keep the change minimal in my opinion.

divergentdave (approved):

PR #180, opened 2023-03-01
Define the IDPF "prefix tree" and add privacy considerations for it

### LINKED ISSUE #102: Review: Henry Corrigan-Gibbs

cjpatton:
Closes #102.

In general the ability to aggregate a batch multiple times with different aggregation parameters may lead to unintended leakage. We don't really have a great solution for this right now, but point to what we know about this.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2558,7 +2558,8 @@ where `l <= BITS`. We will refer to the latter as the set of "candidate
 prefixes". The Aggregators' goal is to count how many inputs are prefixed by
 each candidate prefix.
 
-This functionality is the core component of the Poplar protocol {{BBCGGI21}}. At
+This functionality is the core component of the Poplar protocol {{BBCGGI21}},
+which was designed to compute the heavy hitters over a set of input strings. At
 a high level, the protocol works as follows.
 
 1. Each Client splits its input string into input shares and sends one share to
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -3432,6 +3443,28 @@ on the set of all aggregation parameters used for a Client's input share, and
...
+by the application.
+
+For example, when Poplar1 is used for heavy hitters, the Aggregators learn not
+only the heavy hitters themselves, but also the prefix tree (as defined in
+{{poplar1}}) computed along the way. Indeed, this leakage is inherent to any
+construction that uses an IDPF ({{idpf}}) in the same way. Depending on the
+distribution of the measurments, the prefix tree can leak a significant amount
+of information about unpopular inputs. For instance, it is possible (though
+perhaps unlikely) for a large set of non-heavy-hitter values to share a common
+prefix, which would be leaked by a prefix tree with a sufficinetly small
```
divergentdave, line 3459:
```suggestion
prefix, which would be leaked by a prefix tree with a sufficiently small
```

divergentdave (approved):

cjpatton:
Squashed and rebased.

Issue #173, opened 2023-02-28
Specify randomness used for test vectors


cjpatton:
Here is my pitch for fixing up the test vectors.

Update the `Vdaf` API so that `measurement_to_input_shares` takes in `rand: Bytes[Vdaf.RAND_SIZE]`. Normally we would pick this value by calling `gen_rand(Vdaf.RAND_SIZE]`; but for test vectors, we fix the value.

This would allow us to exercise what we need for the IDPF correction words. It's not the *minimal* change required for this I think, but it's the easiest to understand.

How does this sound?

If you agree, let's close this PR and I'll open up a new one. (You can also handle it if you'd like to, but no pressure. I think your time is better spent updating libprio.)

_Originally posted by @cjpatton in https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/162#issuecomment-1448413761_

cjpatton:
Once the current PR stack is merged, all that remains is to propagate the changes to the draft.

PR #187, opened 2023-03-02
poc: Use an increasing sequence for test vector `gen_rand()`

cjpatton:
Based on #186 (merge that first).
Partially addresses #173.

Instead of outputting `01010101...`, `gen_rand()` now outputs `00010203...`. This ensures that the IDFP key shares generated for the test vectors are distinct, thereby improving coverage of the vectors.


poc/common.sage
```diff
@@ -48,7 +48,10 @@ def zeros(length):
 # Return the requested number of random bytes.
 def gen_rand(length):
     if TEST_VECTOR:
-        return bytes([0x01] * length)
+        out = []
+        for i in range(length):
+            out.append(i % length)
+        return bytes(out)
```
divergentdave, line 54: `i % length` seems strange, as it will never wrap around
```suggestion
        return bytes(int(i & 0xff) for i in range(length))
```

cjpatton: Uhhh yup, this was supposed to be `256`. Good catch.

divergentdave (approved):

PR #192, opened 2023-03-08
Fix encoding of output shares in test vectors

divergentdave:
There's a stray big-endian encoding of field elements in the test vector formatting code, this PR fixes it and brings it in line with how the other fields are encoded.


cjpatton (approved):
👍

Issue #159, opened 2023-02-23
Distinguish between PRGs and Random Oracles


schoppmp:
We currently have a single Prg interface, but we use it in places where either a PRG (in the cryptographic sense) or a Random Oracle is required. We should make clear which of these we need where.

cjpatton:
One potential solution is to dump the `Prg` API altogether and just hard-code SHA-3 in Prio3 and Poplar1 verification (we will have something else for IDPF; see #32). I think @bifurcation suggested this route might make sense? If SHA-3 is "good enough", then perhaps we don't need agility?

schoppmp:
Eventually hard-coding would make sense, yes. Should we make an effort though to limit it to where we actually need a random oracle? For example, the Prio version from ENPA uses an [AES-CTR based PRG](https://github.com/google/libprio-cc/tree/master/prio/prng), and I'm not sure if a random oracle / an XOF is actually needed there. Same for the places where we essentially extend a short random seed to a longer pseudorandom string.

cjpatton:
Ideally we would -- admittedly our analysis in {{DPRS23}} is a bit heavy-handed with ROs :)

I think the right way to prioritize changes here is based on performance evaluation. From what I've seen so far (see https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/106) it seems like the cost of SHA-3 over AES for Prio3 is negligible. This makes sense because the cost is dominated by FF arithmetic, not the PRG. My hunch is that the same is true for ENPA-Prio (aka [Prio2](https://github.com/divviup/libprio-rs/blob/main/src/vdaf/prio2.rs)), but I haven't actually looked closely at this.

PR #193, opened 2023-03-09
Add fixed-key AES specification

### LINKED ISSUE #32: PRG: Fixed-key AES

schoppmp:
Closes #32.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -138,6 +138,18 @@ informative:
     seriesinfo: EUROCRYPT 2014
     target: https://link.springer.com/chapter/10.1007/978-3-642-55220-5_35
 
+  GKWWY20:
+    title: Better concrete security for half-gates garbling (in the multi-instance setting)
+    authors:
+      - ins: C. Guo
+      - ins: J. Katz
+      - ins: X. Wang
+      - ins: C. Weng
+      - ins: Y. Yu
+    date: 2020
+    seriesinfo: CRYPTO 2020
+    target: https://link.springer.com/chapter/10.1007/978-3-030-56880-1_28
+
   OriginTelemetry:
     title: "Origin Telemetry"
     date: 2020
@@ -1337,6 +1349,7 @@ def expand_into_vec(Prg,
 
 This section describes PrgSha3, a PRG based on the Keccak permutation of SHA-3
 {{FIPS202}}. Keccak is used in the cSHAKE128 mode of operation {{SP800-185}}.
+This Prg is RECOMMENDED for all use cases within VDAFs.
 
 ~~~
 class PrgSha3(Prg):
@@ -1362,6 +1375,57 @@ class PrgSha3(Prg):
         return stream[-length:]
 ~~~
 {: title="Definition of PRG PrgSha3."}
+
+### PrgFixedKeyAes128 {#prg-fixed-key-aes128}
+
+While PrgSha3 as described above can be securely used in all cases where a Prg
+is needed in the VDAFs described in this document, there are some cases where
+a more efficient instantiation based on fixed-key AES is possible. For now, this
+is limited to the Prg used inside the Idpf {{idpf}} implementation in Poplar1
+{{idpf-poplar}}. It is NOT RECOMMENDED to use this Prg anywhere else.
+See Security Considerations {{security}} for a more detailed discussion.
+
+~~~
+class PrgFixedKeyAes128(Prg):
+    # Associated parameters
+    SEED_SIZE = 16
+
+    def __init__(self, seed, custom, binder):
+        self.length_consumed = 0
+
+        # Use SHA-3 to derive a key from the binder and customization
+        # strings. Note that the AES key does not need to be kept
+        # secret from any party. However, when used with IpdfPoplar,
+        # we require the binder to be a random nonce.
+        #
+        # Implementation note: This step can be cached across PRG
+        # evaluations with many different seeds.
+        self.fixed_key = cSHAKE128(binder, 16, b'', custom)
+        self.seed = seed
+
+    def next(self, length: Unsigned) -> Bytes:
+        offset = self.length_consumed % 16
+        new_length = self.length_consumed + length
+        block_range = range(
+            int(self.length_consumed / 16),
+            int(new_length / 16) + 1)
+        self.length_consumed = new_length
+
+        hashed_blocks = [
+            self.hash_block(xor(self.seed, to_le_bytes(i, 16))) \
+                         for i in block_range
+        ]
+        return concat(hashed_blocks)[offset:offset+length]
+
+    # The multi-instance tweakable circular correlation-robust hash function of
+    # [GKWWY20] (Section 4.2).
+    #
+    # Function `AES128(key, block)` is the AES-128 blockcipher.
+    def hash_block(self, block):
+        lo, hi = block[:8], block[8:]
+        sigma = hi + xor(hi, lo)
+        return xor(AES128(self.fixed_key, sigma), sigma)
+~~~
 
 ### The Customization and Binder Strings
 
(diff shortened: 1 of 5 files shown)


cjpatton (commented):
I’m still working through the PR, but looking good so far. I should have a full pass done by the time we meet. In the meantime, three high-level things:

1. In the markdown version of PrgSha3 I replaced the PyCryptodome APIs with the functions as defined in the SHA-3 spec. Could we do the same for PrgFixedKeyAes128?
2. I think we should "fix" PrgFixedKeyAes128 as the only Prg used for IdpfPoplar. The main reason is that we want to generate test vectors for this, since we'll have to implement it elsewhere (in libprio-rs in particular).
3.  You mentioned you consulted with one of the authors of {{GKWWY?20}} -- it would be appropraite to acknowledge them in the Acknwledgements section.

poc/prg.sage
```diff
@@ -99,6 +99,45 @@ class PrgSha3(Prg):
...
+
+    def next(self, length: Unsigned) -> Bytes:
+        offset = self.length_consumed % 16
+        new_length = self.length_consumed + length
+        blocks = range(int(self.length_consumed / 16), int(new_length / 16) + 1)
+        self.length_consumed = new_length
+
+        # hash_block is a circular correlation-robust hash function.
+        # See https://eprint.iacr.org/2019/074.pdf pp. 18-19.
+        def hash_block(input):
```
cjpatton, line 131: nit: Rather than define this inline, I think it would make more sense as a helper function defined in the main scope.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3480,6 +3544,21 @@ on the set of all aggregation parameters used for a Client's input share, and
...
+Pseudorandom Generators in the security analyses of our protocols. Instead, most
+of them are modeled as Random Oracles. For these use cases, we want to be
+conservative in our assumptions, and hence prescribe PrgSha3 as the only
+RECOMMENDED Prg instantiation.
+
+The one exception is the PRG used in the Idpf implementation IdpfPoplar
+{{idpf-poplar}}. Here, a Random Oracle is not needed to prove security, and
+hence a construction based on fixed-key AES {{prg-fixed-key-aes-128}} can be
+used. However, as PrgFixedKeyAes128 has been shown to be differentiable from
+a Random Oracle {{GKWWY20}}, it is NOT RECOMMENDED to use it anywhere else.
```
cjpatton, line 3560: The end of this section should point to the OPEN ISSUE of #159.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3480,6 +3544,21 @@ on the set of all aggregation parameters used for a Client's input share, and
 only proceed with the preparation and aggregation phases if that function call
 returns `True`.
 
+## Pseudorandom Generators and Random Oracles
+
+The objects we describe in {{prg}} share a common interface, which we have
+called Prg. However, these are not necessarily all modeled as cryptographic
+Pseudorandom Generators in the security analyses of our protocols. Instead, most
+of them are modeled as Random Oracles. For these use cases, we want to be
```
cjpatton, line 3552: * No need to capitalize this IMO (here and below)
* We should add a reference for a definition of "random oracle". BR93 ("Random Oracles are Practical") would be sufficient, or maybe there's a section in {{GKWW20}} we could point to?
```suggestion
of them are modeled as random oracles {{needs-ref}}. For these use cases, we want to be
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3389,8 +3453,8 @@ throws an error.
 ## Instantiation {#poplar1-inst}
 
 By default, Poplar1 is instantiated with IdpfPoplar (`VALUE_LEN == 2`) and
-PrgSha3 ({{prg-sha3}}). This VDAF is suitable for any positive value of `BITS`.
-Test vectors can be found in {{test-vectors}}.
+PrgFixedKeyAes128 ({{prg-fixed-key-aes-128}}). This VDAF is suitable for any
+positive value of `BITS`. Test vectors can be found in {{test-vectors}}.
```
divergentdave, line 3457: This should be clarified to say that Poplar1 is instantiated with PrgSha3 for the VDAF itself, and the IdpfPoplar inside is instantiated with PrgFixedKeyAes128.

cjpatton: Replaced the reference here with PrgSha3.

cjpatton (approved):
I've made addressed some of the comments and have pushed a few changes. I think this is in decent shape, but I would appreciate careful feedback before reviewing. And of course if there are any changes I made that you don't like, feel free to override me.

Non-blocking comments in-line.

cjpatton:
Let's make sure to ACK whoever you worked with on this.

poc/idpf.sage
```diff
@@ -53,13 +53,15 @@ class Idpf:
...
+            rand: Bytes[Idpf.RAND_SIZE],
+            binder: Bytes) -> (Bytes, Vec[Bytes]):
         raise Error('not implemented')
 
     # Evaluate an IDPF key at a given level of the tree and with the given set
     # of prefixes. The output is a vector where each element is a vector of
     # length `VALUE_LEN`. The output field is `FieldLeaf` if `level == BITS` and
-    # `FieldInner` otherwise.
+    # `FieldInner` otherwise. `binder` must match the string passed by the
+    # client to `gen`.
```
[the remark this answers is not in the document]
schoppmp: The binder string. Updated it.

schoppmp:
> Let's make sure to ACK whoever you worked with on this.

Done.

schoppmp:
The current version looks great, thanks for the help!

draft-irtf-cfrg-vdaf.md

cjpatton:
Squashed.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1363,6 +1376,57 @@ class PrgSha3(Prg):
...
+        ]
+        return concat(hashed_blocks)[offset:offset+length]
+
+    # The multi-instance tweakable circular correlation-robust hash function of
+    # [GKWWY20] (Section 4.2).
+    #
+    # Function `AES128(key, block)` is the AES-128 blockcipher.
+    def hash_block(self, block):
+        lo, hi = block[:8], block[8:]
+        sigma = hi + xor(hi, lo)
```
martinthomson, line 1426: Is the '+' operator here a concatenation?  It might be better to make that explicit because I came away with the distinct impression that you were losing the upper bits of the value here.

martinthomson: sigma is a function, not a variable, so perhaps you could frame it that way

draft-irtf-cfrg-vdaf.md
```diff
@@ -1363,6 +1376,57 @@ class PrgSha3(Prg):
...
+        self.length_consumed = new_length
+
+        hashed_blocks = [
+            self.hash_block(xor(self.seed, to_le_bytes(i, 16))) \
+                         for i in block_range
+        ]
+        return concat(hashed_blocks)[offset:offset+length]
+
+    # The multi-instance tweakable circular correlation-robust hash function of
+    # [GKWWY20] (Section 4.2).
```
martinthomson, line 1421: I don't see any tweaking going on here.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1363,6 +1376,57 @@ class PrgSha3(Prg):
...
+        self.length_consumed = 0
+
+        # Use SHA-3 to derive a key from the binder and customization
+        # strings. Note that the AES key does not need to be kept
+        # secret from any party. However, when used with IpdfPoplar,
+        # we require the binder to be a random nonce.
+        #
+        # Implementation note: This step can be cached across PRG
+        # evaluations with many different seeds.
+        self.fixed_key = cSHAKE128(binder, 16, b'', custom)
```
martinthomson, line 1403: Does this really need to use a hash function that barely anyone has?  I know that ... shiny... but this is a bit of a pain.

draft-irtf-cfrg-vdaf.md
cjpatton: * Yes, "+" means concatenation here. We can replace with "concat()" for clarity?
* Actually, sigma is meant to be a variable (it is input to AES, then xored with the output).

draft-irtf-cfrg-vdaf.md
cjpatton: cc/ @schoppmp

draft-irtf-cfrg-vdaf.md
cjpatton: Do you mean SHA-3 or cSHAKE in particular?

We're using SHA-3 because we needed an XOF. We looked at using HKDF-SHA-256 as an alternative, but the performance was pretty bad. Plus, we have some cases where we need longer outputs than HKDF permits.

We're using cSHAKE for PrgSha3 as well. What's nice about it is that we get distinct notions of "customization" and "binder" strings. (See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-05#section-6.2.3.)

cjpatton: FWIW, @simon-friedberger may be planning to implement this in NSS :)

cjpatton: We could use SHA-2 here if for example you waned PrgFixedKeyAes128 but not PrgSha3.

draft-irtf-cfrg-vdaf.md
martinthomson: https://eprint.iacr.org/2019/074.pdf Section 7.3 (last paragraph, "Instantiating $sigma$") has it as a function.  That function needs to be a linear orthomorphism, as this is.  The value that is input to AES is the output of that function (which I guess you can call sigma too if you like).

draft-irtf-cfrg-vdaf.md
martinthomson: Simon already reached out about the nasty complexity that this produces for us.  I think that the current approach involves a bunch of non-NSS crypto, which is sort-of fine, but suboptimal.

I realize that there are advantages to using the new stuff, but that doesn't make it less awkward.

martinthomson: Oh, I should say, why are you concerned about the performance of something that you instantiate so few times during a protocol run?  We're using HKDF with SHA-256 in our prototypes of IPA and it's in the noise, even for tiny runs.

draft-irtf-cfrg-vdaf.md
schoppmp: Agreed, explicitly calling `concat` would be better here. We can call the variable `sigma_block` or so to emphasize that it is the function sigma evaluated on `block`?

draft-irtf-cfrg-vdaf.md
schoppmp: If you look at Section 4.2 in https://eprint.iacr.org/2019/1168.pdf, the tweak i is passed as the key to the block cipher. So for a single client, the tweak stays the same, but it differs across clients.

draft-irtf-cfrg-vdaf.md
schoppmp: The reason I chose cSHAKE here was to be consistent with PrgSha3 and not require yet another hash function dependency. If this is creating problems for implementations, I don't have a problem with switching to HKDF-SHA-256, or even simply SHA-256/128. We just want the AES key to be random for each client and don't require anything else from the hash function.

simon-friedberger: It's reasonable to rely on the same primitive here. If we ever allow more options for the PrgSha3 this should be updated as well but it's fine for now.

cjpatton: @martinthomson the performance issue with HKDF comes up when we need an XOF, like we need in Prio3. (See https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/106#issuecomment-1397492095.) I agree the performance cost would be negligible, and I also think plain SHA-2 would be fine.

PR #194, opened 2023-03-09
Prio3: Change order of seeds in random coins

divergentdave:
This PR rearranges the order in which Prio3 is defined to take seeds for various purposes from its random input string. This will change test vector output. This change is intended to simplify implementation in libprio-rs, see divviup/libprio-rs#499 for prior discussion. The key changes are that a) helper measurement shares and proof shares are interleaved, b) helper joint randomness blinds are also interleaved after other per-helper seeds, if present, (both to keep related seeds near each other) c) the leader joint randomness blind comes after all helper-related seeds (since it will naturally be used last, as its measurement share must be computed last), d) the proof randomness seed is taken from the last seed in the random input. (since it's used last)

As a nice side benefit, the variable `k_blinds` in the proof-of-concept implementation is split up, so array indexes now line up.


cjpatton (approved):
Cool!

Issue #204, opened 2023-03-17
Convert proof of concept to Sage-flavored-Python


divergentdave:
We'd like to rework the proof of concept to be written in Python syntax, and import parts of Sage as library modules. The syntactic sugar of Sage syntax is not worth it when compared to the hurdles it imposes on code reuse through importing. Numeric literals being Sage objects instead of Python primitive types has caused a lot of small issues, because so much of the code is traditional imperative programming rather than algebraic. We can replace .sage files with .py files, and then run them with `sage -python filename.py` so we still get access to Sage's various libraries, which aren't otherwise available in plain Python interpreters. Once we have finished this conversion, we can start using traditional Python code quality tools, as well.

cjpatton:
Before we decide to do this, it would be useful to get some insight from folks in the CFRG who have developed reference code in Sage. I'm looking at you, @armfazh and @chris-wood!

chris-wood:
@divergentdave this is a _great_ idea. I think it has the same benefits as using Sage natively, but brings the added benefits of using existing tooling to improve the code.

divergentdave:
We should set up a `format` Makefile target to run Python files through `autopep8` and `isort`.

armfazh:
(A different task, but related) it's worth to investigate the ability to run Sage in CI.
In github actions, you can run a docker image containing Sage.

chris-wood:
@dconnolly did this for [FROST](https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/.github/workflows/poc.yml) =)

dconnolly:
> @dconnolly did this for [FROST](https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/.github/workflows/poc.yml) =)

The Sage in homebrew is generally up to date with latest and was a quick and simple set up

PR #208, opened 2023-03-21
Run Sage tests in CI

divergentdave:
This installs Sage and runs the test suite. See discussion on #204.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   test:
-    name: "Run Sage tests"
+    name: "Run unit tests for reference code"
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


divergentdave:
This is repeatably failing with a 404 when fetching one package, but I can't reproduce the 404 locally. I may try again tomorrow in case this is a package caching issue.

(I didn't go with Docker initially because `sagemath/sagemath:9.6` doesn't include GNU make in its image)

cjpatton:
Take a look at what @dconnolly set up for FROST: https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/.github/workflows/poc.yml

cjpatton (changes requested):

cjpatton (approved):

PR #210, opened 2023-03-22
poc: Change superclass of Error to Exception

divergentdave:
This changes the superclass of `Error` to be `Exception` instead of `BaseException`. The [exception documentation](https://docs.python.org/3/library/exceptions.html) says that new exception types should typically do so. Inheriting directly from `BaseException` is reserved for special things like `GeneratorExit`, `KeyboardInterrupt`, `SystemExit`, etc.

### DIFF
--- a/poc/common.py
+++ b/poc/common.py
@@ -20,7 +20,7 @@ Vec = List
 
 
 # Base class for errors.
-class Error(BaseException):
+class Error(Exception):
     def __init__(self, msg):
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
At some point we should consider revisiting error handling: as it is, it's not very helpful with actually debugging. However this would impact the spec, so perhaps best to raise an issue  if we want to work on this.

PR #213, opened 2023-03-22
common.py: Move comments to docstrings

divergentdave:
This moves comments describing methods into docstrings, so they'll appear in IDEs, etc.

### DIFF
--- a/poc/common.py
+++ b/poc/common.py
@@ -1,4 +1,4 @@
-# Functionalities used by other modules.
+"""Functionalities used by other modules."""
 
 from functools import reduce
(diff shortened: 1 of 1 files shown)


divergentdave:
FWIW, some external references regarding quotes: [PEP 257](https://peps.python.org/pep-0257/#what-is-a-docstring) recommends triple double quotes for docstrings. (though any string literal would work, syntactically) The opinionated [Black formatter](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings) prefers double-quoted string literals everywhere unless a single-quoted literal would need less escapes.

PR #214, opened 2023-03-30
Move comments to docstrings

divergentdave:
This moves comments before classes/functions/methods to docstrings in all Sage files. I also did the same in the document, where the existing extracts already included such comments.


draft-irtf-cfrg-vdaf.md
```diff
@@ -1247,16 +1247,14 @@ remainder of this document. Note that an exception is raised by each function if
 the operands are not the same length.
 
 ~~~
-# Compute the inner product of the operands.
-def inner_product(left: Vec[Field], right: Vec[Field]) -> Field:
-    return sum(map(lambda x: x[0] * x[1], zip(left, right)))
```
divergentdave, line 1252: This function is not used anywhere, and no longer appears in the proof of concept implementation, so I deleted it rather than update it.

poc/prg.sage
```diff
@@ -138,10 +153,15 @@ class PrgFixedKeyAes128(Prg):
...
-    # [GKWWY20] (Section 4.2). The tweak here is the key that stays constant for
-    # all PRG evaluations of the same client, but differs between clients.
     def hash_block(self, block):
+        """
+        The multi-instance tweakable circular correlation-robust hash
+        function of [GKWWY20] (Section 4.2). The tweak here is the key
+        that stays constant for all PRG evaluations of the same client,
+        but differs between clients.
+
+        Function `AES128(key, block)` is the AES-128 blockcipher.
```
divergentdave, line 163: This second paragraph is copied from the corresponding comment in the Markdown file.

divergentdave (commented):
Reviewer notes:

draft-irtf-cfrg-vdaf.md
```diff
@@ -3355,20 +3368,29 @@ def eval(IdpfPoplar, agg_id, public_share, init_seed,
...
+    """
+    Compute the next node in the IDPF tree along the path determined by
+    a candidate prefix. The next node is determined by `bit`, the bit of
+    the prefix corresponding to the next level of the tree.
+
+    TODO Consider implementing some version of the optimization
+    discussed at the end of [BBCGGI21, Appendix C.2]. This could on
+    average reduce the number of AES calls by a constant factor.
+    """
+
```
cjpatton, line 3393: Consistency: Do we prefer a blankline after docucomment or not?

cjpatton (approved):

draft-irtf-cfrg-vdaf.md
divergentdave: I'm not aware of a norm regarding this, I looked around in the `requests` codebase as an example and saw a combination. I tried to include a blank line when it helped with clarity, such as when a line comment immediately followed, or the function was big enough that it already had some blank lines. I think this gets treated as a special case of blank lines within a function/method body (as the docstring is technically a statement containing a string literal expression) and left up to the author's discretion.

PR #221, opened 2023-04-14
Convert daf and vdaf modules to python

divergentdave:
This is part of #204. These two modules are mostly abstract base classes, so this conversion was straightforward.

### DIFF
--- a/poc/vdaf.py
+++ b/poc/vdaf.py
@@ -4,9 +4,9 @@ from __future__ import annotations
 from functools import reduce
 import json
 import os
-from common import ERR_VERIFY, VERSION, Bytes, Error, Unsigned, Vec, \
-                   format_custom, gen_rand, to_le_bytes, \
-                   print_wrapped_line
+from common import ERR_VERIFY, VERSION, Bool, Bytes, Error, \
+                   Unsigned, Vec, format_custom, gen_rand, \
+                   to_le_bytes, print_wrapped_line
 import sagelib.field as field
 from sagelib.prg import PrgSha3
 from typing import Optional, Tuple, Union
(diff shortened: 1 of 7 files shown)


poc/vdaf.py
```diff
@@ -51,9 +51,9 @@ class Vdaf:
     @classmethod
     def measurement_to_input_shares(Vdaf,
                                     measurement: Measurement,
-                                    nonce: Bytes[Vdaf.NONCE_SIZE],
-                                    rand: Bytes[Vdaf.RAND_SIZE],
-                                    ) -> (Bytes, Vec[Bytes]):
+                                    nonce: Bytes["Vdaf.NONCE_SIZE"],
+                                    rand: Bytes["Vdaf.RAND_SIZE"],
+                                    ) -> tuple[Bytes, Vec[Bytes]]:
```
cjpatton, line 56: nit nit nit: In `common.py` you'll see we're overriding native types with capitalized versions, e.g., `Bool` instead of `bool`. If we wanted to be consistent we would do the same for `tuple`.

Here's my question to you: Is this a terrible idea? Visually it's nice to be able to distinguish types from variables using camel and snake case respectively. However, arguably we should just be following Python's lead here and using lower case for primitive types.

WDYT:
1. `tuple` -> `Tuple` for consistency; or
2. leave as is and file an issue to use native type names (i.e., `bool` instead of `Bool`)?

I'm leaning towards 2.

cjpatton (approved):

poc/vdaf.py
divergentdave: Same, I'd prefer to drop the aliases.

PR #232, opened 2023-04-28
Split up inputs and proofs in run_flp()

### LINKED ISSUE #227: `run_flp()` takes `num_shares` but doesn't actually split the inputs
```
An FLP is executed by the prover and verifier as follows:

~~~
def run_flp(Flp, inp: Vec[Flp.Field], num_shares: Unsigned):
    joint_rand = Flp.Field.rand_vec(Flp.JOINT_RAND_LEN)
    prove_rand = Flp.Field.rand_vec(Flp.PROVE_RAND_LEN)
    query_rand = Flp.Field.rand_vec(Flp.QUERY_RAND_LEN)

    # Prover generates the proof.
    proof = Flp.prove(inp, prove_rand, joint_rand)

    # Verifier queries the input and proof.
    verifier = Flp.query(inp, proof, query_rand, joint_rand, num_shares)

    # Verifier decides if the input is valid.
    return Flp.decide(verifier)
~~~
```

This text is meant to clarify the text that comes before, which explains "full inearity" and the technical issues that are relevant to implementers:

```
Our application requires that the FLP is "fully linear" in the sense defined in
{{BBCGGI19}}. As a practical matter, what this property implies is that, when
run on a share of the input and proof, the query-generation algorithm outputs a
share of the verifier message. Furthermore, the privacy property of the FLP
system ensures that the verifier message reveals nothing about the input's
validity. Therefore, to decide if an input is valid, the Aggregators will run
the query-generation algorithm locally, exchange verifier shares, combine them
to recover the verifier message, and run the decision algorithm.

The query-generation algorithm includes a parameter `num_shares` that specifies
the number of shares of the input and proof that were generated. If these data
are not secret shared, then `num_shares == 1`. This parameter is useful for
certain FLP constructions. For example, the FLP in {{flp-generic}} is defined in
terms of an arithmetic circuit; when the circuit contains constants, it is
sometimes necessary to normalize those constants to ensure that the circuit's
output, when run on a valid input, is the same regardless of the number of
shares.
```

However it currently doesn't do that very well, since we don't actually do any input splitting.

divergentdave:
This improves `run_flp()` to actually split the input and proof into shares, query each, and reassemble the verifier before running the decision algorithm.

Fixes #227.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1615,8 +1615,26 @@ def run_flp(Flp, inp: Vec[Flp.Field], num_shares: Unsigned):
     # Prover generates the proof.
     proof = Flp.prove(inp, prove_rand, joint_rand)
 
-    # Verifier queries the input and proof.
-    verifier = Flp.query(inp, proof, query_rand, joint_rand, num_shares)
+    # Shard the input and the proof.
+    input_shares = linear_secret_share(inp, num_shares, Flp.Field)
+    proof_shares = linear_secret_share(proof, num_shares, Flp.Field)
+
+    # Verifier queries the input shares and proof shares.
+    verifier_shares = [
+        Flp.query(
+            input_share,
+            proof_share,
+            query_rand,
+            joint_rand,
+            num_shares,
+        )
+        for input_share, proof_share in zip(input_shares, proof_shares)
+    ]
+
+    # Combine the verifier shares into the verifier.
+    verifier = Flp.Field.zeros(len(verifier_shares[0]))
+    for verifier_share in verifier_shares:
+        verifier = vec_add(verifier, verifier_share)
 
     # Verifier decides if the input is valid.
     return Flp.decide(verifier)
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -518,6 +518,11 @@ Some common functionalities:
 * `next_power_of_2(n: Unsigned) -> Unsigned` returns the smallest integer
   greater than or equal to `n` that is also a power of two.
 
+* `additive_secret_share(vec: Vec[Field], num_shares: Unsigned, field: type)
+  -> Vec[Vec[Field]]` takes a vector of field elements and returns multiple
+  vectors of the same length, such that they all add up to the input vector,
+  and each vector on its own is indistinguishable from uniform randomness.
+
 # Overview
 
 ~~~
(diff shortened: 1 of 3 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -1615,8 +1615,26 @@ def run_flp(Flp, inp: Vec[Flp.Field], num_shares: Unsigned):
     # Prover generates the proof.
     proof = Flp.prove(inp, prove_rand, joint_rand)
 
-    # Verifier queries the input and proof.
-    verifier = Flp.query(inp, proof, query_rand, joint_rand, num_shares)
+    # Shard the input and the proof.
+    input_shares = linear_secret_share(inp, num_shares, Flp.Field)
```
cjpatton, line 1619: It's pretty self-explanatory what `linear_secret_share` does, but it's a good idea to at least define it somewhere in prose for completeness. One option is to stick in the list of functions in {{conventions-and-definitions}}. In particular we should define what we mean by "linear" and "secret share".

cjpatton (changes requested):

draft-irtf-cfrg-vdaf.md
divergentdave: Sounds good. I think I should also be calling it "additive secret sharing", as it's more specific.

cjpatton: Agreed. These are sometimes used interchangeably, but it's good for the standard to be specific.

draft-irtf-cfrg-vdaf.md
```diff
@@ -518,6 +518,11 @@ Some common functionalities:
 * `next_power_of_2(n: Unsigned) -> Unsigned` returns the smallest integer
   greater than or equal to `n` that is also a power of two.
 
+* `additive_secret_share(vec: Vec[Field], num_shares: Unsigned, field: type)
+  -> Vec[Vec[Field]]` takes a vector of field elements and returns multiple
+  vectors of the same length, such that they all add up to the input vector,
+  and each vector on its own is indistinguishable from uniform randomness.
```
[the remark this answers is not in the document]
divergentdave: I don't think we can express in Python that the type of an argument or return value is related to the value of a type object argument passed in to the function. That would require dependent types more powerful than what type hints can currently express. For example, the following passes `mypy`:

```
class A:
    pass


class B(A):
    pass


def func(thetype: type, arg: A) -> A:
    return arg


func(B, B())
```

Changing the signature to `def func(thetype: type, arg: thetype) -> thetype` or `def func(thetype: type, arg: "thetype") -> "thetype"` results in an error, `error: Name "thetype" is not defined`.

Moreover, in `encode_vec`, the argument `Field` is shadowing the class definition `Field`, so the type hint in the function parameter points to the class, while the `Field.ENCODED_SIZE` attribute access inside the function body points to the argument. We should get rid of variable name shadowing as a general matter.

Issue #220, opened 2023-04-05
Review: 05: Chris Wood: Restrictions on aggregation parameter


cjpatton:
> * `Daf.is_valid(agg_param: AggParam, previous_agg_params: Vec[AggParam]) ->
  Bool`: Checks if the `agg_param` is compatible with all elements of
  `previous_agg_params`.
>
> Is it important that `previous_agg_params` is a list as opposed to a set?

Similar question applies for VDAFs.

cjpatton:
For existing (V)DAFs the answer is no. I also suspect this unlikely to come up in the future.

PR #235, opened 2023-05-05
Change is_valid() argument from list to set

divergentdave:
This changes the second argument to `is_valid()` from a list of aggregation parameters to a set of aggregation parameters, closing #220. I also had to change Poplar1's aggregation parameter to use a tuple of prefixes, instead of a list, because lists are mutable, and therefore not hashable.


cjpatton (approved):
Hmm, given the change

PR #240, opened 2023-05-23
Describe the ping-pong communication pattern used in DAP

cjpatton:
Defines three new API calls on `Vdaf` used to describe state transitions while preparing a report.

1. `prep_start()` is a derived method that defines the initial state transition `START` -> `CONTINUED(host_prep_state)`

2. `ping_pong_prep_next()` is a derived method that defines all subsequent state transitions `CONTINUED(host_prep_state)` -> `FINISHED(out_share)` in the "ping pong" flow being discussed for DAP-05 (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393).

3. `prep_size()` is a required method that indicates the size of the prep message for a given round. This is needed in order to coalesce prep shares and prep messages into the same, opaque byte string.

This allows us to expose the simplest possible interface to DAP.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1190,6 +1190,252 @@ how the nonces are chosen, but security requires that the nonces be unique. See
 execution of a VDAF requires the application to instantiate secure channels
 between each of the protocol participants.
 
+## Communication Patterns for Preparation {#vdaf-prep-comm}
+
+In each round of preparation, each Aggregator writes a prep share to the
+channel, which is then processed into the prep message using the public
+`prep_shares_to_prep()` algorithm and broadcast to the Aggregators to start the
+next round. In this section we describe some approaches to realizing this
+broadcast channel functionality in protocols that use VDAFs with at least one
+round of preparation.
+
+The state machine of each Aggregator for VDAF preparation is shown in
+{{vdaf-prep-state-machine}}.
+
+~~~ state
+               +----------------+
+               |                |
+               v                |
+Start ------> Continued(prep_state) --> Finished(out_share)
+ |                |
+ |                |
+ +--> Rejected <--+
+~~~
+{: #vdaf-prep-state-machine title="State machine for VDAF preparation."}
+
+State transitions are made when the state is acted upon by the host's local
+inputs and/or messages sent by the peers. The initial state is `Start`. The
+terminal states are `Rejected`, indicating that the report was rejected and
+cannot be processed any further, and `Finished(out_share)`, indicating that
+the Aggregator has recovered an output share `out_share`.
+
+## Ping-Pong Topology (Only Two Aggregators)
+
+For VDAFs with precisely two Aggregators (i.e., `Vdaf.SHARES == 2`), the
+following "ping pong" communication pattern can be used. It is compatible with
+any request/response transport protocol, such as HTTP.
+
+The first state transition, from `Start` to `Continued` or `Rejected`, is
+induced by the following transition rule. No messages are sent or received
+during this transition.
+
+~~~ transition
+def ping_pong_start(Vdaf,
+                    vdaf_verify_key: bytes[Vdaf.VERIFY_KEY_SIZE],
+                    is_leader: bool,
+                    agg_param: Vdaf.AggParam,
+                    nonce: bytes[Vdaf.NONCE_SIZE],
+                    public_share: bytes,
+                    host_input_share: bytes) -> State:
+    try:
+        prep_state = Vdaf.prep_init(
+            vdaf_verify_key,
+            0 if is_leader else 1,
+            agg_param,
+            nonce,
+            public_share,
+            host_input_share,
+        )
+        return Continue(prep_state)
+    except:
+        return Rejected()
+~~~
+
+If the state is `Rejected`, then processing halts. Otherwise, if the state is
+`Continued`, then processing continues using the following transition rule until
+a terminal state is reached.
+
+Let us call the initiating party the "Leader" and the responding party the
+"Helper". The high-level idea is that the Leader and Helper will take turns
+running the computation locally until input from their peer is required:
+
+* For a 1-round VDAF (e.g., Prio3 ({{prio3}})), the Leader sends its prep share
+  to the Helper, who computes the prep message locally, computes its output
+  share, then sends the prep message to the Leader. Preparation requires just
+  one round trip between the Leader and the Helper.
+
+* For a 2-round VDAF (e.g., Poplar1 ({{poplar1}})), the Leader sends its
+  first-round prep share to the Helper, who replies with the first-round prep
+  message and its second-round prep share. In the next request, the Leader
+  computes its second-round prep share locally, computes its output share, and
+  sends the second-round prep message to the Helper. Finally, the Helper
+  computes its own output share.
+
+* In general, each request includes the Leader's prep share for the previous
+  round and/or the prep message for the current round; correspondingly, each
+  response consists of the prep message for the current round and the Helper's
+  prep share for the next round.
+
+The Aggregators proceed in this ping-ponging fashion until a step of the
+computation fails, indicating the report is invalid and should be rejected, or
+preparation is completed. All told there there are `ceil((Vdaf.ROUNDS+1)/2)`
+requests sent.
+
+Each message in the ping-pong protocol is structured as follows (expressed in
+TLS syntax {{Section 3 of !RFC8446}}):
+
+~~~
+enum {
+  initialize(0),
+  continue(1),
+  finish(2),
+  (255)
+} MessageType;
+
+struct {
+  MessageType type;
+  select (Message.type) {
+    case initialize:
+      opaque prep_share<0..2^32-1>;
+    case continue:
+      opaque prep_msg<0..2^32-1>;
+      opaque prep_share<0..2^32-1>;
+    case finish:
+      opaque prep_msg<0..2^32-1>;
+  };
+} Message;
+~~~
+
+The Leader computes each state transition according to the following algorithm:
+
+~~~ transition
+def ping_pong_req(Vdaf,
+                  agg_param: Vdaf.AggParam,
+                  state: State,
+                  inbound: Optional[Message],
+                  ) -> (State, Optional[Message]):
+    if inbound == None:
+        prep_msg = None
+        peer_prep_share = None
+    elif inbound.type == 1: # continue
+        prep_msg = inbound.prep_msg
+        peer_prep_share = inbound.prep_share
+    elif inbound.type == 2: # finish
+        prep_msg = inbound.prep_msg
+        peer_prep_share = None
+    else:
+        return (Rejected(), None)
+
+    return Vdaf.ping_pong_transition(
+        agg_param,
+        prep_msg,
+        peer_prep_share,
+        state.prep_state,
+    )
+~~~
+
+(The auxiliary function `ping_pong_transition()` is defined at the end of this
+section.) The input `inbound` denotes the last message received from the Helper. This
+parameter is optional since initially there is no inbound message.
+
+Likewise, the Helper computes each state transition according to the following
+algorithm:
+
+~~~ transition
+def ping_pong_resp(Vdaf,
+                   agg_param: Vdaf.AggParam,
+                   state: State,
+                   inbound: Message,
+                   ) -> (State, Optional[bytes]):
+    if inbound.type == 0: # initialize
+        prep_msg = None
+        peer_prep_share = inbound.prep_share
+    elif inbound.type == 1: # continue
+        prep_msg = inbound.prep_msg
+        peer_prep_share = inbound.prep_share
+    else: # finish
+        prep_msg = inbound.prep_msg
+        peer_prep_share = None
+
+    return Vdaf.ping_pong_transition(
+        agg_param,
+        prep_msg,
+        peer_prep_share,
+        state.prep_state,
+    )
+~~~
+
+At the start of each request/response cycle, the Leader runs:
+
+~~~
+(state, outbound) = Vdaf.ping_pong_req(agg_param, state, inbound)
+~~~
+
+with `state == Continued(leader_prep_state)` and, if `outbound != None`, sends
+`outbound` to the Helper. For the initial request, `inbound == None`. To
+respond to the Leader, the Helper runs
+
+~~~
+(state, outbound) = Vdaf.ping_pong_resp(agg_param, state, inbound)
+~~~
+
+with `state == Continued(helper_prep_state)`, where `inbound` is the message it
+received from the Leader, and sends `outbound` to the Leader.
+
+If `state == Finished(out_share)` at the end of a request/response cycle, then
+processing is complete. Note that, depending on the number of rounds of
+preparation that are required, there may be one more message to send before the
+peer can also finish processing (i.e., `outbound != None`).
+
+The core state transition logic is the same for both Aggregators:
+
+~~~
+def ping_pong_transition(Vdaf,
+                         agg_param: Vdaf.AggParam,
+                         prep_msg: Optional[bytes],
+                         peer_prep_share: Optional[bytes],
+                         host_prep_state: Vdaf.Prep,
+                         ) -> (State, Optional[Message]):
+    try:
+        # If `prep_msg == None` then this is the start of the
+        # first round. Otherwise, `prep_msg` is the prep message
+        # from the end of the previous round.
+        out = Vdaf.prep_next(host_prep_state, prep_msg)
+        if type(out) == Vdaf.OutShare:
+            return (Finished(out), None)
+        (host_prep_state, host_prep_share) = out
+
+        if peer_prep_share != None:
+            prep_shares = [peer_prep_share, host_prep_share]
+            if is_leader:
+                prep_shares.reverse()
+            prep_msg = Vdaf.prep_shares_to_prep(
+                agg_param,
+                prep_shares,
+            )
+            out = Vdaf.prep_next(host_prep_state, prep_msg)
+            if type(out) == Vdaf.OutShare:
+                outbound = Message.finish(prep_msg)
+                return (Finished(out), outbound)
+            (host_prep_state, host_prep_share) = out
+            outbound = Message.continue(prep_msg, host_prep_share)
+        else:
+            outbound = Message.initialize(host_prep_share)
+        return (Continued(host_prep_state), outbound)
+    except:
+        return (Rejected(), None)
+~~~
+
+## Star Topology (Any Number of Aggregators)
+
+The ping-pong topology of the previous section is only suitable for VDAFs
+involving exactly two Aggregators. If more Aggregators are required, the star
+topology described in this section can be used instead.
+
+> TODO Describe the Leader-emulated broadcast channel architecture that was
+> originally envisioned for DAP. (As of DAP-05 we are going with the ping pong
+> architecture described in the previous section.)
+
 # Preliminaries {#prelim}
 
 This section describes the primitives that are common to the VDAFs specified in
@@ -3743,8 +3989,8 @@ analysis of {{DPRS23}}. Thanks to Hannah Davis and Mike Rosulek, who lent their
 time to developing definitions and security proofs.
 
 Thanks to Henry Corrigan-Gibbs, Armando Faz-Hernández, Simon Friedberger, Tim
-Geoghegan, Mariana Raykova, Jacob Rothstein, Xiao Wang, and Christopher Wood for
-useful feedback on and contributions to the spec.
+Geoghegan, Brandon Pitman, Mariana Raykova, Jacob Rothstein, Xiao Wang, and
+Christopher Wood for useful feedback on and contributions to the spec.
 
 # Test Vectors {#test-vectors}
 {:numbered="false"}


draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+    try:
+        host_prep_state = Vdaf.prep_init(
+            vdaf_verify_key,
+            0 if is_leader else 1,
+            agg_param,
+            nonce,
+            public_share,
+            host_input_share,
+        )
+        state CONTINUED(0, host_prep_state)
```
cjpatton, line 1243: Reviewer note: This obviously is not grammatically correct Python. Our goal thus far was to have all pseudocode in the draft be runnable, so we might consider how to do so neatly. However in the interest of moving DAP-05 forward, I would prefer to not be too nit-picky, at least for the moment.

cjpatton:
@branlwyd PTAL to make sure this is correct. If so, then I think we'd remove the state machine from https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/461, merge, then rewrite https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393 to use this new API.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+def ping_pong_prep_next(Vdaf,
+                        is_leader: bool,
+                        agg_param: Vdaf.AggParam,
+                        inbound_type: Optional[str],
+                        inbound: Optional[bytes]):
+    if inbound_type == "initialize":
+        prep_msg, peer_prep_share = None, inbound
+    elif inbound_type == "continue":
+        l = Vdaf.prep_size(agg_param, round)
+        prep_msg, prep_share = inbound[:l], inbound[l:]
```
cjpatton, line 1298: @branlwyd note that we're coalescing the previous prep message with the next prep share here. This is a change from https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/461.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                        is_leader: bool,
+                        agg_param: Vdaf.AggParam,
+                        inbound_type: Optional[str],
+                        inbound: Optional[bytes]):
+    if inbound_type == "initialize":
+        prep_msg, peer_prep_share = None, inbound
+    elif inbound_type == "continue":
+        l = Vdaf.prep_size(agg_param, round_num)
+        prep_msg, prep_share = inbound[:l], inbound[l:]
+        round_num += 1
```
cjpatton, line 1299: @branlwyd to ensure that the round number is computed properly I believe we need increment the counter here. Please double check my math here. (This is a change from https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/461.)

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+"Helper". The high-level idea is that the Leader and Helper will take turns
+running the computation locally until input from their peer is required:
+
+* For a 1-round VDAF (e.g., Prio3 ({{prio3}})), The Leader sends its prep share
+  to the Helper, who computes the prep message locally (the computations are
+  specified in the following sections), computes its output share, then sends
+  the prep message to the Leader. Preparation requires just one request from
+  the Leader to the Helper.
+
+* For a 2-round VDAF (e.g., Poplqr1 ({{poplar1}})), the Leader sends its
```
divergentdave, line 1269: nit: typo
```suggestion
* For a 2-round VDAF (e.g., Poplar1 ({{poplar1}})), the Leader sends its
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+To initialize preparation, the Leader would call
+
+~~~
+(outbound_type, outbound) = Vdaf.ping_pong_prep_next(True,
+                                                     agg_param,
+                                                     None,
+                                                     None)
+~~~
+
+in state `CONTINUED(0, leader_prep_state)` and send `outbound` to the Helper.
```
divergentdave, line 1352: This should probably be reworded, "to initialize preparation" seems to be counter to already being in the state CONTINUED. If anything is initializing preparation, it seems like it ought to be `prep_start`. Maybe, "At the start of a request/response cycle, ..."?

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+The protocol might also include an indication of what type of message to expect
+(i.e., some encoding of `outbound_type`).
```
divergentdave, line 1354: Shouldn't the outbound type be known based on the number of rounds and the current round? This sentence seems to hedge enough that we could probably delete it.

draft-irtf-cfrg-vdaf.md
branlwyd: Thinking to a large-round VDAF, this round-counting scheme tracks a consistent sequence of rounds between the Leader & Helper; specifically 1→3→5→...

But I think this line should be dropped, so that "continue" messages only increment the round counter once, and the Leader & Helper compute the round sequence 1→2→3→...

The reason I think this is a good change is that the purpose of the round-counter is to track progress/allow for retries at the level of the VDAF-evaluation protocol (e.g. DAP). If the sequence of rounds is 1→3→5→..., then the higher-level protocol has a failure mode if an attempt is made to retry/advance to an even-numbered round; this failure mode does not exist if the round sequence is 1→2→3→...

(I'm not sure we need to include round counting at all in these functions. I'd be happy if these functions defined the sequence of VDAF calls, and left the definition of any retry mechanisms up to the VDAF-evaluation protocol. ISTM that a hypothetical DAP-competitor protocol might define a different retry mechanism, at which point this round-counting is vestigal. Also, these functions are defining a round counter for each report, while even DAP tracks round at the level of an aggregation job, i.e. across many reports.)

branlwyd (changes requested):

draft-irtf-cfrg-vdaf.md
cjpatton: Right, my goal here was to have `round_num` match the VDAF round, not the ping-pong round. If we go with your suggestion then we'd need to change
```python
if round_num == Vdaf.ROUNDS:
```
to
```python
if round_num = ceil((Vdaf.ROUNDS+1)/2):
```
Does that sound right?

draft-irtf-cfrg-vdaf.md
cjpatton: This is tricky because we're trying to get as close to runnable code as possible. But your suggestion here matches @divergentdave's suggestion above pretty well, so let me see if I can finesse this.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1194,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+               vdaf_verify_key: bytes[Vdaf.VERIFY_KEY_SIZE],
+               is_leader: bool,
+               agg_param: Vdaf.AggParam,
+               nonce: bytes[Vdaf.NONCE_SIZE],
+               public_share: bytes,
+               host_input_share: bytes):
+    try:
+        host_prep_state = Vdaf.prep_init(
+            vdaf_verify_key,
+            0 if is_leader else 1,
```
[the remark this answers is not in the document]
cjpatton: Sorry, I'm not sure I understand this comment.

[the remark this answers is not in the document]
divergentdave: Since prep_start can only pass an aggregator ID of 0 or 1 to prep_init, it should go under "Ping-Pong Topology (Only Two Aggregators)", rather than this more generic section.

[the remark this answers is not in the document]
cjpatton: Ohhhhh good call!

draft-irtf-cfrg-vdaf.md
cjpatton: Done. Changed to "At the start of the **first** request/response cycle..."

draft-irtf-cfrg-vdaf.md
cjpatton: Following @branlwyd's suggestion I just dropped the inbound/outbound type indication and simplified the text here.

draft-irtf-cfrg-vdaf.md
cjpatton: Ahhh you're right! What would be useful for DAP would be the "request number" instead of the round number. I started working on this change but found a problem: We now have a callback on `Vdaf` for getting the size of the prep message (`Vdaf.prep_size()`) and unfortunately this needs the round number as input.

cjpatton: One way around this would be to length-prefix the prep message in the coalesced `Continue` message type so that we can parse the two. It sort of breaks with the design pattern we've been following for VDAF but I'd be fine with this if you think tracking the request number would be useful.

cjpatton: Good call :)

I've removed `Vdaf.prep_size()` and removed the round counter from the state.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+def ping_pong_next(Vdaf,
+                   state: State,
+                   is_leader: bool,
+                   agg_param: Vdaf.AggParam,
+                   inbound: Optional[Message],
+                   ) -> (State, Optional[Message]):
+    if type(inbound) == InitializeMessage:
+        prep_msg, peer_prep_share = None, inbound.prep_share
+    elif type(inbound) == ContinueMessage:
+        prep_msg, prep_share = inbound.prep_msg, inbound.prep_share
```
divergentdave, line 1289: These ought to match
```suggestion
        prep_msg, peer_prep_share = inbound.prep_msg, inbound.prep_share
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+    elif type(inbound) == FinishMessage:
+        prep_msg, peer_prep_share = inbound.prep_msg, None
+    else:
+        prep_msg, peer_prep_share = None, None
+
+    try:
+        # If prep_msg == None then this is the start of the
+        # first round. Otherwise, this is the prep message
```
divergentdave, line 1297: nit: better parallelism ("this" refers to what round it is earlier in the sentence, not `prep_msg`)
```suggestion
        # first round. Otherwise, `prep_msg` is the prep message
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+        # from the end of the previous round.
+        out = Vdaf.prep_next(host_prep_state, prep_msg)
+        if type(out) == Vdaf.OutShare:
+            # The Leader reaches this state upon processing a
+            # response from the Helper, so there is no outbound
+            # message.
+            outbound = None if is_leader else FinishedMessage()
```
divergentdave, line 1304: We should probably enumerate and describe the different message types before this pseudocode, because the difference between FinishMessage and FinishedMessage is easy to miss.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                   agg_param: Vdaf.AggParam,
+                   inbound: Optional[Message],
+                   ) -> (State, Optional[Message]):
+    if type(inbound) == InitializeMessage:
+        prep_msg, peer_prep_share = None, inbound.prep_share
+    elif type(inbound) == ContinueMessage:
+        prep_msg, prep_share = inbound.prep_msg, inbound.prep_share
+    elif type(inbound) == FinishMessage:
+        prep_msg, peer_prep_share = inbound.prep_msg, None
+    else:
```
divergentdave, line 1292: We should be explicit about what message type this branch is for, for clarity.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+~~~
+outbound = Vdaf.ping_pong_next(False,
+                               state,
+                               agg_param,
+                               InitializeMessage(inbound))
+~~~
+
+with `state == Continued(helper_prep_state)`, where `inbound` was the message
+it received from the Leader, and send `outbound` to the Leader. All subsequent
+transitions are computed by calling `Vdaf.ping_pong_next(is_leader, agg_param,
```
divergentdave, line 1349: nit: missing the state argument
```suggestion
transitions are computed by calling `Vdaf.ping_pong_next(is_leader, state, agg_param,
```

draft-irtf-cfrg-vdaf.md
branlwyd: I agree -- it would be very nice to describe these messages along with their purpose/contents.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+            outbound = ContinueMessage(prep_msg, host_prep_share)
+        return (Continued(host_prep_state), outbound)
+    except:
+        return (Rejected(), None)
+~~~
+
+At the start of the first request/response cycle, the Leader would call
+
+~~~
+outbound = Vdaf.ping_pong_next(True, state, agg_param, None)
```
branlwyd, line 1333: nit: `state` comes before `is_leader`, here and on L1341.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  |                |                  |
+  |                |                  v
+  +--> Rejected <--+  Finished(out_share)
+~~~
+{: #vdaf-prep-state-machine title="State machine for VDAF preparation."}
+
+State transitions are made when the state is acted upon by the host's local
+inputs and/or messages sent by the peers. The initial state is `Start`. The
+terminal states are `Rejected`, indicating that the report was rejected and
+cannot be processed any further; and `Rejected(out_share)` where `out_share` is
```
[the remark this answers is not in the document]
cjpatton: 😬

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+channel, which is then processed into the prep message using the public
+`prep_shares_to_prep` algorithm and broadcast to the Aggregators to start the
+next round. In this section we describe some approaches to realizing this
+braodcast channel functionality in protocols that use VDAFs.
+
+The state machine of each Aggregator for VDAF preparation is shown in
+{{vdaf-prep-state-machine}}.
+
+~~~ state
+ Start ------> Continued(host_prep_state)
```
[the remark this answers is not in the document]

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+
+If the state is `Rejected`, then processing halts. Otherwise, if the state is
+`Continued`, then processing continues using the following transition rule until
+a terminal state is reached.
+
+Let us call the initiating party the "Leader" and the responding party the
+"Helper". The high-level idea is that the Leader and Helper will take turns
+running the computation locally until input from their peer is required:
+
+* For a 1-round VDAF (e.g., Prio3 ({{prio3}})), The Leader sends its prep share
```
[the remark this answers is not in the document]

draft-irtf-cfrg-vdaf.md

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,175 @@ how the nonces are chosen, but security requires that the nonces be unique. See
 execution of a VDAF requires the application to instantiate secure channels
 between each of the protocol participants.
 
+## Communication Patterns for Preparation {#vdaf-prep-comm}
+
+In each round of preparation, each Aggregator writes a prep share to the
+channel, which is then processed into the prep message using the public
+`prep_shares_to_prep` algorithm and broadcast to the Aggregators to start the
+next round. In this section we describe some approaches to realizing this
+braodcast channel functionality in protocols that use VDAFs.
```
branlwyd, line 1199:
```suggestion
broadcast channel functionality in protocols that use VDAFs.
```

draft-irtf-cfrg-vdaf.md
cjpatton: Added a comment (should be `inbound == None` and `type(inbound) == FinishedMessage()`

draft-irtf-cfrg-vdaf.md
cjpatton: Done. I also put prep_msg == None above in backticks.

draft-irtf-cfrg-vdaf.md
cjpatton: Good catch.

draft-irtf-cfrg-vdaf.md
cjpatton: Done.

draft-irtf-cfrg-vdaf.md
cjpatton: Done! While at it I realized we maybe don't need `FinishedMessage` at this level of abstraction-- We could just have the Helper respond with `None`. Of course, in DAP we'll have to distinguish between "finish(prep_msg)" and "finished" in the wire protocol.

Let me know what you think -- if desired I can add this back in, I don't feel strongly about it either way.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  prep share for the next round.
+
+The Aggregators proceed in this ping-ponging fashion until a step of the
+computation fails, indicating the report is invalid and should be rejected, or
+the last round is reached and each Aggregator computes an output share. All
+told there there are `ceil((Vdaf.ROUNDS+1)/2)` requests sent.
+
+The message types are as follows:
+
+* `InitializeMessage(prep_share)` - Sent by the Leader to the Helper in the
```
cjpatton, line 1284: WDTA about "InitializeMessage" -> "PingPongInitialize", etc.?

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,190 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                   state: State,
+                   inbound: Optional[Message],
+                   ) -> (State, Optional[Message]):
+    if type(inbound) == InitializeMessage:
+        prep_msg, peer_prep_share = None, inbound.prep_share
+    elif type(inbound) == ContinueMessage:
+        prep_msg, peer_prep_share = inbound.prep_msg, inbound.prep_share
+    elif type(inbound) == FinishMessage:
+        prep_msg, peer_prep_share = inbound.prep_msg, None
+    else: # inbound == None
```
branlwyd, line 1310: What do you think about changing the `None` case to be an explicit `FinishedMessage`? This would mirror what's in DAP, and would make it much clearer here which case is being handled. In practice, any request/response protocol will need to send _some_ message, so I think it makes some sense to define what that message is as `FinishedMessage`. (name is up for debate, just suggesting what we went with in the analogous scenario in DAP)

draft-irtf-cfrg-vdaf.md
branlwyd: SGTM, reduces the odds that these names somehow conflict with another part of the document.

branlwyd (approved):
LGTM, with one suggestion (happy to review another round if you decide to take it).

draft-irtf-cfrg-vdaf.md
cjpatton: See this comment: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240#discussion_r1204848392. I'll add it back.

cjpatton: Done!

draft-irtf-cfrg-vdaf.md

branlwyd (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,198 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                   inbound: Optional[Message],
+                   ) -> (State, Optional[Message]):
+    if type(inbound) == PingPongMessage:
+        prep_msg, peer_prep_share = None, inbound.prep_share
+    elif type(inbound) == PingPongContinue:
+        prep_msg = inbound.prep_msg
+        peer_prep_share = inbound.prep_share
+    elif type(inbound) == PingPongFinish:
+        prep_msg, peer_prep_share = inbound.prep_msg, None
+    else: # inbound == None, type(inbound) == PingPongFinished
```
divergentdave, line 1314: I think `type(inbound) == PingPongFinished` here should be unreachable when executing this between two honest aggregators. (setting that aside, calling prep_next in such a situation would be incorrect) Only the helper sends PingPongFinished, thus only the leader can receive it. Walking backwards in the protocol, the leader would have transitioned to the Finished state before sending PingPongFinish for the prior round, so the leader would no longer be in the Continued state. Thus, if the leader receives a PingPongFinished message, it will not be calling `ping_pong_next()` with it as the inbound message.

We should have a separate branch here that transitions to Rejected and returns if either aggregator receives PingPongFinished while they're still in the Continued state.

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: Done. I added the rejection branch, but did not condition on being in the Continued state, since we don't expect ping_pong_next() to be called in any ogther state.

cjpatton: Oh wait, I'm now realizing htat's wrong; the Leder will be in Finished() while awaiting PingPongFinished().

cjpatton: On third thought, currently we say that after reaching Finished the Leader won't await any other messages. I think this fine, but we might consider having the Leader go to Continued while waiting for PingPongFinished. WDYT?

branlwyd: I think we can have the Leader go to Finished; going to Continued is only tricky because it stores a prep state instead of an out share (so we'd really have two states, a "normal continued" state and a "continued awaiting finished" state, I suppose).

Maybe we should just say that once the Leader goes to the Finished state, it should expect one more PingPongFinished message? I think that is effectively the semantics we want.

divergentdave: Having the leader go to Finished makes sense to me -- otherwise we'd need two separate state machines for the leader and helper. The helper only sends PingPongFinished because we have to send something in a request-response transport. Since the leader should already be in a terminal state machine state, the message can be dropped on the floor -- the VDAF is already done as far as all parties are concerned.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  the current round and the sender's prep share for the next round.
+
+* `PingPongFinish(prep_msg)` - Sent by the Leader in the last request
+  (even-round VDAF) or the Helper in the last response (odd-round VDAF). The
+  payload is the final-round prep message.
+
+* `PingPongFinished()` - Sent by the Helper in the final response in even-round
+  VDAFs. There is no payload.
+
+Each Aggregator handles these messages according to the following transition rule:
```
divergentdave, line 1300: We should be explicit about when this state transition rule is applicable.
```suggestion
Each Aggregator handles these messages according to the following transition rule,
so long as they are in the `Continued` state:
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+inputs and/or messages sent by the peers. The initial state is `Start`. The
+terminal states are `Rejected`, indicating that the report was rejected and
+cannot be processed any further; and `Finished(out_share)` where `out_share` is
+the Aggreagtor's recovered output share.
+
+## Ping-Pong Topology (Only Two Aggregators)
+
+For VDAFs with precisely two Aggregators (i.e., `Vdaf.SHARES == 2`), the
+following "ping pong" communication pattern can be used. It is compatible with
+any request/response protocol, such as HTTP.
```
tgeoghegan, line 1225:
```suggestion
any request/response transport protocol, such as HTTP.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+    try:
+        host_prep_state = Vdaf.prep_init(
+            vdaf_verify_key,
+            0 if is_leader else 1,
+            agg_param,
+            nonce,
+            public_share,
+            host_input_share,
+        )
+        return Continue(host_prep_state)
```
tgeoghegan, line 1248: At the risk of being a pedant: if the VDAF being run was 0-round, then `Vdaf.prep_init` would transition right to `Finished`. I suppose you wouldn't bother running the ping-pong mode for a 0-round VDAF at all, but I feel like either this section could state that it's only defined for VDAFs with at least one round, or `{{vdaf}}` might say that `VDAF.ROUNDS > 0`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+`Continued`, then processing continues using the following transition rule until
+a terminal state is reached.
+
+Let us call the initiating party the "Leader" and the responding party the
+"Helper". The high-level idea is that the Leader and Helper will take turns
+running the computation locally until input from their peer is required:
+
+* For a 1-round VDAF (e.g., Prio3 ({{prio3}})), the Leader sends its prep share
+  to the Helper, who computes the prep message locally (the computations are
+  specified in the following sections), computes its output share, then sends
```
tgeoghegan, line 1263: nit: explicit section links would be nice

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  the prep message to the Leader. Preparation requires just one request from
+  the Leader to the Helper.
```
tgeoghegan, line 1265: nit: it also requires a response, so maybe
```suggestion
  the prep message to the Leader. Preparation requires just one round trip between
  the Leader and the Helper.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+
+* For a 2-round VDAF (e.g., Poplar1 ({{poplar1}})), the Leader sends its
+  first-round prep share to the Helper, who replies with the first-round prep
+  message and its second-round prep share. In the next request, the Leader
+  computes its second-round prep share locally, computes its output share, and
+  sends the second-round prep message to the Helper. Finally, the Helper
+  computes its own output share. Preparation requires two requests.
```
tgeoghegan, line 1272:
```suggestion
  computes its own output share. Preparation requires two round trips.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                   state: State,
+                   inbound: Optional[Message],
+                   ) -> (State, Optional[Message]):
+    if type(inbound) == PingPongMessage:
+        prep_msg, peer_prep_share = None, inbound.prep_share
+    elif type(inbound) == PingPongContinue:
+        prep_msg = inbound.prep_msg
+        peer_prep_share = inbound.prep_share
+    elif type(inbound) == PingPongFinish:
+        prep_msg, peer_prep_share = inbound.prep_msg, None
```
tgeoghegan, line 1315: nit: all the branches should consistently do either

```
foo, bar = baz, qux
```

or

```
foo = baz
bar = qux
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+peer's inbound message.
+
+## Star Topology (Any Number of Aggregators)
+
+The ping-pong topology of the previous section is only suitable for VDAFs
+involving exactly two Aggregators. If more Aggregators are required, the star
+topology described in this section can be used instead.
+
+> TODO Describe the Leader-emulated broadcast channel architecture that was
+> origianlly envisioned for DAP. (As of DAP-05 we are going with the ping pong
```
tgeoghegan, line 1386:
```suggestion
> originally envisioned for DAP. (As of DAP-05 we are going with the ping pong
```

tgeoghegan (approved):
some nitty nits but this looks sound to me!

draft-irtf-cfrg-vdaf.md
cjpatton: SGTM. Added a note about even-round VDAFs.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,202 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+ |                |
+ +--> Rejected <--+
+~~~
+{: #vdaf-prep-state-machine title="State machine for VDAF preparation."}
+
+State transitions are made when the state is acted upon by the host's local
+inputs and/or messages sent by the peers. The initial state is `Start`. The
+terminal states are `Rejected`, indicating that the report was rejected and
+cannot be processed any further; and `Finished(out_share)` where `out_share` is
+the Aggreagtor's recovered output share.
```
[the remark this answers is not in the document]
divergentdave: ^bumping this

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
tgeoghegan: Well, whichever the "following sections" mentioned on this line are.

cjpatton: Ohhh got it. Looking at this once more I think the paranethetical is kind of confusing so I just deleted it.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,206 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  prep share for the next round.
+
+The Aggregators proceed in this ping-ponging fashion until a step of the
+computation fails, indicating the report is invalid and should be rejected, or
+the last round is reached and each Aggregator computes an output share. All
+told there there are `ceil((Vdaf.ROUNDS+1)/2)` requests sent.
+
+The message types are as follows:
+
+* `PingPongMessage(prep_share)` - Sent by the Leader to the Helper in the
```
branlwyd, line 1286: nitpick: could this message be called something like `PingPongInitialize`? `PingPongMessage` is a more generic name than necessary IMO -- all of these message types could equally-accurately be called `PingPongMessage`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,206 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                return (Finished(out), PingPongFinish(prep_msg))
+            (host_prep_state, host_prep_share) = out
+
+        if is_leader and inbound == None:
+            outbound = PingPongInitialize(host_prep_share)
+        else:
+            outbound = PingPongContinue(prep_msg, host_prep_share)
+        return (Continued(host_prep_state), outbound)
+    except:
+        return (Rejected(), None)
```
branlwyd, line 1354: WDYT about introducing a `PingPongRejected` message instead of returning `None`?

Other than being more explicit about the meaning of this return value, the way things are currently written I suppose someone could misinterpret things such that the other aggregator would continue to call `ping_pong_next` with `inbound == None`, which is not what we want.

cjpatton: When would that happen? If I'm in the Rejected state, I'm not going to keep processing messages for that report.

branlwyd: This aggregator would be in `Rejected` state, but the other aggregator wouldn't be in `Rejected` yet. By returning `None` here, someone misinterpreting this text might think that this means they need to send "no message" to the other aggregator, which would then call `ping_pong_next` with `inbound == None`, which would cause the peer aggregator to misbehave.

But this is just my guess at how things might go wrong. My bigger concern is being explicit: `PingPongRejected` has a more explicit meaning than `None`, and will make it easier for protocols using these functions to specify handling.

cjpatton: Hmm, let me think about this. We get to `Rejected()` in three ways:
1. prep_init() raises an error
2. prep_next() or prep_shares_to_prep() raises an error
3. ping_pong_next() is called on `PingPongFinished()`

Whether there is an explicit message sent in DAP kind of depends: in cases (1) and (2) only the Helper sends an explicit message, at least for now (this might change with https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/438). Case (3) should actually be unreachable if the ping pong API is used correctly.

I think if we want to be more explicit, we should be consistent. I.e., in case (1) there should be an explicit outbound message. But the fact that whether there is an explicit message depends on protocol details, I think I'd prefer to leave the type as `None` and leave it to the protocol to sort out.

branlwyd: Good point; one way to resolve this, but perhaps too slowly for this PR, would be to take https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/244 -- then ping-pong initialization could be similarly fused such that `ping_pong_start` would return a state & outbound message, which could then be `PingPongRejected`.

Another option would be not to address this at all: I updated https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393 to use the ping-pong API from this PR, rejection handling isn't too awkward (and not all of the awkwardness would be resolved even if we took this suggestion).

cjpatton: I'd prefer the second option if it's not too awkward. I just saw your changes to 393 and am reviewing now.

cjpatton:
I've updated this PR (https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240/commits/2faf238312af61b95460ac99a9e96593b638f1ae) and harmonized the DAP change (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393/commits/4fe9db2052249a0acf008844a15892f303664046) please take a look.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,213 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+        return (Continued(round, host_prep_state), outbound)
+    except:
+        return (Rejected(), None)
+~~~
+
+Methods `encode_continue()` and `decode_contine()` are used to coalesce a prep
+message and the next prep share:
+
+~~~
+def encode_method(prep_msg: bytes, prep_share: bytes):
```
branlwyd, line 1350: nit: `encode_method` should be `encode_continue` I think (and similarly below)

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,213 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+                   inbound: Optional[bytes],
+                   ) -> (State, Optional[PingPongMessage]):
+    round, host_prep_state = state.round, state.prep_state
+    if inbound != None:
+        if round == 0:
+            if is_leader:
+                return (Rejected(), None)
+            else:
+                prep_msg, peer_prep_share = None, inbound
+        elif round == Vdaf.ROUNDS:
```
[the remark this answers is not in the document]
cjpatton: I'll think on it more, but I'm not totally convinced we can do away with the round number altogether. In any case I think we can make parsing context-free here.

A couple of points:
* I'm not sure notwork failures are or should be a consideration for VDAF. Ideally we can think of VDAF as the primitive building block that is used in a protocol that deals with this stuff (DAP). As an anology, we don't typically think about network failures when writing down the spec for Diffie-Hellman key exchange. (And I'd argue we don't need to.)
* Elsewhere in VDAF, parsing is not context-free: For instance in Prio3 the field being used is not encoded in any message. I'm fine doing so here however, since this is connective tissue for a specific protocol.

draft-irtf-cfrg-vdaf.md
cjpatton: These methods have been deleted.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,213 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+
+Methods `encode_continue()` and `decode_contine()` are used to coalesce a prep
+message and the next prep share:
+
+~~~
+def encode_method(prep_msg: bytes, prep_share: bytes):
+    encoded = bytes()
+    encoded += to_be_bytes(len(prep_msg), 2)
+    encoded += prep_msg
+    encoded += to_be_bytes(len(prep_share), 2)
```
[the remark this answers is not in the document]
cjpatton: I've decided to specify the ping pong messages in TLS syntax. I think to make it easier to use in DAP we want both length prefixes. (That way parsing DAP messages can be done with the `Decode` trait in libprio.)

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
cjpatton: i've added the discriminator byte and removed the round number from the Continue state once more.

cjpatton (commented):
I've taken most of your comments, including making serialization context free. I next need to update the DAP PR to match -- there I think we can just refer to the `Message` stsructure explicitly.

branlwyd (approved):
I like the usage of TLS syntax -- I was wary of suggesting it as I thought that to be too DAP-specific, but if you are happy with it then so am I.

I'm still reviewing the DAP PR but I think this one is good to go.

cjpatton:
> I like the usage of TLS syntax -- I was wary of suggesting it as I thought that to be too DAP-specific, but if you are happy with it then so am I.


Yeah I was wary of doing this as well, but I think here it solves the problem so let's just do it :)

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,252 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  MessageType type;
+  select (Message.type) {
+    case initialize:
+      opaque prep_share<0..2^32-1>;
+    case continue:
+      opaque prep_msg<0..2^32-1>;
+      opaque prep_share<0..2^32-1>;
+    case finish:
+      opaque prep_msg<0..2^32-1>;
+  };
```
divergentdave, line 1305: I would prefer adding the `finished` message type back, as otherwise DAP will need an additional discriminator bit outside the `Message` struct to indicate whether or not a helper's response includes a `Message` or not.

cjpatton: True, but I think that discriminator bit is needed in DAP in any case. Having done a few passes of this, I think this is the simplest design we have. (I'd encourage you to also review the corresponding DAP PR -- perhaps you'll have an idea to simplify things a bit more.)

branlwyd: I think this is OK (if slightly confusing due to similar names being used in DAP & VDAF): `finished` exists at the DAP level, as one of the possible states of a `PrepareStep`, and works as you describe; this VDAF `finish` message would be embedded in a DAP `PrepareStep` of type `continue`.

branlwyd: (that said, this _is_ somewhat confusing; maybe there's a better way to state this abstraction boundary?)

cjpatton: Perhaps. I sort of like what we have though now. It's at least consistent: If there is an actual message to send, then `outbound` will be `None`; otherwise `outbound != None`. This is useful in DAP because the Leader and Helper do different things in this case.

tgeoghegan (commented):
Here's a first pass with some ticky-tacky spelling and terminology nits. I'm going to do another pass where I read this in conjunction with https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393 and may come out of that with more substantive structural comments.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,252 @@ how the nonces are chosen, but security requires that the nonces be unique. See
 execution of a VDAF requires the application to instantiate secure channels
 between each of the protocol participants.
 
+## Communication Patterns for Preparation {#vdaf-prep-comm}
```
tgeoghegan, line 1193: All of this is written in terms of VDAFs. Does that mean this is undefined for DAFs?

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,252 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+~~~
+
+At the start of each request/response cycle, the Leader runs
+
+~~~
+(state, outbound) = Vdaf.ping_pong_req(agg_param, state, inbound)
+~~~
+
+with `state == Continued(leader_prep_state)` and sends `outbound` to the
+Helper. For the initial request, `inbound == None`. To response to the Leader,
```
tgeoghegan, line 1375:
```suggestion
Helper. For the initial request, `inbound == None`. To respond to the Leader,
```

tgeoghegan (changes requested):
I have a couple more nits or suggestions, but I believe this is fundamentally sound and fits with the corresponding DAP change (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393).

draft-irtf-cfrg-vdaf.md
cjpatton: Yes.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,252 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+during this transition.
+
+~~~ transition
+def ping_pong_start(Vdaf,
+                    vdaf_verify_key: bytes[Vdaf.VERIFY_KEY_SIZE],
+                    is_leader: bool,
+                    agg_param: Vdaf.AggParam,
+                    nonce: bytes[Vdaf.NONCE_SIZE],
+                    public_share: bytes,
+                    host_input_share: bytes) -> State:
```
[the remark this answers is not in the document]
cjpatton: Ack, I think this would be overly verbose. I'd rather keep it implicit.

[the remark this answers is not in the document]
cjpatton: Host and Peer are common terms used in IETF specs (host == "us", peer == "them")

draft-irtf-cfrg-vdaf.md
```diff
@@ -1190,6 +1190,252 @@ how the nonces are chosen, but security requires that the nonces be unique. See
...
+  response consists of the prep message for the current round and the Helper's
+  prep share for the next round.
+
+The Aggregators proceed in this ping-ponging fashion until a step of the
+computation fails, indicating the report is invalid and should be rejected, or
+preparation is completed. All told there there are `ceil((Vdaf.ROUNDS+1)/2)`
+requests sent.
+
+Each message in the ping-pong protocol is structured as follows (expressed in
+TLS syntax {{!RFC8446}}):
```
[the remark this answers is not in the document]

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
tgeoghegan: I don't agree, but I suppose this isn't ambiguous or impossible to implement as written, so I'll defer to the VDAF editors.

tgeoghegan (approved):
There's nothing left I strictly object to. 🚢

cjpatton:
Nits from self-review.

PR #243, opened 2023-05-30
Redefine Prio3Histogram measurement and parameters

### LINKED ISSUE #239: Making Prio3Histogram's measurements more flexible
Currently, Prio3Histogram is parameterized by a list of bucket boundaries, and its measurement is a `Integer` that gets compared to all these boundaries, to determine which bucket's counter should be incremented. The same validity circuit has broader application than 1D histograms of integers that are directly supported. For instance, histograms with floating point numbers, multi-dimensional histograms, and multiple choice survey questions could all benefit from the same one-hot/mutual exclusion validity circuit, but using them with the VDAF as defined today would require defining awkward mappings. (i.e. creating a synthetic list of bucket boundaries, and providing measurements chosen to land in particular buckets) Similarly, someone may want a histogram with bucket intervals that are closed on the left instead of on the right.

I think we should either define a separate Prio3 VDAF, or redefine the Prio3Histogram VDAF, so that it takes a number of buckets as a parameter, and either a bucket index or a vector with a single 1 as a measurement. This would require most use cases to do some preprocessing before calling the shard function, but it wouldn't require turning floating point or vector inputs into an integer to be compared against synthetic bucket boundaries. Plus, replacing the array in the `buckets` parameter with a single cardinal number would make passing around VDAF parameters out of band easier. In ENPA, we handled metrics with over a thousand buckets, as a point of reference. (note that the aggregators have only ever cared about the number of buckets, and not the boundary values) Since this only affects the measurement and parameters, we could maintain wire and proof compatibility. I could prototype out the new interface in libprio-rs, maybe under a name like 'raw histogram'.

divergentdave:
This changes Prio3Histogram so that it is parameterized by a number of buckets, instead of a list of bucket boundaries. The measurement is now an index into the buckets, instead of an integer that's compared with the bucket boundaries. Closes #239. I confirmed that the messages in the Prio3Histogram test vector data are unchanged. (the parameter and measurement naturally changed)

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2657,25 +2657,21 @@ def Sum(inp: Vec[Field128], joint_rand: Vec[Field128]):
 
 ### Prio3Histogram
 
-This instance of Prio3 allows for estimating the distribution of the
-measurements by computing a simple histogram. Each measurement is an arbitrary
-integer and the aggregate result counts the number of measurements that fall in
-a set of fixed buckets.
+This instance of Prio3 allows for estimating a distribution of measurements by
+computing a simple histogram. Each measurement increments one histogram bucket,
+out of a set of fixed buckets. The aggregate result counts the number of
+measurements in each bucket.
 
 This instance of Prio3 uses PrgSha3 ({{prg-sha3}}) as its PRG. Its validity
-circuit, denoted `Histogram`, uses `Field128` ({{fields}}) as its finite
-field. The measurement is encoded as a one-hot vector representing the bucket
-into which the measurement falls (let `bucket` denote a sequence of
-monotonically increasing integers):
+circuit, denoted `Histogram`, uses `Field128` ({{fields}}) as its finite field.
+Let `length` be the number of histogram buckets. The measurement is encoded as a
+one-hot vector representing the bucket into which the measurement falls:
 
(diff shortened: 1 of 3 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -2657,25 +2657,21 @@ def Sum(inp: Vec[Field128], joint_rand: Vec[Field128]):
 
 ### Prio3Histogram
 
-This instance of Prio3 allows for estimating the distribution of the
-measurements by computing a simple histogram. Each measurement is an arbitrary
-integer and the aggregate result counts the number of measurements that fall in
-a set of fixed buckets.
+This instance of Prio3 allows for estimating a distribution of measurements by
+computing a simple histogram. Each measurement increments one histogram bucket,
+out of a set of fixed buckets. The aggregate result counts the number of
```
cjpatton, line 2662: I think we need to be a bit more clear about what the measurements are. It may also be helpful to be a bit more concrete with an example.

```suggestion
computing a simple histogram. Each measurement is the index of the histogram bucket,
out of a fixed-length sequence of buckets. For example, the buckets might quantize the real numbers, and each measurement would report the bucket into which a client's value (a real number) would fall. The aggregate result counts the number of
```

cjpatton (approved):

PR #250, opened 2023-06-09
Format Python code

divergentdave:
This formats all the code with autopep8 and isort, and then runs those in a new CI job.


cjpatton (approved):
Great idea :)

PR #251, opened 2023-06-14
Prepare to cut VDAF-06

cjpatton:


draft-irtf-cfrg-vdaf.md
```diff
@@ -328,6 +328,24 @@ security considerations for VDAFs.
...
 
+06:
+
+* Vdaf: Define a wrapper interface for preparation that is suitable for the
+  "ping-pong" topology in which two Aggregators exchange messages over a
+  request/response protocol, like HTTP, and take turns executing the
+  computation until input from the peer is required.
+
+* Prio3Histogram: Generalize the measurement type so that the histogram can be
+  used more easily with discrete domains. (\*)
```
cjpatton, line 339: @divergentdave We previously decided that this change is *not* wire breaking. Thinking about this more: a given measurement will be interpreted differently in 05 versus 06. For example, say the Client's measurement is 2:
* In 05, if the buckets are [0,10,100] then the Client's histogram contribution will be [0, 1, 0, 0]
* In 06, if the number of buckets is 4, then the Client's histogram contribution will be [0, 0, 1, 0]

So it's not wire breaking in the sense that the algorithm has changed (though `shard()` would now fail on inputs that it didn't fail before), but the aggregate result would be different. WDYT?

divergentdave: I think including the asterisk makes sense. It would be incorrect to provide the same measurement to both versions of the VDAF, since its semantics have changed. Whether it's wire-breaking, non-backwards compatible, etc. depends on where you draw the box around it ultimately, but we may as well be a little conservative here.

Issue #255, opened 2023-06-19
Define more common types


wangshan:
Some of the commonly used types are not explicitly defined using TLS syntax, or mentioned in "Conventions and Definitions". For example:  

- prep share
- prep msg
- input share
- public share

while similar ones like OutShare, Measurement is.

other types like `Prep` is treated like a type but it's content not specified when it's mentioned, but only at a later time, for e.g https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/draft-irtf-cfrg-vdaf.md#communication-patterns-for-preparation-vdaf-prep-comm.

cjpatton:
Typically you won't see cryptographic algorithms described in CFRG documents described in TLS syntax --- many predate RFC 8446 --- but I agree this would add clarity to the draft. As a bonus, it would allow us to remove some auxiliary functions that deal with serialization.

I think the only question is whether TLS syntax is expressive enough. I think we should investigate.

wangshan:
I agree, if TLS is not mandatory or fit for purpose, perhaps a more advanced general purpose language will make the description easier to understand.

wangshan:
Related to this, in the generic FLP section (https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/draft-irtf-cfrg-vdaf.md#validity-circuits-flp-generic-valid), many types should have been defined and used, but instead a generic Vec[...] is used:

`Valid.encode(measurement: Measurement) -> Vec[Field] returns a vector of length INPUT_LEN representing a measurement.`

cjpatton:
> ... many types should have been defined and used, but instead a generic Vec[...] is used

Which types do you think are missing?

cjpatton:
I've started digging this but I'm confused about what the actual ask is here. Currently:
1. If a value *must* be written to the wire in a protocol that uses VDAF, then our convention is to treat it as `Bytes`. This includes the prep share, prep message, input share, and public share.
2. Otherwise if a value is not necessarily written to the wire, then our convention is to define an explicit type for it so that we can use it. This includes the prep state, output share, and aggregate result.

It sounds to me that folks would favor replacing the `Bytes` with an explicit type for the first category, i.e., prep share, prep message, input share, and public share?

wangshan:
>  sounds to me that folks would favor replacing the Bytes with an explicit type for the first category, i.e., prep share, prep message, input share, and public share?

For me, yes.

[...]

Issue #222, opened 2023-04-14
Remove aliases of primitive Python types


divergentdave:
We would like to remove aliases of primitive types in `common.py`, and just use the primitive types straight out of their global bindings.

_Originally posted by @cjpatton in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/221#discussion_r1167331240_

cjpatton:
We should just change `Vec` to `list`, right?

PR #275, opened 2023-08-02
Require type definitions for encoded values

### LINKED ISSUE #58: Should the draft be prescriptive about encoding? (was "Serialization of aggregate shares needs to be specified")
`Vdaf` implementations need to define a type`Vdaf.AggShare` for the aggregate shares. Because the aggregate share are transmitted over the network, their encoding needs to be specified. There are two solutions to consider:
(1) Replace `Vdaf.AggShare` with `Bytes`
(2) Add methods on `Vdaf.AggShare` for encoding/decoding

The other quantities written to the network are input shares, prepare message shares, and the combined prepare message for each round. If we do (2), then for consistency we should also consider defining separate encoding/decoding methods for these. Currently the methods are implicit, since the type of each of these values is `Bytes`.

cjpatton:
Closes #58.
Partially addresses #255 (see TODOs)
Partially addresses #222

The current convention for the `Daf` and `Vdaf` interfaces is to any value that must be written to the network in order to carry out the computation as an opaque byte string (type `bytes`). This includes the public share, input share, prep share, prep message, and aggregate share. These values implicitly have type `bytes` and thus their types don't appear in the interface.

This mixture of explicit and implicit types makes the interface confusing. At the same time, the current adhoc approach to specifying encoding of types is hard to understand.

This commit requires (V)DAF to specify explicit types for all previously implicitly-`bytes` types. Furthermore, it RECOMMENDS that concrete (V)DAFs specify encodings for these. Finally, this change:

* Replaces the corresponding types for Prio3 and Poplar1 with more explicit types.
* Moves the previous encodings to their own sections and replaces the adhoc methods with TLS syntax.


draft-irtf-cfrg-vdaf.md
```diff
@@ -3144,39 +3100,42 @@ length-3 prefix of 25 (11001), but 7 (111) is not.
...
-* `Idpf.eval(agg_id: Unsigned, public_share: Bytes, key: Bytes, level:
-  Unsigned, prefixes: Tuple[Unsigned, ...], binder: Bytes) -> Idpf.Vec` is the
+A concrete IDPF defines the types and constants enumerated in {{idpf-param}}.
+In the remainder we write `Output` as shorthand for the type
+`Union[list[list[FieldInner]], list[list[FieldLeaf]]]`. (This type
+denotes either a vector of inner node field elements or leaf node field
+elements.) The scheme is comprised of the following algorithms:
+
+* `Idpf.gen(alpha: Unsigned, beta_inner: list[list[FieldInner]], beta_leaf:
+  list[FieldLeaf], binder: bytes, rand: bytes[Idpf.RAND_SIZE]) -> tuple[ bytes,
```
branlwyd, line 3150:
```suggestion
  list[FieldLeaf], binder: bytes, rand: bytes[Idpf.RAND_SIZE]) -> tuple[bytes,
```
(typo nit)

branlwyd (approved):
This is much easier to follow IMO, nice change!

draft-irtf-cfrg-vdaf.md
```diff
@@ -917,25 +916,29 @@ listed in {{vdaf-param}} are defined by each concrete VDAF.
...
+| `PublicShare`     | Type of each public share. |
+| `InputShare`      | Type of each input share. |
+| `AggParam`        | Type of aggregation parameter. |
+| `PrepState`       | State of each Aggregator during Preparation ({{sec-vdaf-prepare}}). |
+| `OutShare`        | Type of each output share. |
+| `AggShare`        | Type of the aggregate share. |
+| `AggResult`       | Type of the aggregate result. |
+| `PrepState`       | Aggregator's state during preparation. |
+| `PrepShare`       | Type of each prep share. |
+| `PrepMessage`     | Type of each prep messaeg. |
```
tgeoghegan, line 935:
```suggestion
| `PrepMessage`     | Type of each prep message. |
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1964,9 +1968,14 @@ methods refer to constants enumerated in {{prio3-const}}.
 | `SHARES`          | in `[2, 256)`                                   |
 | `Measurement`     | `Flp.Measurement`                               |
 | `AggParam`        | `None`                                          |
-| `PrepState`       | `tuple[Vec[Flp.Field], Optional[Bytes]]`        |
-| `OutShare`        | `Vec[Flp.Field]`                                |
+| `PublicShare`     | `Optional[list[bytes]]`                         |
+| `InputShare`      | `tuple[Union[tuple[list[Flp.Field],list[Flp.Field]],tuple[bytes, bytes]],Optional[bytes]]` |
```
tgeoghegan, line 1972: For readability, but I'm not sure if Markdown or the RFC renderer will like it:
```suggestion
| `InputShare`      | `tuple[`
|                   | `   Union[`
|                   | `       tuple[list[Flp.Field],list[Flp.Field]],`
|                   | `       tuple[bytes, bytes],`
|                   | `   ],`
|                   | `   Optional[bytes],`
|                   | `]`
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3230,9 +3226,14 @@ subsections. These methods make use of constants defined in {{poplar1-const}}.
 | `SHARES`          | `2`                                     |
 | `Measurement`     | `Unsigned`                              |
 | `AggParam`        | `Tuple[Unsigned, Tuple[Unsigned, ...]]` |
-| `PrepState`       | `Tuple[Bytes, Unsigned, Idpf.Vec]`      |
-| `OutShare`        | `Idpf.Vec`                              |
+| `PublicShare`     | `bytes` (IDPF public share)             |
+| `InputShare`      | `tuple[bytes, bytes, list[Idpf.FieldInner], list[Idpf.FieldLeaf]] ` |
```
tgeoghegan, line 3230:
```suggestion
| `InputShare`      | `tuple[bytes, bytes, list[Idpf.FieldInner], list[Idpf.FieldLeaf]]` |
```

tgeoghegan (commented):
I like this change because it makes the specification easier to read and also implement, because I'm less likely to mix up two values of type `PrepShare` and `PrepMessage` than two values both of type `bytes`.

I'm aware that this VDAF change exists in the context of a broader discussion about how CFRG documents should be written: as a protocol designer and implementer who consumes CFRG documents, I am very much in favor of using explicit types like this.

The remaining question is whether it's appropriate for a CFRG document to specify a wire encoding. While it's true that RFC 8446 struct syntax implies a wire encoding for messages, I think it does still allow other encodings. For instance a higher level protocol that builds on VDAF could choose to translate the RFC 8446 structs into JSON messages.

draft-irtf-cfrg-vdaf.md
cjpatton: RFC tools seem to hate this unforutnately.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1047,19 +1050,19 @@ class methods:
...
   preparation state and its message share for the current round or, if this is
   the last round, its output share. An exception is raised if a valid output
   share could not be recovered. The input of this algorithm is the inbound
   preparation message.
 
-* `Vdaf.prep_shares_to_prep(agg_param: AggParam, prep_shares: Vec[Bytes]) ->
-  Bytes` is the deterministic preparation-message pre-processing algorithm. It
-  combines the prep shares generated by the Aggregators in the previous round
-  into the prep message consumed by each in the next round.
+* `Vdaf.prep_shares_to_prep(agg_param: AggParam, prep_shares: lec[PrepShare])
```
tgeoghegan, line 1062:
```suggestion
* `Vdaf.prep_shares_to_prep(agg_param: AggParam, prep_shares: Vec[PrepShare])
```

tgeoghegan (commented):
I'm working on implementing the ping-pong topology in libprio-rs, which has caused me to closely inspect those routines and how they treat the values they work with. I think this PR needs to revisit the ping-pong routines. For example, consider the current:

~~~
def ping_pong_transition(
            Vdaf,
            agg_param: Vdaf.AggParam,
            prep_shares: Vec[bytes],
            prep_state: Vdaf.PrepState,
         ) -> (State, Message):
    prep_msg = Vdaf.prep_shares_to_prep(agg_param,
                                        prep_shares)
    out = Vdaf.prep_next(prep_state, prep_msg)
    if type(out) == Vdaf.OutShare:
        outbound = Message.finish(prep_msg)
        return (Finished(out), outbound)
    (prep_state, prep_share) = out
    outbound = Message.continue(prep_msg, prep_share)
    return (Continued(prep_state), outbound)
~~~

Note that the input is `prep_shares: Vec[bytes]`, which cannot be passed to `prep_shares_to_prep`. I think the function would have to decode each `prep_shares` into a `Vdaf.PrepShare` before calling `Vdaf.prep_shares_to_prep`. Since the ping pong functions are generic over `Vdaf`, then I believe you need to specify generic routines for encoding/decoding the various objects to/from bytes, which would correspond to the concrete encoding/decoding routines that are already in the Prio3 and Poplar1 "Auxiliary Functions" sections.

cjpatton:
Good catch, will fix before merging.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1234,6 +1238,35 @@ terminal states are `Rejected`, which indicates that the report cannot be
 processed any further, and `Finished(out_share)`, which indicates that the
 Aggregator has recovered an output share `out_share`.
```
cjpatton, line 1240: @tgeoghegan, @branlwyd PTAL at this new bit of normative language.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1234,6 +1238,35 @@ terminal states are `Rejected`, which indicates that the report cannot be
...
+
+* `Vdaf.encode_prep_msg(prep_msg: Vdaf.PrepMessage) -> bytes` encodes a prep
+  message.
+
+* `Vdaf.decode_prep_msg(prep_state: Vdaf.PrepState, encoded: bytes) ->
+  Vdaf.PrepMessage` decodes a prep message, using the prep state as optional
+  decoding context.
+
+Implementations of Prio3 and Poplar1 MUST use the encoding scheme specified in
+{{prio3-encode}} and {{poplar1-encode}} respectively.
```
cjpatton, line 1268: This MUST seems, unfortunately, necessary to me. One open question: how will the ping-pong API extend to other VDAFs that aren't in this document?

cjpatton (commented):
I've fixed up the ping-pong API, as @tgeoghegan suggested. I'd appreciate another look from everyone.

Does anyone still feel the spec is easier to understand this way? https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/275/files#r1286486411 is a bit unfortunate. I was supportive of this PR before, but I'm starting to think the spec is more complex now.

I would be in favor of going back to treating the public share, input share, prep share, and prep messages as byte strings, _but_ specifying them in TLS syntax rather than with adhoc methods.

junyechen1996 (commented):
I think it makes the spec easier to understand for me. Some comments and suggestions inline.

> Does anyone still feel the spec is easier to understand this way?

Yes. I think the message definition is clearer in each VDAF after this PR. Before this PR one would have to dig the auxiliary functions to find out the type definition. I see ping-pong APIs as a way to facilitate the network communication, by serializing and deserializing messages in and out of the VDAF. Correct me if I'm wrong.

> I would be in favor of going back to treating the public share, input share, prep share, and prep messages as byte strings, but specifying them in TLS syntax rather than with adhoc methods.

Will this require a VDAF to use TLS encoding for all its messages, i.e. one has to embed length prefix in the serialized message? I don't think VDAF messages should be forced to use TLS encoding. The current encoding/decoding interface for each type uses some "decoding context" (combination of the initialized VDAF and also `PrepState`), which should be extensible to other VDAFs. I think either way works, using the latter probably saves a couple of bytes in network transmission, but not sure if that's a strong argument yet...

draft-irtf-cfrg-vdaf.md
```diff
@@ -1234,6 +1238,35 @@ terminal states are `Rejected`, which indicates that the report cannot be
 processed any further, and `Finished(out_share)`, which indicates that the
 Aggregator has recovered an output share `out_share`.
 
+For convenience, the methods described in this section are defined in terms of
+opaque byte strings. A compatible `Vdaf` MUST specify methods for encoding
+public shares, input shares, prep shares, and prep messages. Minimally:
+
+* `Vdaf.decode_public_share(agg_id: Unsigned, encoded: bytes) ->
```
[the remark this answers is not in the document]
cjpatton: You're right, great catch!

draft-irtf-cfrg-vdaf.md
```diff
@@ -1047,19 +1050,19 @@ class methods:
   Protocols MUST ensure that public share consumed by each of the Aggregators is
   identical. This is security critical for VDAFs such as Poplar1.
 
-* `Vdaf.prep_next(prep_state: PrepState, prep_msg: Bytes) ->
-  Union[tuple[PrepState, Bytes], OutShare]` is the deterministic
+* `Vdaf.prep_next(prep_state: PrepState, prep_msg: PrepMessage) ->
+  Union[tuple[PrepState, PrepShare], OutShare]` is the deterministic
```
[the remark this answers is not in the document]
cjpatton: Interesting. Feel free to follow up with a PR sketching the alternative you have in mind.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1363,17 +1400,20 @@ the prep message, and computes the next prep state of the caller:
 def ping_pong_transition(
             Vdaf,
             agg_param: Vdaf.AggParam,
-            prep_shares: Vec[bytes],
+            prep_shares: list[Vdaf.PrepShare],
             prep_state: Vdaf.PrepState,
          ) -> (State, Optional[Message]):
```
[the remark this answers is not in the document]
cjpatton: Ack. This was intentionally left implicit, but `State` would be a superclass of `Continued` et al.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1396,11 +1436,17 @@ def ping_pong_continued(
         if inbound.type == 0: # initialize
             return (Rejected(), None)
 
-        out = Vdaf.prep_next(state.prep_state, inbound.prep_msg)
-        if type(out) == tuple[Vdaf.PrepState, Bytes] and inbound.type == 1:
+        prep_msg = state.prep_state(
+            Vdaf.decode_prep_msg(state.prep_state, inbound.prep_msg))
+        out = Vdaf.prep_next(state.prep_state, prep_msg)
+        if type(out) == tuple[Vdaf.PrepState, Vdaf.PrepShare] \
+                and inbound.type == 1:
```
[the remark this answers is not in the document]
cjpatton: We discussed this a bit in the initial version (https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240) and decided this was abusing TLS syntax too much.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
tgeoghegan: We talked about the lack of a definition of `State`. when ping pong was first added. I've just been working on implementing ping pong in libprio-rs, and I found the implicit definition of the `State` struct to be confusing. I would strongly recommend adding an explicit definition of this enumeration. You will have to do it anyway if you add a ping pong implementation to `poc`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2384,140 +2433,124 @@ def joint_rand(Prio3, k_joint_rand_seed):
...
 
 ~~~
-def encode_public_share(Prio3,
-                        k_joint_rand_parts):
-    encoded = Bytes()
-    if Prio3.Flp.JOINT_RAND_LEN > 0:
-        encoded += concat(k_joint_rand_parts)
-    return encoded
+struct {
+    Prio3.Flp.Field meas_share[Prio3.Flp.MEAS_LEN];
```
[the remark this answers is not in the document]
cjpatton: That's a good idea. One annoyance is that we'd have to name the field, so instead of 
`leader_share.meas_share` we'd refer to `leader_share.inner.meas_share`:

```
struct {
    Prio3LeaderShare inner;
    Prio3Seed k_blind;
} Prio3LeaderShareWithJointRand;
```

But I think the dedup is really nice so I took your suggestion.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
cjpatton: @tgeoghegan can you send a PR please?

cjpatton:
Sounds like folks still like this change, so we'll go ahead with it. Once it's approved I'll merge.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
tgeoghegan: Sure -- I will need to make some DAP changes to catch up with the current ping pong wrapper definitions anyway, and I think I might have a VDAF tweak or two to make.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2384,140 +2431,122 @@ def joint_rand(Prio3, k_joint_rand_seed):
...
+randomness part and is structured as follows:
 
-def encode_prep_msg(Prio3, k_joint_rand_check):
-    encoded = Bytes()
-    if Prio3.Flp.JOINT_RAND_LEN > 0:
-        encoded += k_joint_rand_check
-    return encoded
+~~~
+struct {
+    Prio3.Flp.Field verifier_share[Prio3.Flp.VERIFIER_LEN];
```
junyechen1996, line 2525: nit: here is another opportunity to reuse `Prio3PrepShare`, similar to `Prio3LeaderShare`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2340,6 +2380,13 @@ def helper_proof_share(Prio3, agg_id, k_share):
         Prio3.Flp.PROOF_LEN,
     )
 
+def expand_input_share(Prio3, agg_id, input_share):
+    (meas_share, proof_share, k_blind) = input_share
```
junyechen1996, line 2384: I think the intent is to map each field in the TLS struct to a variable, but it might be confusing to some people. Is it better to use field name to access each variable. For input share, it's more annoying that you have to condition based on leader or helper here. Will something like this be better?

```
if agg_id > 0:
    # helper, `input_share` is a `Prio3HelperShareWithJointRand`.
    meas_share, proof_share, k_blind = input_share.meas_share, input_share.proof_share, input_share.k_blind
    meas_share = Prio3.helper_meas_share(agg_id, meas_share)
    proof_share = Prio3.helper_proof_share(agg_id, proof_share)
else:
    # leader, `input_share` is a `Prio3LeaderShareWithJointRand`.
    meas_share, proof_share, k_blind = input_share.meas_share, input_share.proof_share, input_share.k_blind
return (meas_share, proof_share, k_blind)
```

junyechen1996 (approved):
Have one more suggestion for you to consider, but otherwise LGTM.

divergentdave (commented):
There are some issues with how this uses the TLS presentation language, most notably a classic case of vector lengths needing to count the number of bytes, not the number of nested messages. I think we also need to re-emphasize that field elements are little-endian, because TLS says multi-byte values are serialized in big-endian order. We don't thus far have a TLS-level declaration for a field element -- `opaque[ENCODED_LEN] Field;` may help.

More generally, I think that the TLS presentation language syntax is not adding much for our use case, at least beyond the ping-pong messages. We mainly use it to express concatenation, and we do very little nesting of structures. Much of the serialization formats lives in the prose gluing different structures for different cases together, especially when types are represented by the empty string.

As an alternative approach to achieve our clarity goals, we could replace the anonymous tuple types we use with named structures/classes in pseudocode/Python, and then give imperative encoding and decoding routines. This would allow us to show how the decoding context arguments are used, and would let us represent the different cases (particularly for Poplar1) more concisely and precisely.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3542,60 +3587,151 @@ def unshard(Poplar1, agg_param,
...
+`Poplar1.Idpf.FieldLeaf`. PRG seeds are represented as follows:
 
-Implementation note: The aggregation parameter includes the level of the IDPF
-tree and the sequence of indices to evaluate. For implementations that perform
-per-report caching across executions of the VDAF, this may be more information
-than is strictly needed. In particular, it may be sufficient to convey which
-indices from the previous execution will have their children included in the
-next. This would help reduce communication overhead.
+~~~
+uint8 Poplar1Seed[S];
```
divergentdave, line 3607: Likewise
```suggestion
opaque Poplar1Seed[S];
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3542,60 +3587,151 @@ def unshard(Poplar1, agg_param,
...
-        raise ERR_INPUT
-    return (key, corr_seed, corr_inner, corr_leaf)
+uint8 Poplar1FieldInner[Fi];
+~~~
+
+Likewise, elements of the leaf field are encoded in little-endian byte order
+(as defined in {{field}}) and are represented as follows:
+
+~~~
+uint8 Poplar1FieldLeaf[Fl];
```
divergentdave, line 3621:
```suggestion
opaque Poplar1FieldInner[Fi];
~~~

Likewise, elements of the leaf field are encoded in little-endian byte order
(as defined in {{field}}) and are represented as follows:

~~~
opaque Poplar1FieldLeaf[Fl];
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3542,60 +3587,151 @@ def unshard(Poplar1, agg_param,
...
+The public share is equal to the IDPF public share, which is a byte string.
+(See {{idpf}}.)
+
+#### Input Share
+
+Each input share is structured as follows:
+
+~~~
+struct {
+    uint8 idpf_key[Poplar1.Idpf.KEY_SIZE];
```
divergentdave, line 3635:
```suggestion
    opaque idpf_key[Poplar1.Idpf.KEY_SIZE];
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3542,60 +3587,151 @@ def unshard(Poplar1, agg_param,
...
+~~~
+struct {
+    Poplar1FieldLeaf sketch[Fl * 3];
+} Poplar1PrepMessageRoundOneLeaf;
+
+Note that these messages are identical to the prep shares for the same round.
+
+The second-round prep message is the empty string. This is because the sketch
+shares are expected to sum to a particular value if the output shares are
+valid; we represent this value by the empty string.
```
divergentdave, line 3703: These paragraphs seem a little confusing, here's a suggested clarification.
```suggestion
Note that these messages have the same structures as the prep shares for the
first round.

The second-round prep message is the empty string. This is because the sketch
shares are expected to sum to a particular value if the output shares are
valid; we represent a successful preparation with the empty string and
otherwise return an error.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3542,60 +3587,151 @@ def unshard(Poplar1, agg_param,
...
+
+#### Aggregate Share
 
+The encoding of the aggregate share depends on whether the inner or leaf field
+is used. It also depends on the number of candidate prefixes, which is
+determined by the aggregation parameter.
```
divergentdave, line 3709: It's probably worth restating that the choice of field is determined by the aggregation parameter here.
```suggestion
The encoding of the aggregate share depends on whether the inner or leaf field
is used, and the number of candidate prefixes. Both of these are determined by
 the aggregation parameter.
```

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -1396,11 +1434,16 @@ def ping_pong_continued(
         if inbound.type == 0: # initialize
             return (Rejected(), None)
 
-        out = Vdaf.prep_next(state.prep_state, inbound.prep_msg)
-        if type(out) == tuple[Vdaf.PrepState, Bytes] and inbound.type == 1:
+        prep_msg = Vdaf.decode_prep_msg(state.prep_state, inbound.prep_msg)
+        out = Vdaf.prep_next(state.prep_state, prep_msg)
+        if type(out) == tuple[Vdaf.PrepState, Vdaf.PrepShare] \
```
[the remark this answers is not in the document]
cjpatton: Ack. @tgeoghegan this may be worth addressing when you try to rework things so that there no invalid states.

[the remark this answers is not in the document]
cjpatton: Filed https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/283

draft-irtf-cfrg-vdaf.md
```diff
@@ -2384,140 +2432,133 @@ def joint_rand(Prio3, k_joint_rand_seed):
...
-                        meas_share,
-                        proof_share,
-                        k_blind):
-    encoded = Bytes()
-    encoded += Prio3.Flp.Field.encode_vec(meas_share)
-    encoded += Prio3.Flp.Field.encode_vec(proof_share)
-    if Prio3.Flp.JOINT_RAND_LEN > 0:
-        encoded += k_blind
-    return encoded
+uint8 Prio3Seed[S];
```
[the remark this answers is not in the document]
cjpatton: Done, here and elsewhere.

cjpatton:
Squashed.

Issue #226, opened 2023-04-24
Review: 05: Chris Wood: Prose description of the FLP lacks precision


cjpatton:
#flp-generic-construction-prove (and the other sections for the generic FLP) uses prose to describe steps of the generic FLP like proof generation, querying, and so on. This is distinctly less precise than the rest of the document, which uses pseudocode to describe steps. For example, consider the step to actually compute gadget polynomials:

 >      Let `poly_gadget_i = G_i(poly_wire_i[0], ..., poly_wire_i[L_i-1])`. That
 >    is, evaluate the circuit `G_i` on the wire polynomials for the `i`th
 >    gadget. (Arithmetic is in the ring of polynomials over `Field`.)

I would not expect most people to be familiar with the concept of a ring, nor would most people know how to intuitively implement circuit evaluation. Perhaps it might be good to give examples of common gadgets -- like Mul -- and how to implement this evaluation?

Moreover, consider the following:

>   Let `poly_wire_i[j-1]` be the lowest degree polynomial for which
>     `poly_wire_i[j-1](alpha_i^k) == padded_w[k]` for all `k` in `[P_i]`.

How does one actually implement this? In general, prose that is not followed by an algorithm seems like it could be replaced with more precise implementation guidance. The reference implementation could be used as the basis for this content, if desired.

PR #276, opened 2023-08-02
Add Prio3SumVec

### LINKED ISSUE #124: Sum of vectors with Prio3
I think it would be useful to have a VDAF instantiation to sum vectors of integers. This has come up a few times in conversations as a useful primitive. We could define it with two parameters: the length of the vector in elements, and the bit width of each element, so that their range would be $[0, 2^{bits}-1)$. I presume that we'd be able to make similar improvements to proving time or proof size here as with the proposed [Prio3Aes128CountVec](https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/83), compared to running multiple independent `Prio3Aes128Sum` VDAF instances.

divergentdave:
This introduces the ParallelSum gadget and uses it in Prio3SumVec, which computes sums of vectors of integers.

This differs from the `Prio3SumVec` implementation that's been in libprio-rs for a while in three respects. First, the chunk length is now an explicit parameter, rather than being floor(sqrt(length * bits)). This both lets us dodge compatibility concerns about how exactly to compute that square root, and provides flexibility to squeeze a few more bytes out by using more optimal chunk length choices. (There are two different kinds of rounding involved in the proof size formula, plus there's a factor of two in front of the rounded number of calls) Second, libprio-rs defined a `BlindPolyEval(x, r) = r * p(x)` gadget for use in this validity circuit, but this PR just uses `Mul(x, y) = x * y` and extra affine gates on the inputs. This lets us keep the gadget's degree at two even when using the parallel sum gadget. We're still calculating the same random linear combination of $\sum_i {r^{i+1}(x_i^2-x_i)}$. Third, to pad out the inputs to the last ParallelSum invocation, I decided to substitute zero instead of substituting the measurement's last element. I figured that using zeros here is simpler.

Closes #124.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3072,6 +3072,127 @@ measurement is sharded. This is provided to the FLP by Prio3.
 | `Field`          | `Field128` ({{fields}}) |
 {: title="Parameters of validity circuit Histogram."}
 
+### Prio3SumVec
+
+This instance of Prio3 supports summing a vector of integers. It has three
+parameters, `length`, `bits`, and `chunk_length`. Each measurement is a vector
+of positive integers with length equal to the `length` parameter. Each element
+of the measurement is an integer in the range `[0, 2^bits)`. It is RECOMMENDED
+to set `chunk_length` to an integer near to the square root of `length * bits`.
+The optimal choice for any measurement size will vary slightly due to rounding.
+
+This instance uses PrgSha3 ({{prg-sha3}}) as its PRG. Its validity circuit,
+denoted `SumVec`, uses `Field128` ({{fields}}) as its finite field.
+
+Measurements are encoded as a vector of field elements with length
+`length * bits`. The field elements in the encoded vector represent all the bits
+of the measurement vector's elements, consecutively, in LSB to MSB order:
+
+~~~
+def encode(self, measurement: Vec[Unsigned]):
+    if len(measurement) != self.length:
+        raise ERR_INPUT
+
+    encoded = []
+    for val in measurement:
+        if 0 > val or val >= 2 ** self.bits:
+            raise ERR_INPUT
+
+        for l in range(self.bits):
+            encoded.append(self.Field((val >> l) & 1))
+
+    return encoded
+
+def truncate(self, meas):
+    truncated = [self.Field(0) for _ in range(self.length)]
+    for i in range(self.length):
+        for j in range(self.bits):
+            weight = self.Field(1 << j)
+            truncated[i] += weight * meas[i * self.bits + j]
+    return truncated
+
+def decode(self, output, _num_measurements):
+    return [x.as_unsigned() for x in output]
+~~~
+
+This validity circuit uses a `ParallelSum` gadget to achieve a smaller proof
+size. This optimization for "parallel-sum circuits" is described in
+{{BBCGGI19}}, section 4.4. Briefly, for circuits that add up the output of
+multiple identical subcircuits, it is possible to achieve smaller proof sizes
+(on the order of O(sqrt(n)) instead of O(n)) by packaging more than one such
+subcircuit into a gadget.
+
+The `ParallelSum` gadget is parameterized with an arithmetic subcircuit, and a
+`count` of how many times it evaluates that subcircuit. It takes in a list of
+inputs and passes them through to instances of the subcircuit in the same order.
+It returns the sum of the subcircuit outputs. Note that only the `ParallelSum`
+gadget itself, and not its subcircuit, participates in `FlpGeneric`'s wire
+recording during evaluation, gadget consistency proofs, and proof validation,
+even though the subcircuit is provided to `ParallelSum` as an implementation
+of the `Gadget` interface.
+
+~~~
+def eval(self, Field, inp):
+    self.check_gadget_eval(inp)
+    out = Field(0)
+    for i in range(self.count):
+        start_index = i * self.subcircuit.ARITY
+        end_index = (i + 1) * self.subcircuit.ARITY
+        out += self.subcircuit.eval(Field, inp[start_index:end_index])
+    return out
+~~~
+
+The validity circuit checks that the encoded measurement consists of ones and
+zeros. Rather than use the `Range2` gadget on each element, as in the `Sum`
+validity circuit, it instead uses a `Mul` subcircuit and "free" constant
+multiplication and addition gates to simultaneously evaluate the same range
+check polynomial on each element, and multiply by a constant. One of the two
+`Mul` subcircuit inputs is equal to a measurement element multiplied by a power
+of the joint randomness value, and the other is equal to the same measurement
+element minus one. These `Mul` subcircuits are evaluated by a `ParallelSum`
+gadget, and the results are added up both within the `ParallelSum` gadget and
+after it.
+
+~~~
+def eval(self, meas, joint_rand, num_shares):
+    self.check_valid_eval(meas, joint_rand)
+
+    out = Field128(0)
+    r = joint_rand[0]
+    r_power = r
+    shares_inv = self.Field(num_shares).inv()
+
+    for i in range(self.GADGET_CALLS[0]):
+        inputs = [None] * (2 * self.chunk_length)
+        for j in range(self.chunk_length):
+            index = i * self.chunk_length + j
+            if index < len(meas):
+                meas_elem = meas[index]
+            else:
+                meas_elem = self.Field(0)
+
+            inputs[j * 2] = r_power * meas_elem
+            inputs[j * 2 + 1] = meas_elem - shares_inv
+
+            r_power *= r
+
+        out += self.GADGETS[0].eval(self.Field, inputs)
+
+    return out
+~~~
+
+| Parameter        | Value                                                  |
+|:-----------------|:-------------------------------------------------------|
+| `GADGETS`        | `[ParallelSum(Mul(), chunk_length)]`                   |
+| `GADGET_CALLS`   | `[(length * bits + chunk_length - 1) // chunk_length]` |
+| `MEAS_LEN`       | `length * bits`                                        |
+| `OUTPUT_LEN`     | `length`                                               |
+| `JOINT_RAND_LEN` | `1`                                                    |
+| `Measurement`    | `Vec[Unsigned]`, each element in range `[0, 2^bits)`   |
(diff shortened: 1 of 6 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
 | `Field`          | `Field128` ({{fields}}) |
 {: title="Parameters of validity circuit Histogram."}
 
+### Prio3SumVec
+
+This instance of Prio3 supports summing a vector of integers. It has three
+parameters, `length`, `bits`, and `chunk_length`. Each measurement is a vector
```
cjpatton, line 3068:
```suggestion
parameters, `length`, `bits`, and `chunk_length`. Each measurement is a vector of positive integers
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+denoted `SumVec`, uses `Field128` ({{fields}}) as its finite field.
+
+Measurements are encoded as a vector of field elements with length
+`length * bits`. The field elements in the encoded vector represent all the bits
+of the measurement vector's elements, consecutively, in LSB to MSB order:
+
+~~~
+def encode(SumVec, measurement: Vec[Unsigned]):
+    if len(measurement) != SumVec.length:
+        raise ERR_INPUT
```
cjpatton, line 3084: Orthogonal to this PR, but: WDYTA about using normal exceptions for these kinds of things?
```suggestion
        raise Exception('input: invalid length')
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+
+    encoded = []
+    for elem in measurement:
```
cjpatton, line 3087: `elem` is reminiscent of "field element". 
```suggestion
    for val in measurement:
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+        if 0 > elem or elem >= 2 ** SumVec.bits:
+            raise ERR_INPUT
+
+        for l in range(SumVec.bits):
+            encoded.append(Sum.Field((elem >> l) & 1))
```
cjpatton, line 3092:
```suggestion
            encoded.append(SumVec.Field((elem >> l) & 1))
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+(on the order of O(sqrt(n)) instead of O(n)) by packaging more than one such
+subcircuit into a gadget.
+
+The `ParallelSum` gadget is parameterized with an arithmetic subcircuit, and a
+`count` of how many times it evaluates that subcircuit. It takes in a list of
+inputs and passes them through to instances of the subcircuit in the same order.
+It returns the sum of the subcircuit outputs.
+
+~~~
+def ParallelSum(self, inputs):
```
cjpatton, line 3121: * We should assert any necessary preconditions, i.e., `len(inputs)` should be equal to `self.cont * self.subcircuit.ARITY`, correct?
* Would `subgadget` be a more precise term to use? Since it has the an `ARITY` field, we're expecting it to implement the `Gadget` interface.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+    out = Field(0)
+    for i in range(self.count):
+        start_index = i * self.subcircuit.ARITY
+        end_index = (i + 1) * self.subcircuit.ARITY
+        out += self.subcircuit(inputs[start_index:end_index])
+    return out
+~~~
+
+The validity circuit checks that the encoded measurement consists of ones and
+zeros. Rather than use the `Range2` gadget on each element, it instead uses a
```
cjpatton, line 3131: "Rather than use the `Range2` gadget..." It sounds like you mean to draw a comparison to `Sum`? If so, I'd make this more explicit.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+`Mul` subcircuit and "free" constant multiplication and addition gates to
+simultaneously evaluate the same range check polynomial on each element, and
+multiply by a constant. (Arranging the circuit this way allows us to use the
+`ParallelSum` gadget, because the summands are identical subcircuits. Using
+`Range2` and then multiplying by different constants would not fit the parallel
+sum pattern.) One of the two `Mul` subcircuit inputs is equal to a measurement
```
cjpatton, line 3137: This parenthetical seems only useful if you're familiar with the (non-standardized) VDAF in libprio-rs with the same name (which we plan to deprecate when we cut the next draft). I'd suggest adding a bit more context here, perhaps describing in greater detail how the "naive" approach to parallelizing `Prio3Sum`would work (i.e., what we have in libprio-rs today). Alternatively, we might just cut this.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3062,6 +3062,129 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+                meas_elem = meas[index]
+            else:
+                meas_elem = self.Field(0)
+
+            inputs[j * 2] = r_power * meas_elem
+            inputs[j * 2 + 1] = meas_elem - shares_inv
+
+            r_power *= r
+
+        out += ParallelSum(Mul(), self.chunk_length, inputs)
```
cjpatton, line 3171: * How is `ParallelSum.count` computed from `self.chunk_length`? It looks to me like what we mean to is `len(inputs) / self.chunk_length` (`+1` if there is a partial chunk)
* Hmm, we have a class method called `ParrallelSum` on an undefined class and a constructor for a class called `ParallelSum`. This seems to suggest an odd edge case in how we've been specifying circuits up to this point. This suggests to me that we may want to refactor the API a bit. What do you think about renaming the former to a class method called `call()`?

cjpatton (commented):
I've only looked at the draft so far, I'll do a pass of the reference code once the draft is a bit more settled. Looking good so far!

draft-irtf-cfrg-vdaf.md
cjpatton: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/277

draft-irtf-cfrg-vdaf.md
divergentdave: The actual implementation includes a call to `check_gadget_eval()`, which compares the input length to the gadget's arity, which is exactly what we need. This is moreso a symptom of us hiding implementation details in our descriptions of gadgets and validity circuits in the document.

I wanted to avoid calling it a gadget, because it doesn't participate in recording of input wires during evaluation and proof/verification of gadget polynomial consistency. Perhaps we should define a separate `Circuit` interface that both `Valid` and `Gadget` inherit from, that just covers `ARITY`, `DEGREE`, `eval()`, and `eval_poly()`?

draft-irtf-cfrg-vdaf.md
divergentdave: The chunk_length VDAF parameter is passed directly through to be `ParallelSum.count`. The gadget's `count` parameter is the number of times it calls the circuit, see the description above. We do that division by chunk_length to decide how many times to call the ParallelSum gadget instead.

Yeah, all the pseudocode for gadgets or validity circuits thus far doesn't match the Python implementation, as they're written as if they are very short functions. The shorthand is breaking down with ParallelSum, since it takes parameters upon its construction, and takes a variable number of input wires. I think making the distinction between the class and the `call()` method under the covers would clear this up.

draft-irtf-cfrg-vdaf.md
divergentdave: My intent was to compare to the `Prio3Sum` circuit, and explain why it isn't in a parallel sum form. I think I'll cut the parenthetical.

draft-irtf-cfrg-vdaf.md
cjpatton: Agreed on the first point 👍 

To the second point: That's a good point about not overloading the term "gadget". Perhaps all that's needed is a comment clarifying that `subcircuit.ARITY` refers to the arity of the gadget that we're parallelizing. I'm wary of making a bigger change this, as it may end up creating more work when we address https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/226.

draft-irtf-cfrg-vdaf.md
cjpatton: Ohhh gotcha, I misunderstood the intent. I think what I would do is have separate paragraph somewhere that is dedicated to contrasting Prio3Sum with Prio3SumVec, perhaps even with its own heading. Totally optional, but I think we're likely to get this question when the CFRG reviews this draft.

draft-irtf-cfrg-vdaf.md
cjpatton: Do you want to take a stab at adding `call()` this PR, or do you think we should do it in a separate PR and rebase? Your call -- I think it's a pretty minor change and impcts the reference code more than the draft,

divergentdave: Yeah, a separate PR sounds like a good idea, I keep finding things I want to clean up. (like fixing shadowing class names with the first parameter to class methods, and moving check_valid_eval etc. to be superclass methods instead of free functions)

divergentdave: Rebased over #280

poc/vdaf.py
```diff
@@ -139,13 +139,12 @@ def domain_separation_tag(Vdaf, usage: Unsigned) -> Bytes:
         return format_dst(0, Vdaf.ID, usage)
 
     @classmethod
-    def test_vec_set_type_param(Vdaf, test_vec):
+    def test_vec_set_type_param(Vdaf, test_vec) -> Vec[str]:
```
cjpatton, line 142:
```suggestion
    def test_vec_set_type_param(Vdaf, test_vec) -> list[str]:
```

cjpatton (approved):

PR #279, opened 2023-08-03
Add security considerations for fixed-key AES

### LINKED ISSUE #216: Review: 05: Chris Wood: Security implications of using PrgFixedKeyAes128 for IDPF
A correlation-robust hash function is a new primitive for the CFRG. What do we need from it for IDPF, and why does it achieve this?

> Have the PRG constructions seen security analysis? Some of them seem straightforward, but I think it would be helpful to sketch proofs in the security considerations. We did this in hash-to-curve for the benefit of future readers who might ask. (I ask this because I saw "The multi-instance tweakable circular correlation-robust hash function of" and thought to myself, "gee, this seems new!")

schoppmp:
Closes #216.


schoppmp:
I didn't add a lot of detail on *why* fixed-key AES implements a correlation-robust hash function, since the proofs in GKWY20 and GKWWY20 are quite technical. @wangxiao1254 do you have a good high-level intuition for this? Also, for DPFs we don't really need circular correlation robustness, right? Does that mean we can avoid the sigma orthomorphism? That would make things simpler, but it probably won't matter for performance, and the only variant that has a formal proof in GKWWY (in the multi-instance setting) is the circular variant. But if there is any other advantage of having CCR here then please let me know.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4035,16 +4046,31 @@ differential privacy.
...
+want to be conservative in our assumptions, and hence prescribe PrgSha3 as the
+only RECOMMENDED Prg instantiation.
 
 The one exception is the PRG used in the Idpf implementation IdpfPoplar
-{{idpf-poplar}}. Here, a random oracle is not needed to prove security, and
-hence a construction based on fixed-key AES {{prg-fixed-key-aes128}} can be
-used. However, as PrgFixedKeyAes128 has been shown to be differentiable from
-a random oracle {{GKWWY20}}, it is NOT RECOMMENDED to use it anywhere else.
+{{idpf-poplar}}. Here, a random oracle is not needed to prove security, since
+the security analysis of {{BBCGGI21}}, Proposition 1, only requires a PRG.
```
cjpatton, line 4056: What about extractability? We used the ROM to prove our IDPF variant extractable in [[DPRS23](https://eprint.iacr.org/2023/130), Lemma 1]. Could we have maybe used correlation robustness instead?

schoppmp: Hmm yeah, extractability may in fact be an issue. BBCGGI21 also assume a random oracle for the extractable DPF, although they implement it using fixed-key AES. Now the question is how bad is it if we don't have extractability. One-hotness of the *evaluated* part of the client's vector is ensured through the sketch in BBCGGI21 (or through the technique from [DP22](https://eprint.iacr.org/2021/580.pdf) in the case of Doplar). So the main concern here is the attack described in [[BBCGGI21](https://eprint.iacr.org/2021/017.pdf), Section 4.2]: A client using the fact that fixed-key AES is *not* a random oracle in order to craft a malicious DPF key pair that evaluates to 1 in more than one position, but where the servers only end up evaluating one of these positions. Personally I would feel fine with documenting this, and offering PrgSha3 as an alternative for users who want the additional robustness guarantees that extractability gives them. For privacy both fixed-key AES and SHA-3 should be fine (under the respective assumptions).

draft-irtf-cfrg-vdaf.md
```diff
@@ -4035,16 +4046,31 @@ differential privacy.
...
+As observed in {{GKWY20}}, a PRG can be instantiated from a correlation-robust
+hash function `H`. Informally, correlation robustness requires that for a random
+`r`, `H(xor(r, x))` is indistinguishable from a random function for any `x`.
```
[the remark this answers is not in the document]
schoppmp: Correlation Robustness is just a property of the hash function, that you can prove in a certain model. The fixed-key AES construction is proven to be correlation-robust in the random permutation model [[GKWY20]](https://eprint.iacr.org/2019/074.pdf) for a single fixed key, and the in the ideal cipher model for the multi-instance variant that we use [[GKWWY20]](https://eprint.iacr.org/2019/1168.pdf).

wangxiao1254:
I will comment on this soon with some intuition on Phillipp's first comment, but note that random-permutation model, where we prove correlation robustness is also an idealized model. orthomorphism can be avoided if circularity is not needed. The main advantage between CCR and RO is that CCR can be constructed using RPM, which is commonly instantiated using fixed-key AES (of course, this final step is not provable, just like how we instantiate SHA for RO).

I want to quickly note that using fixed-key AES to instantiate random oracle is unconventional. fixed-key AES is invertible while random oracle is not.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
cjpatton: In that case I'd add a paranthetical here that says our particular construction is proven secure in the ideal cipher model.

draft-irtf-cfrg-vdaf.md
cjpatton: That sounds reasonable, but:
1. We'll need to define a separate codepoint for Poplar1 with SHA-3
2. We'll need to provide concrete guidance (i.e., if attack X is possible, then users MUST choose Poplar1 with SHA-3)

Question: Is there anything about the attack you describe that's not captured by robustness as defined in DPRS23? Do you think that extractability would be needed to prove that Poplar1 is robust?

For what it's worth, I'm not sure the analysis of Doplar will be entirely analogous. We use the fact that we can extract from the inputs from the random oracle in order to then do a reduction to DFLP. This initial simplifying step necessary because if we don't know if the IDPF output is onehot, then a DFLP for the sum of the output vector is vacuous. (When we were working on the proof, we couldn't figure out a different assumption about IDPF that would have admitted the same argument.)

cjpatton:
> I will comment on this soon with some intuition on Phillipp's first comment, but note that random-permutation model, where we prove correlation robustness is also an idealized model. orthomorphism can be avoided if circularity is not needed. The main advantage between CCR and RO is that CCR can be constructed using RPM, which is commonly instantiated using fixed-key AES (of course, this final step is not provable, just like how we instantiate SHA for RO).

Thanks Xiao, we appreciate your feedback. Overall the CFRG is happy to work in idealized models of computation when necessary, but we just need to make clear in the draft what that model is.


> I want to quickly note that using fixed-key AES to instantiate random oracle is unconventional. fixed-key AES is invertible while random oracle is not.

Yup, this is at least clear to the authors of this draft, but we now need to make sure this is clear to anyone who reads this.

draft-irtf-cfrg-vdaf.md
schoppmp: I added a paragraph talking about extractability and what it means for robustness. Given that the (theoretical) attacks by malicious clients are very limited, I don't think it's necessary to define Poplar1 with SHA-3, and I doubt anyone would actually use it in practice.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
schoppmp: That part is already there in the next paragraph that talks about our concrete instantiation of the correlation-robust hash function.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4515,19 +4526,36 @@ differential privacy.
...
+for most situations.
+
+The one exception is the Idpf implementation IdpfPoplar {{idpf-poplar}}.
+Here, a random oracle is not needed to prove privacy, since the analysis of
+{{BBCGGI21}}, Proposition 1, only requires a Pseudorandom Generator (PRG).
+As observed in {{GKWY20}}, a PRG can be instantiated from a correlation-robust
+hash function `H`. Informally, correlation robustness requires that for a random
+`r`, `H(xor(r, x))` is indistinguishable from a random function for any `x`.
+A PRG can therefore be constructed as
+`PRG(r) = H(xor(r, 1)) || H(xor(r, 2)) || ...`,
```
cjpatton, line 4542: Do you intend for this to be on its own line? If so, then:
```suggestion

~~~
PRG(r) = H(xor(r, 1)) || H(xor(r, 2)) || ...`
~~~

```
If you want this math to be in-lined, then, make sure lines break at 80 characters.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4515,19 +4526,36 @@ differential privacy.
...
+mostly modelled as random oracles. XofShake128 is designed to be
+indifferentiable from a random oracle {{MRH04}}, making it a suitable choice
+for most situations.
+
+The one exception is the Idpf implementation IdpfPoplar {{idpf-poplar}}.
+Here, a random oracle is not needed to prove privacy, since the analysis of
+{{BBCGGI21}}, Proposition 1, only requires a Pseudorandom Generator (PRG).
+As observed in {{GKWY20}}, a PRG can be instantiated from a correlation-robust
+hash function `H`. Informally, correlation robustness requires that for a random
+`r`, `H(xor(r, x))` is indistinguishable from a random function for any `x`.
```
cjpatton, line 4540:
```suggestion
`r`, `H(xor(r, x))` is computationally indistinguishable from a random function of `x`.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4515,19 +4526,36 @@ differential privacy.
...
+function.
+
+Our construction at {{xof-fixed-key-aes128}} implements a correlation-robust
+hash function using fixed-key AES. For security, it assumes that AES with a
+fixed key can be modeled as a random permutation {{GKWY20}}. Additionally, we
+use a different AES key for every client, which in the ideal cipher model leads
+to better concrete security {{GKWWY20}}.
+
+We note that for robustness, the analysis of {{DPRS23}} still assumes a random
+oracle to make the Idpf *extractable*. While XofFixedKeyAes128 has been shown
```
cjpatton, line 4553:
```suggestion
oracle to make the Idpf extractable. While XofFixedKeyAes128 has been shown
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4515,19 +4526,36 @@ differential privacy.
...
+since each individual hash function evaluation is indistinguishable from a random
+function.
+
+Our construction at {{xof-fixed-key-aes128}} implements a correlation-robust
+hash function using fixed-key AES. For security, it assumes that AES with a
+fixed key can be modeled as a random permutation {{GKWY20}}. Additionally, we
+use a different AES key for every client, which in the ideal cipher model leads
+to better concrete security {{GKWWY20}}.
+
+We note that for robustness, the analysis of {{DPRS23}} still assumes a random
```
cjpatton, line 4552: I don't think our analysis applies here, since Doplar requires a stronger IDPF (extractable VIDPF). Our claim is that Poplar1 is robust as-is, due to the arithmetic sketching MPC. Doplar on the other hand would require an extractable VIDPF.

Question: Do {{BBCGGI21}} require a random oracle in order to get their notion of extracability for IdpfPoplar? If so, then it's worth noting that IdpfPoplar is not extractable.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4515,19 +4526,36 @@ differential privacy.
...
+oracle to make the Idpf *extractable*. While XofFixedKeyAes128 has been shown
+to be differentiable from a random oracle {{GKWWY20}}, there are no known
+attacks exploiting this difference.
+We also stress that beven if the Idpf is not extractable, the Poplar1 guarantees
```
cjpatton, line 4556:
```suggestion
We also stress that been if the Idpf is not extractable, the Poplar1 guarantees
```

cjpatton (changes requested):

cjpatton:
Please rebasse and remove the merge commit from the history. It's useful to have a linear commit history when we're getting ready to cut a draft :)

draft-irtf-cfrg-vdaf.md
schoppmp: Yes, {{BBCGGI21}} also requires a RO. I'll replace the reference in the text.

schoppmp:
> Please rebasse and remove the merge commit from the history. It's useful to have a linear commit history when we're getting ready to cut a draft :)

Of course, I always squash the entire PR into a single commit when merging.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4515,19 +4526,40 @@ differential privacy.
 
 ## Requirements for XOFs {#xof-vs-ro}
 
-XofShake128 is designed to be indifferentiable from a random oracle {{MRH04}},
-making it a suitable choice for most situations, in particular wherever the XOF
-is modeled as a random oracle in existing security analysis.
+As described in {{xof}}, our constructions rely on an interface for eXtendable
+Output Functions (XOFs). In the security analyses of our protocols, these are
```
cjpatton, line 4530: Could we just say this?
```suggestion
As described in {{xof}}, our constructions rely on eXtendable
Output Functions (XOFs). In the security analyses of our protocols, these are
```

cjpatton (approved):

PR #280, opened 2023-08-03
Clean up Gadget, Valid, and Flp APIs, update extracts in document

divergentdave:
This makes a variety of changes to the FlpGeneric implementation to simplify things and better follow Python conventions, and then replaces the existing gadget/circuit pseudocode in the document with extracts of the corresponding `eval()` methods. The biggest change is that all the Flp and Valid APIs that were in class methods were moved to plain methods, and the subclass factories were replaced with plain constructors. I renamed several arguments to either avoid upper-case names or shadowing of bindings from outer scopes. (I left `Field` as-is for now, since we also import a module named `field`) I also defined a `Range2` gadget, effectively a specialization of `PolyEval`, so that we can use it in the validity circuits that already claim to use it, and I can extract its `eval()` method into the document.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1022,7 +1022,7 @@ measurement. At the end of the computation, each Aggregator holds an output
 share or an error."}
 
 To facilitate the preparation process, a concrete VDAF implements the following
-class methods:
+methods:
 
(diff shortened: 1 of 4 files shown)


cjpatton (approved):
This is much better! Changes to the reference code look correct. One non-blocking comment about `call_poly()`.

cjpatton:
By the way, I think it will do us a favor later on if we squash this stack into a single commit before merging.

PR #281, opened 2023-08-03
Ping-pong topology fixes

tgeoghegan:
A handful of fixes to the specification of the ping-pong aggregator
communication topology encountered during implementation.

 - Explicitly describe Python objects for ping-pong State
 - `ping_pong_leader_init` and `_helper_init` return `Message`, not `bytes`
 - `ping_pong_transition` always returns a `Message`
 - typo in `ping_pong_transition`
 - `ping_pong_continued` fails if `state` is not `Continued`
 - remove unused `is_leader` parameter from `ping_pong_transition`

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1237,9 +1237,33 @@ terminal states are `Rejected`, which indicates that the report cannot be
 processed any further, and `Finished(out_share)`, which indicates that the
 Aggregator has recovered an output share `out_share`.
 
+~~~
+class State:
+    pass
+
+class Start(State):
+    pass
+
+class Continued(State):
+    def __init__(self, prep_state):
+        self.prep_state = prep_state
+
+class Finished(State):
+    def __init__(self, output_share):
+        self.output_share = output_share
+
+class Rejected(State):
+    def __init__(self):
+        pass
+~~~
+
+Note that there is no representation of the `Start` state as it is never
+instantiated in the ping-pong topology.
+
 For convenience, the methods described in this section are defined in terms of
 opaque byte strings. A compatible `Vdaf` MUST specify methods for encoding
-public shares, input shares, prep shares, and prep messages. Minimally:
+public shares, input shares, prep shares, prep messages, and aggregation
+parameters. Minimally:
 
 * `Vdaf.decode_public_share(encoded: bytes) -> Vdaf.PublicShare` decodes a
   public share.
@@ -1261,6 +1285,12 @@ public shares, input shares, prep shares, and prep messages. Minimally:
 * `Vdaf.decode_prep_msg(prep_state: Vdaf.PrepState, encoded: bytes) ->
   Vdaf.PrepMessage` decodes a prep message, using the prep state as optional
   decoding context.
+
+* `Vdaf.decode_agg_param(encoded: bytes) -> Vdaf.AggParam` decodes an
+  aggregation parameter.
+
+* `Vdaf.encode_agg_param(agg_param: Vdaf.AggParam) -> bytes` encodes an
+  aggregation parameter.
 
 Implementations of Prio3 and Poplar1 MUST use the encoding scheme specified in
 {{prio3-encode}} and {{poplar1-encode}} respectively.
@@ -1322,13 +1352,24 @@ struct {
 } Message;
 ~~~
 
+These messages are used to transition between the states described in
+{{vdaf-prep-comm}}. They are encoded and decoded to or from byte buffers as
+described {{Section 3 of !RFC8446}}) using the following routines:
+
+* `encode_ping_pong_message(message: Message) -> bytes` encodes a `Message` into
+  an opaque byte buffer.
+
+* `decode_pong_pong_message(encoded: bytes) -> Message` decodes an opaque byte
+  buffer into a `Message`, raising an error if the bytes are not a valid
+  encoding.
+
 The Leader's initial transition is computed with the following procedure:
 
 ~~~ transition
 def ping_pong_leader_init(
             Vdaf,
             vdaf_verify_key: bytes[Vdaf.VERIFY_KEY_SIZE],
-            agg_param: Vdaf.AggParam,
+            agg_param: bytes,
             nonce: bytes[Vdaf.NONCE_SIZE],
             public_share: bytes,
             input_share: bytes,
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -1365,7 +1365,7 @@ def ping_pong_transition(
             agg_param: Vdaf.AggParam,
             prep_shares: Vec[bytes],
             prep_state: Vdaf.PrepState,
-         ) -> (State, Optional[Message]):
```
tgeoghegan, line 1368: `ping_pong_continued`, below, returns `(State, Optional[Message])` and does `return Vdaf.ping_pong_transition(...)`. Do we have to do anything there to translate the `Message` value into something like `Some(message)`? Or is returning non-`None` good enough?

tgeoghegan:
Downgrading to draft because I may find other stuff as I implement this and will tack on commits to this PR.

tgeoghegan:
This is ready for review now and corresponds to what I have implemented in https://github.com/divviup/libprio-rs/pull/683. There are further changes I'd like to suggest to ping-pong, but more immediately, what I think we should do is review and merge this change, and then make a corresponding change to DAP that adapts to these new definitions. Then we can release DAP-06 and VDAF-07 and have published draft versions we can refer to in implementations. Further refinements to ping-pong can wait for subsequent drafts of the pair of documents.

tgeoghegan:
I think there's a problem with how aggregation parameters are handled in ping-pong. We have: 

```
def ping_pong_leader_init(
            Vdaf,
            vdaf_verify_key: bytes[Vdaf.VERIFY_KEY_SIZE],
            agg_param: Vdaf.AggParam,
            nonce: bytes[Vdaf.NONCE_SIZE],
            public_share: bytes,
            input_share: bytes,
        ) -> tuple[State, Message]
```

But at the DAP layer, we always represent aggregation parameters as opaque byte buffers. For example:

```
struct {
  opaque agg_param<0..2^32-1>;
  PartialBatchSelector part_batch_selector;
  PrepareInit prepare_inits<1..2^32-1>;
} AggregationJobInitReq;
```

I added a commit here that resolves this by aligning ping-pong's handling of agg param with its handling of things like the public share or input shares: the arguments are now `bytes` and VDAFs are required to define encode/decode routines for their agg param. However we could also rewrite these interfaces to use the expressive types, and then make it DAP's responsibility to call encode/decode. I think VDAF has to require encode/decode routines no matter what, though.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1262,6 +1262,12 @@ public shares, input shares, prep shares, and prep messages. Minimally:
   Vdaf.PrepMessage` decodes a prep message, using the prep state as optional
   decoding context.
```
cjpatton, line 1264: Add `aggregation parameter` to the list above on line 1242.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1322,22 +1328,45 @@ struct {
 } Message;
 ~~~
 
+These messages are used to transition between the states described in
+{{vdaf-prep-comm}, more specifically structured as follows:
```
cjpatton, line 1332:
```suggestion
{{vdaf-prep-comm}}, which are defined as:
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1322,22 +1328,45 @@ struct {
...
+    def __init__(self, output_share):
+        self.output_share = output_share
+
+class Rejected(State):
+    def __init__(self):
+        pass
+~~~
+
+Note that there is no representation of the `Start` state as it is never
+instantiated in the ping-pong topology.
```
cjpatton, line 1352: We could just define it as
```python3
class Start(State)
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1322,22 +1328,45 @@ struct {
 } Message;
 ~~~
 
+These messages are used to transition between the states described in
+{{vdaf-prep-comm}, more specifically structured as follows:
+
+~~~
```
cjpatton, line 1334: Can we move these to the supersection, just after defining the state machine?

cjpatton (commented):
Looks good to me! Just some editorial things.

tgeoghegan:
Corresponding DAP changes: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/494

draft-irtf-cfrg-vdaf.md
cjpatton: Whoops, I meant
```python3
class Start(State):
    pass
```

tgeoghegan: I don't want to suggest that implementations should ever do anything about the `Start` state, though. At least with ping-pong, there's no reason for any implementation to ever instantiate a `Start` value or deal with one.

cjpatton (approved):

cjpatton (approved):
LGTM. Let's way to merge until we've checked the corresponding changes in the VDAF spec. I also want to look at libprio's implementation.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1262,6 +1283,12 @@ public shares, input shares, prep shares, and prep messages. Minimally:
   Vdaf.PrepMessage` decodes a prep message, using the prep state as optional
   decoding context.
 
+* `Vdaf.decode_agg_param(encoded: bytes) -> Vdaf.AggParam` decodes an
+  aggregation parameter.
+
+* `Vdaf.encode_agg_param(agg_param: Vdaf.AggParam) -> bytes` encodes an
```
divergentdave, line 1289: We should adjust the definition of `encode_agg_param` for Poplar1 in 8.2.6.6 to match this, so it takes in the aggregation parameter as a single tuple argument (rather than two separate arguments).

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -1237,9 +1237,30 @@ terminal states are `Rejected`, which indicates that the report cannot be
...
+    def __init__(self, output_share):
+        self.output_share = output_share
+
+class Rejected(State):
+    def __init__(self):
+        pass
+~~~
+
+Note that there is no representation of the `Start` state as it is never
+instantiated in the ping-pong topology.
```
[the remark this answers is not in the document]
tgeoghegan: You and Chris both think it makes more sense, and you're the editors, so I'll add it.

cjpatton:
@tgeoghegan can you squash and summarize the changes in the commit message?

Issue #274, opened 2023-08-01
Apply parallel sum gadget optimization to Prio3Histogram


divergentdave:
The range check portion of the `Prio3Histogram` circuit could be rewritten to use the parallel sum gadget. This would result in more wire polynomials, each of smaller degree, smaller proofs ($O(\sqrt{n})$ instead of $O(n)$<span/>), faster proving, and larger preparation message shares ($O(\sqrt{n})$ instead of $O(1)$<span/>). See https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/124#issuecomment-1656741283.

PR #284, opened 2023-08-16
Apply parallel sum optimization to Histogram FLP

divergentdave:
This implements #274, hoisting the description of the ParallelSum gadget up a section, and changing the Histogram circuit to use it as well. This reduces report sizes by a bit under two thirds asymptotically. Sharding performance is improved for large histograms as well (I got 27% reduction in runtime for a length of 1000)

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3046,88 +3046,21 @@ def eval(self, meas, joint_rand, _num_shares):
 | `Field`          | `Field128` ({{fields}})  |
 {: title="Parameters of validity circuit Sum."}
 
-### Prio3Histogram
-
-This instance of Prio3 allows for estimating the distribution of some quantity
-by computing a simple histogram. Each measurement increments one histogram
-bucket, out of a set of fixed buckets. (Bucket indexing begins at `0`.) For
-example, the buckets might quantize the real numbers, and each measurement
-would report the bucket that the corresponding client's real-numbered value
-falls into. The aggregate result counts the number of measurements in each
-bucket.
-
-This instance of Prio3 uses PrgSha3 ({{prg-sha3}}) as its PRG. Its validity
-circuit, denoted `Histogram`, uses `Field128` ({{fields}}) as its finite field.
-Let `length` be the number of histogram buckets. The measurement is encoded as a
-one-hot vector representing the bucket into which the measurement falls:
-
-~~~
-def encode(self, measurement):
-    encoded = [self.Field(0)] * self.length
-    encoded[measurement] = self.Field(1)
-    return encoded
-
-def truncate(self, meas):
-    return meas
-
-def decode(self, output, _num_measurements):
-    return [bucket_count.as_unsigned() for bucket_count in output]
-~~~
-
-The `Histogram` validity circuit uses `Range2` (see {{prio3sum}}) as its single
-gadget. It checks for one-hotness in two steps, as follows:
-
-~~~
-def eval(self, meas, joint_rand, num_shares):
-    self.check_valid_eval(meas, joint_rand)
-
-    # Check that each bucket is one or zero.
-    range_check = self.Field(0)
-    r = joint_rand[0]
-    for b in meas:
-        range_check += r * self.GADGETS[0].eval(self.Field, [b])
-        r *= joint_rand[0]
-
-    # Check that the buckets sum to 1.
-    sum_check = -self.Field(num_shares).inv()
-    for b in meas:
-        sum_check += b
-
-    out = joint_rand[1] * range_check + \
-        joint_rand[1] ** 2 * sum_check
-    return out
-~~~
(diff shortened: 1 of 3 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -3074,8 +3080,47 @@ def decode(self, output, _num_measurements):
     return [bucket_count.as_unsigned() for bucket_count in output]
 ~~~
 
-The `Histogram` validity circuit uses `Range2` (see {{prio3sum}}) as its single
-gadget. It checks for one-hotness in two steps, as follows:
+This validity circuit uses a `ParallelSum` gadget to achieve a smaller proof
+size. This optimization for "parallel-sum circuits" is described in
+{{BBCGGI19}}, section 4.4. Briefly, for circuits that add up the output of
+multiple identical subcircuits, it is possible to achieve smaller proof sizes
+(on the order of O(sqrt(n)) instead of O(n)) by packaging more than one such
```
cjpatton, line 3087: nit: Since `n` is undefined :)
```suggestion
(on the order of O(sqrt(length)) instead of O(length)) by packaging more than one such
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3074,8 +3080,47 @@ def decode(self, output, _num_measurements):
...
+subcircuit into a gadget.
+
+The `ParallelSum` gadget is parameterized with an arithmetic subcircuit, and a
+`count` of how many times it evaluates that subcircuit. It takes in a list of
+inputs and passes them through to instances of the subcircuit in the same order.
+It returns the sum of the subcircuit outputs. Note that only the `ParallelSum`
+gadget itself, and not its subcircuit, participates in `FlpGeneric`'s wire
+recording during evaluation, gadget consistency proofs, and proof validation,
+even though the subcircuit is provided to `ParallelSum` as an implementation
+of the `Gadget` interface.
```
cjpatton, line 3097: Thanks for moving this explanation here. However now that we have Prio3SumVec, I wonder if it should go right after Prio3Sum, which would mean moving this explanation back to Prio3SumVec 😮‍💨

draft-irtf-cfrg-vdaf.md
```diff
@@ -3074,8 +3080,47 @@ def decode(self, output, _num_measurements):
...
+~~~
+
+The `Histogram` validity checks for one-hotness in two steps, by checking that
+the encoded measurement consists of ones and zeros, and by checking that the sum
+of all elements in the encoded measurement is equal to one. All the individual
+checks are combined together in a random linear combination.
+
+The range checks on each element of the encoded measurement are performed by the
+`ParallelSum` gadget, which evaulates `Mul` subcircuits. Rather than use the
+`Range2` gadget on each element, as in the `Sum` validity circuit, it instead
```
cjpatton, line 3117: This comparison would be a tad cleaner if we started with Prio3SumVec

cjpatton (commented):
Looks good, but I propose, for editorial reasons, we present Prio3SumVec first such that it immediately follows Prio3Sum. This ought to make the transitional text a little cleaner, as well being a little more logical of an order:
Prio3Count
Prio3Sum
Prio3SumVec
Prio3Hisotgram

I'd also propose that we swap the code points for Prio3SumVec and Prio3Histogram for consistency with the ordering. We're already going to need to bump the `VERSION` so this shouldn't be an interop issue.

divergentdave:
I'm wary about swapping the codepoints. If we had the IANA registries mentioned in section 10 set up, we wouldn't be able to do that. Plus, they are already baked into [draft-wang-ppm-dap-taskprov](https://wangshan.github.io/draft-wang-ppm-dap-taskprov/draft-wang-ppm-dap-taskprov.html#section-3-9).

cjpatton:
> I'm wary about swapping the codepoints. If we had the IANA registries mentioned in section 10 set up, we wouldn't be able to do that. Plus, they are already baked into [draft-wang-ppm-dap-taskprov](https://wangshan.github.io/draft-wang-ppm-dap-taskprov/draft-wang-ppm-dap-taskprov.html#section-3-9).

The IANA considerations are irrelevant until we actually ask IANA for codepoints. Similarly for taskprov: until this is a WG draft, technically we don't really care. Unless you need to support multiple versions of that draft, I'd suggest we change the codepoints here and make sure the latest version is in line.

Of course, this is mostly a question of aesthetics. If you'd rather not complicate taskprov, then I'm fine leaving things as is.

draft-irtf-cfrg-vdaf.md
divergentdave: I changed this to `MEAS_LEN` instead, now that this is moving back to Prio3SumVec.

cjpatton (approved):
Looks good to me. If you're concerned about changing the codepoints, feel free to leave them where they are, and I'll file issue to follow up if I decide I care enough :)

Issue #209, opened 2023-03-22
Reconsider PRG choice


simon-friedberger:
If we reconsider our choice of PRG we should take into account that:

- The RNG is modeled as a random oracle in the proofs but this does not necessarily mean that we need a XOF. It might be more efficient to create a seed using something like SHA-3 (or -2) and then applying a fast AES based stream derivation.
- cSHAKE (which we are currently using) was only defined in NIST SP 800-185 not the original SHA-3 standard (FIPS 202) so it might not be available in some implementations. FIPS 202 has SHAKE which is also a XOF and we are already defining how to encode the binding string. We could simply define something similar to cSHAKE based on SHAKE. (Exact match is not possible due to non-matching constants.)

cjpatton:
Thanks for kicking off this discussion, Simon! Here are my personal views on these points at the moment.

>     * The RNG is modeled as a random oracle in the proofs but this does not necessarily mean that we need a XOF. It might be more efficient to create a seed using something like SHA-3 (or -2) and then applying a fast AES based stream derivation.

In my opinion, more security analysis is required before we consider this change. I think it would be sufficient to prove that `PrgFixedKeyAes128` is [indifferentiable](https://blog.cryptographyengineering.com/2012/07/17/indifferentiability/) from a random oracle when SHA-3 (or SHA-2) is modeled as a random oracle and fixed-key AES is modeled as an ideal permutation (or something else?). Currently this is is only known to be correlation-robust (see Section 7.4 of [this paper](https://eprint.iacr.org/2019/074.pdf)). Alternatively, we could revisit the security proofs for Prio3 and see if we can push through reduction with a a function that is only correlation-robust.

Regarding performance: `PrgFixedKeyAes128` is much faster than `PrgSha3`. (Incidentally, @divergentdave it would be great to have libprio benchmarks for this, if we don't already!) However, whether we need this speed up depends on how we use the PRG: for `Prio3` the sharding/prep times are very similar with both, since the runtime is dominated by finite finite arithmetic; but for things that use the IDPF, like `Poplar1`, the performance difference is quite significant, since the runtime is dominated by the PRG. (See preliminary benchmarks in #106 for details.) In fact, this is exactly why we wanted `FixedKeyAes128`. (See #32.)

[...]

PR #285, opened 2023-08-18
prg: unrolled cSHAKE padding

hannahdaviscrypto:
#209 Switched cSHAKE128 to SHAKE128 in the draft and reference implementation as discussed. The changes here are intended to be minimal, meaning that we pad the domain separation string in the same way that cSHAKE128 pads its customization string, so the only part of the Keccak input that changes is the SHAKE version-identifying postfix. I think a lot of this padding is unnecessary and we can omit it if we like.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1661,12 +1661,14 @@ def expand_into_vec(Prg,
 ### PrgSha3 {#prg-sha3}
 
 This section describes PrgSha3, a PRG based on the Keccak permutation of SHA-3
-{{FIPS202}}. Keccak is used in the cSHAKE128 mode of operation {{SP800-185}}.
+{{FIPS202}}. Keccak is used in the SHAKE128 mode of operation {{FIPS202}}.
 This Prg is RECOMMENDED for all use cases within VDAFs.
+The length of the domain separation string `dst` passed to PrgSha3 MUST NOT
+exceed 255 bytes.
 
 ~~~
 class PrgSha3(Prg):
-    """PRG based on SHA-3 (cSHAKE128)."""
+    """PRG based on SHA-3 (SHAKE128)."""
 
     # Associated parameters
     SEED_SIZE = 16
@@ -1679,14 +1681,15 @@ class PrgSha3(Prg):
     def next(self, length: Unsigned) -> Bytes:
         self.l += length
 
-        # Function `cSHAKE128(x, l, n, s)` is as defined in
-        # [SP800-185, Section 3.3].
+        # Function `SHAKE128(x, l)` is as defined in
+        # [FIPS 202, Section 6.2].
         #
         # Implementation note: Rather than re-generate the output
         # stream each time `next()` is invoked, most implementations
         # of SHA-3 will expose an "absorb-then-squeeze" API that
         # allows stateful handling of the stream.
-        stream = cSHAKE128(self.x, self.l, b'', self.s)
+        dst_length = to_le_bytes(len(self.s), 1)
+        stream = SHAKE128(dst_length + self.s + self.x, self.l)
         return stream[-length:]
 ~~~
 {: title="Definition of PRG PrgSha3."}
@@ -1698,6 +1701,8 @@ is needed in the VDAFs described in this document, there are some cases where
 a more efficient instantiation based on fixed-key AES is possible. For now, this
 is limited to the Prg used inside the Idpf {{idpf}} implementation in Poplar1
 {{idpf-poplar}}. It is NOT RECOMMENDED to use this Prg anywhere else.
+The length of the domain separation string `dst` passed to PrgFixedKeyAes128
+MUST NOT exceed 255 bytes.
 See Security Considerations {{security}} for a more detailed discussion.
 
 ~~~
@@ -1720,7 +1725,8 @@ class PrgFixedKeyAes128(Prg):
         #
         # Implementation note: This step can be cached across PRG
         # evaluations with many different seeds.
-        self.fixed_key = cSHAKE128(binder, 16, b'', dst)
+        dst_length = to_le_bytes(len(dst), 1)
+        self.fixed_key = SHAKE128(dst_length + dst + binder, 16)
         self.seed = seed
 
     def next(self, length: Unsigned) -> Bytes:
--- a/poc/prg.py
+++ b/poc/prg.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from Cryptodome.Cipher import AES
-from Cryptodome.Hash import CMAC, cSHAKE128
+from Cryptodome.Hash import CMAC, SHAKE128
 from Cryptodome.Util import Counter
 
 from common import (TEST_VECTOR, VERSION, Bytes, Unsigned, concat, format_dst,
@@ -95,7 +95,7 @@ class PrgAes128(Prg):
 
 
 class PrgSha3(Prg):
-    """PRG based on SHA-3 (cSHAKE128)."""
+    """PRG based on SHA-3 (SHAKE128)."""
 
     # Associated parameters
     SEED_SIZE = 16
@@ -104,9 +104,12 @@ class PrgSha3(Prg):
     test_vec_name = 'PrgSha3'
 
     def __init__(self, seed, dst, binder):
-        # `dst` is used as the customization string; `seed || binder` is
-        # used as the main input string.
-        self.shake = cSHAKE128.new(custom=dst)
+        # The input is composed of `dst`, the domain separation tag, the
+        # `seed`, and the `binder` string.
+        self.shake = SHAKE128.new()
+        dst_length = to_le_bytes(len(dst), 1)
+        self.shake.update(dst_length)
+        self.shake.update(dst)
         self.shake.update(seed)
         self.shake.update(binder)
 
@@ -136,7 +139,10 @@ class PrgFixedKeyAes128(Prg):
         #
         # Implementation note: This step can be cached across PRG
         # evaluations with many different seeds.
-        shake = cSHAKE128.new(custom=dst)
+        shake = SHAKE128.new()
+        dst_length = to_le_bytes(len(dst), 1)
+        shake.update(dst_length)
+        shake.update(dst)
         shake.update(binder)
         fixed_key = shake.read(16)
         self.cipher = AES.new(fixed_key, AES.MODE_ECB)
@@ -216,7 +222,8 @@ if __name__ == '__main__':
         b'',  # binder
         5,
     )
-    assert expanded_vec[-1] == Field64(13681157193520586550)
+    # TODO: Update the test to account for the change from cSHAKE128 to SHAKE128.
+    # assert expanded_vec[-1] == Field64(13681157193520586550)
 
     for cls in (PrgAes128, PrgSha3, PrgFixedKeyAes128):
         test_prg(cls, Field128, 23)


poc/prg.py
```diff
@@ -106,13 +106,24 @@ class PrgSha3(Prg):
     def __init__(self, seed, dst, binder):
         # `dst` is used as the customization string; `seed || binder` is
         # used as the main input string.
-        self.shake = cSHAKE128.new(custom=dst)
+        self.shake = SHAKE128.new()
+        dst_length_l = (len(dst).bit_length() + 7)//8
+        dst_length = len(dst).to_bytes(dst_length_l, byteorder='little')
```
cjpatton, line 111: Since `len(dst) < 256` we can simplify this a bit (`to_le_bytes` is in the `common` module)
```suggestion
        dst_length = to_le_bytes(len(dst), 1)
```

poc/prg.py
```diff
@@ -106,13 +106,24 @@ class PrgSha3(Prg):
...
+        dst_length_l = self.__to_bytes__(dst_length_l)  # one byte value
```
cjpatton, line 112: Similarly here (`byte` is in the `common` module)

```suggestion
        dst_length_l = byte(1)
```

poc/prg.py
```diff
@@ -106,13 +106,24 @@ class PrgSha3(Prg):
...
+        pre_pad = self.__to_bytes__(1) + self.__to_bytes__(168)
```
cjpatton, line 113:
```suggestion
        pre_pad = byte(1) + byte(168)
```

poc/prg.py
```diff
@@ -136,7 +147,15 @@ def __init__(self, seed, dst, binder):
...
         # evaluations with many different seeds.
-        shake = cSHAKE128.new(custom=dst)
+        shake = SHAKE128.new()
+        dst_length_l = (len(dst).bit_length() + 7)//8
+        dst_length = len(dst).to_bytes(dst_length_l, byteorder='little')
+        dst_length_l = self.__to_bytes__(dst_length_l)  # one byte value
+        pre_pad = self.__to_bytes__(1) + self.__to_bytes__(168)
+        post_pad_l = (168 - ((len(dst) + len(dst_length)+3) % 168)) % 168
+        post_pad = (0).to_bytes(post_pad_l, byteorder='little')
+        dom_sep = pre_pad + dst_length_l + dst_length + dst + post_pad
```
cjpatton, line 157: Can we deduplicate this logic, i.e., create a wrapper function for it and call it here and above?

cjpatton (commented):
> Switched cSHAKE128 to SHAKE128 in the draft and reference implementation as discussed. The changes here are intended to be minimal, meaning that we pad the domain separation string in the same way that cSHAKE128 pads its customization string, so the only part of the Keccak input that changes is the SHAKE version-identifying postfix.

Just unrolling the cSHAKE domain separation bits is a reasonable first step, but if this is all we want to do, then I think it would make sense to keep the versioning bits so that we don't change the test vectors.

Note that one of the tests is failing in `prg` at the moment because this is a wire-breaking change for `PrgSha3`. I think we can comment it out for now and re-generate it once later:
```
sage -python common.py
sage -python field.py
sage -python prg.py
Traceback (most recent call last):
  File "/Users/christopherpatton/github.com/cfrg/draft-irtf-cfrg-vdaf/poc/prg.py", line 241, in <module>
    assert expanded_vec[-1] == Field64(13681157193520586550)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
make: *** [test] Error 1
```

>  I think a lot of this padding is unnecessary and we can omit it if we like.

Yes, I think we should take this time to simplify the padding. Indeed, this is why I asked _you_ in particular to bring your crypto eyes to this issue :)

cjpatton:
@hannahdaviscrypto can you take a look at https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#section-5.3.2 and decide it might work for us here?

cjpatton:
> @hannahdaviscrypto can you take a look at https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#section-5.3.2 and decide it might work for us here?

I took a closer look and quickly realized this doesn't work for us because it expects to know the length of the output in advance. We don't in our application.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1661,32 +1661,38 @@ def expand_into_vec(Prg,
...
+        # [FIPS 202, Section 6.2].
         #
         # Implementation note: Rather than re-generate the output
         # stream each time `next()` is invoked, most implementations
         # of SHA-3 will expose an "absorb-then-squeeze" API that
         # allows stateful handling of the stream.
-        stream = cSHAKE128(self.x, self.l, b'', self.s)
+        # The length of dst, the domain separation string, should be <256 bytes
+        # A longer dst will cause an OverflowError
+        dst_length = len(dst).to_bytes(1, byteorder = 'little') #one byte value
```
cjpatton, line 1694: Here and below
```suggestion
        dst_length = to_le_bytes(len(dst), 1)
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1661,32 +1661,38 @@ def expand_into_vec(Prg,
 ### PrgSha3 {#prg-sha3}
 
 This section describes PrgSha3, a PRG based on the Keccak permutation of SHA-3
-{{FIPS202}}. Keccak is used in the cSHAKE128 mode of operation {{SP800-185}}.
+{{FIPS202}}. Keccak is used in the SHAKE128 mode of operation {{FIPS202}}.
 This Prg is RECOMMENDED for all use cases within VDAFs.
```
cjpatton, line 1666:
```suggestion
The length of the domain separation string `dst` passed to PrgSha3 MUST NOT
exceed 255 bytes.

```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1661,32 +1661,38 @@ def expand_into_vec(Prg,
...
-    """PRG based on SHA-3 (cSHAKE128)."""
+    """PRG based on SHA-3 (SHAKE128)."""
 
     # Associated parameters
     SEED_SIZE = 16
 
     def __init__(self, seed, dst, binder):
+        # Implementation note: it is required that the length of dst be
+        # at most 255 bytes. This comes from the use of cSHAKE128's padding
+        # scheme (see {{SP800-185}} for details).
```
cjpatton, line 1677: This text needs to be normative, so I think we should move it out of a code comment and into its own paragraph. See the suggestion above.

Similarly for PrgFixedKeyAes128 below.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1720,7 +1726,10 @@ class PrgFixedKeyAes128(Prg):
         #
         # Implementation note: This step can be cached across PRG
         # evaluations with many different seeds.
-        self.fixed_key = cSHAKE128(binder, 16, b'', dst)
+        # The length of dst, the domain separation string, should be <256 bytes
+        # A longer dst will cause an OverflowError
+        dst_length = len(dst).to_bytes(1, byteorder = 'little') #one byte value
+        self.fixed_key = SHAKE128(dst_length + dst + binder, 16)
```
cjpatton, line 1732: (For consideration in a follow-up PR.) @schoppmp We might go with plain SHA-3 in order instead of SHAKE128 since the output length is fixed.

cjpatton (commented):
This approach looks good to me, I have just a few editorial comments.

poc/prg.py
```diff
@@ -104,9 +104,12 @@ class PrgSha3(Prg):
     test_vec_name = 'PrgSha3'
 
     def __init__(self, seed, dst, binder):
-        # `dst` is used as the customization string; `seed || binder` is
+        # `dst` is used for domain separation; `seed || binder` is
         # used as the main input string.
-        self.shake = cSHAKE128.new(custom=dst)
+        self.shake = SHAKE128.new()
+        dst_length = to_le_bytes(len(dst), 1)  # one byte value
```
cjpatton, line 110: nit: This is clear enough from context :)
```suggestion
        dst_length = to_le_bytes(len(dst), 1)
```

poc/prg.py
```diff
@@ -216,7 +222,7 @@ def test_prg(Prg, F, expanded_len):
         b'',  # binder
         5,
     )
-    assert expanded_vec[-1] == Field64(13681157193520586550)
+    # assert expanded_vec[-1] == Field64(13681157193520586550)
```
cjpatton, line 225: When breaking a test, it's important to make clear to the reader why the test is broken (and how it should be fixed).
Bonus: We should file an issue to remind us to fix this test, and reference the issue in this comment.
```suggestion
    # TODO: Update the test to account for the change from cSHAKE128 to SHAKE128.
    # assert expanded_vec[-1] == Field64(13681157193520586550)
```

cjpatton (commented):
Looks good to me. Please add a bit of documentation regarding the broken test (see inline comment).

@divergentdave could you take a quick look at this change?

draft-irtf-cfrg-vdaf.md
```diff
@@ -1679,14 +1681,15 @@ class PrgSha3(Prg):
...
+        # Function `SHAKE128(x, l)` is as defined in
+        # [FIPS 202, Section 6.2].
         #
         # Implementation note: Rather than re-generate the output
         # stream each time `next()` is invoked, most implementations
         # of SHA-3 will expose an "absorb-then-squeeze" API that
         # allows stateful handling of the stream.
-        stream = cSHAKE128(self.x, self.l, b'', self.s)
+        dst_length = to_le_bytes(len(dst), 1)
+        stream = SHAKE128(dst_length + dst + self.x, self.l)
```
divergentdave, line 1692: nit: `dst` is passed through `self` here
```suggestion
        dst_length = to_le_bytes(len(self.s), 1)
        stream = SHAKE128(dst_length + self.s + self.x, self.l)
```

poc/prg.py
```diff
@@ -104,9 +104,12 @@ class PrgSha3(Prg):
     test_vec_name = 'PrgSha3'
 
     def __init__(self, seed, dst, binder):
-        # `dst` is used as the customization string; `seed || binder` is
+        # `dst` is used for domain separation; `seed || binder` is
         # used as the main input string.
```
divergentdave, line 108: Since there's no separate customization string, we don't have a distinction to draw here anymore.
```suggestion
        # The input is composed of `dst`, the domain separation tag, the
        # `seed`, and the `binder` string.
```

divergentdave (approved):
> Just unrolling the cSHAKE domain separation bits is a reasonable first step, but if this is all we want to do, then I think it would make sense to keep the versioning bits so that we don't change the test vectors.

cSHAKE128 output can only be matched by SHAKE128 if the customization string is empty. We always provided a customization string, and in that case, we land in the other branch of cSHAKE128's definition, that uses Keccak[256] directly. In that branch, the bits `00` are appended to Keccak's input after the cSHAKE128 input, whereas the definition of SHAKE128 appends the bits `1111` after its input. We are making this change in the pursuit of ease of implementation, so I think we ought to stick with SHAKE128 as-is, and accept the value-breaking change. Bare Keccak[256] implementations are likely to be as rare as cSHAKE128 implementations.

Regarding removal of the padding bytes after the customization string, section 7.1 of SP 800-185 says they were included to allow precomputation of sponge state for any particular choice of function name and customization string. Our DSTs are 8 bytes, so we're only foreclosing on one precomputation of the Keccak permutation function, and on the other hand doing one fewer absorption iteration about half the time. This sounds fine to me, as it's not a big deal.

poc/prg.py
cjpatton: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/288

cjpatton (approved):

cjpatton:
Thank you @hannahdaviscrypto !

PR #293, opened 2023-08-31
Revert PRG name change in the change log

cjpatton:
We were a bit sloppy in the PRG -> XOF rename in that we changed the names in the change log as well. We want to keep the change log consistent across drafts, so revert these changes.


draft-irtf-cfrg-vdaf.md
```diff
@@ -401,13 +401,13 @@ security considerations for VDAFs.
 
 05:
```
cjpatton, line 403: Reviwer tip: Check out the branch and look the Change Log section in `git diff draft-irtf-cfrg-vdaf-06 draft-irtf-cfrg-vdaf.md`

divergentdave (commented):
Edit: I ran into a GitHub bug, I clicked the "cancel review" button and it left a comment!

PR #294, opened 2023-09-06
poc: Update rejection test for XofShake128

### LINKED ISSUE #288: Re-generate SHA-3 output test
When breaking a test, it's important to make clear to the reader why the test is broken (and how it should be fixed).
Bonus: We should file an issue to remind us to fix this test, and reference the issue in this comment.
```suggestion
    # TODO: Update the test to account for the change from cSHAKE128 to SHAKE128.
    # assert expanded_vec[-1] == Field64(13681157193520586550)
```

_Originally posted by @cjpatton in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/285#discussion_r1303323003_

cjpatton:
Closes #288.

### DIFF
--- a/poc/xof.py
+++ b/poc/xof.py
@@ -184,13 +184,13 @@ if __name__ == '__main__':
     # https://github.com/divergentdave/vdaf-rejection-sampling-search
     expanded_vec = XofShake128.expand_into_vec(
         Field64,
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
LGTM, and I confirmed this falls through in the rejection sampling loop exactly once

PR #296, opened 2023-09-27
Fix __name__ access in poc

junyechen1996:
__name__ is an attribute of a class instead of that of an instance. flp.Valid is an instance, so accessing __name__ will fail with another exception if the FLP test fails.


divergentdave (approved):
Thank you!

Issue #177, opened 2023-02-28
Prio3: Consider using a smaller field and re-running the proof verification multiple times


cjpatton:
This would trade communication cost for CPU time.

divergentdave:
Summarizing some offline discussions: Running verification multiple times to improve soundness should work straightforwardly if there is no joint randomness. Once the report is fixed, the only coin that the soundness probabilities are evaluated over is the query randomness, and checking the polynomials at multiple points gives us independent events, so the protocol's overall soundness error would look something like the product of the soundness error from each query run.

However, once we re-introduce joint randomness, the joint randomness values would be constant across each proof verification re-run, and thus the probability of repeated FLP verification detecting a dishonest submission is correlated. There's now another coin involved, and repeatedly running proof verification doesn't re-flip it. Thus, the soundness error you get from running proof verification repeatedly would be dominated by the terms that come from the joint randomness, and the overall soundness would not get exponentially better with more verifications. The query randomness values are used in polynomial identity tests that show that gadget input and output values are consistent, while the joint randomness values are used by validity circuits to compress intermediate values from multiple gadgets into a single field element. (typically, by computing a random linear combination of polynomials that are each zero for valid inputs) Concretely, a dishonest verifier could choose an invalid measurement, honestly evaluate all the gadgets in the validity circuit, and adversarially choose joint randomness blinds such that the circuit output is zero through cancellation. If they wanted to attack Prio3Sum, they could submit a measurement with, say, 2 in every "bit", and search until they find a joint randomness value that satisfies $x + x^2 + x^3 + ... + x^n = 0$. This would take an average of $\approx(2^{32}/n)/2$ tries with Field32, which is feasible.

We could achieve reasonable soundness with smaller fields if we re-ran the validity circuit (and thus proof generation) multiple times with different joint randomness values, but this would cancel out the client-to-aggregator communication benefits of using a smaller field.

For each circuit that currently uses joint randomness, we could imagine dropping the compression step, and having validity circuits output a vector of field elements. This would increase both client-to-aggregator and aggregator-to-aggregator communication, but the tradeoff for using smaller fields might be worth it with moderately-sized circuits (i.e., only a few gadget evaluations).

Those are the only tricks I can think of at the moment to get around this. Of course, they'd need further analysis.

simon-friedberger:
Can we construct a new joint randomness for each verification using the iteration of the check as input? (Or just iteratively rehashing the joint randomness.)

[...]

PR #298, opened 2023-10-03
Add poc changes for [#177]

albertpl:
1. POC implementation for mutliproof proposal in  [#177](https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/177)
    1. Support multiproofs in `Prio3`
    2. Add new `Prio3SumVec` variant, i.e. `Prio3SumVecWithMultiproof`, with configuration (field size, number of proofs)
    3. Add `with_field` class methods to introduce new `SumVec` with configurable field size

### DIFF
--- a/poc/flp_generic.py
+++ b/poc/flp_generic.py
@@ -120,6 +120,12 @@ class Valid:
         if len(joint_rand) != self.JOINT_RAND_LEN:
             raise ERR_INPUT
 
+    @classmethod
+    def with_field(cls, field: field.FftField):
+        class _ValidWithField(cls):
+            Field = field
+        return _ValidWithField
+
 
 class ProveGadget:
     def __init__(self, Field, wire_seeds, g, g_calls):
@@ -829,6 +835,22 @@ class TestAverage(Sum):
         return total // num_measurements
 
 
+# Test encoding, truncation, then decoding.
+def test_encode_truncate_decode(flp, measurements):
+    for measurement in measurements:
+        assert measurement == flp.decode(
+            flp.truncate(flp.encode(measurement)), 1)
+
+
+def test_encode_truncate_decode_with_fft_fields(cls, measurements, *args):
+    for f in [field.Field64, field.Field96, field.Field128]:
+        cls_with_field = cls.with_field(f)
+        assert cls_with_field.Field == f
+        obj = cls_with_field(*args)
+        assert isinstance(obj, cls)
+        test_encode_truncate_decode(FlpGeneric(obj), measurements)
+
+
 def test():
     flp = FlpGeneric(Count())
     test_flp_generic(flp, [
@@ -848,9 +870,7 @@ def test():
         (flp.encode(2 ** 10 - 1), True),
         (flp.Field.rand_vec(10), False),
     ])
-    # Roundtrip test with no proof generated.
-    for meas in [0, 100, 2 ** 10 - 1]:
-        assert meas == flp.decode(flp.truncate(flp.encode(meas)), 1)
+    test_encode_truncate_decode(flp, [0, 100, 2 ** 10 - 1])
 
     flp = FlpGeneric(Histogram(4, 2))
     test_flp_generic(flp, [
@@ -864,10 +884,9 @@ def test():
     ])
 
     # SumVec with length 2, bits 4, chunk len 1.
-    flp = FlpGeneric(SumVec(2, 4, 1))
-    # Roundtrip test with no proof generated.
-    for meas in [[1, 2], [3, 4], [5, 6], [7, 8]]:
-        assert meas == flp.decode(flp.truncate(flp.encode(meas)), 1)
+    test_encode_truncate_decode_with_fft_fields(SumVec,
+                                                [[1, 2], [3, 4], [5, 6], [7, 8]],
+                                                2, 4, 1)
 
     flp = FlpGeneric(TestMultiGadget())
     test_flp_generic(flp, [
--- a/poc/vdaf_prio3.py
+++ b/poc/vdaf_prio3.py
@@ -7,6 +7,7 @@ import flp_generic
 import xof
 from common import (ERR_INPUT, ERR_VERIFY, TEST_VECTOR, Unsigned, byte, concat,
                     front, vec_add, vec_sub, zeros)
+from field import FftField, Field64, Field128
 from vdaf import Vdaf, test_vdaf
 
 USAGE_MEAS_SHARE = 1
@@ -31,6 +32,9 @@ class Prio3(Vdaf):
     RAND_SIZE = None  # Computed from `Xof.SEED_SIZE` and `SHARES`
     ROUNDS = 1
     SHARES = None  # A number between `[2, 256)` set later
+
+    # Operational parameters
+    PROOFS = 1  # Number of proofs, in range `[1, 256)`
 
     # Types required by `Vdaf`
     Measurement = Flp.Measurement
@@ -73,7 +77,7 @@ class Prio3(Vdaf):
     def prep_init(Prio3, verify_key, agg_id, _agg_param,
                   nonce, public_share, input_share):
         k_joint_rand_parts = public_share
-        (meas_share, proof_share, k_blind) = \
+        (meas_share, proofs_share, k_blind) = \
             Prio3.expand_input_share(agg_id, input_share)
         out_share = Prio3.Flp.truncate(meas_share)
 
@@ -86,18 +90,27 @@ class Prio3(Vdaf):
             k_joint_rand_parts[agg_id] = k_joint_rand_part
             k_corrected_joint_rand = Prio3.joint_rand_seed(
                 k_joint_rand_parts)
-            joint_rand = Prio3.joint_rand(k_corrected_joint_rand)
+            joint_rands = Prio3.joint_rands(k_corrected_joint_rand)
 
         # Query the measurement and proof share.
-        query_rand = Prio3.query_rand(verify_key, nonce)
-        verifier_share = Prio3.Flp.query(meas_share,
-                                         proof_share,
-                                         query_rand,
-                                         joint_rand,
-                                         Prio3.SHARES)
+        query_rands = Prio3.query_rands(verify_key, nonce)
+        verifiers_share = []
+        for _ in range(Prio3.PROOFS):
+            proof_share, proofs_share = front(
+                Prio3.Flp.PROOF_LEN, proofs_share)
+            query_rand, query_rands = front(
+                Prio3.Flp.QUERY_RAND_LEN, query_rands)
+            if Prio3.Flp.JOINT_RAND_LEN > 0:
+                joint_rand, joint_rands = front(
+                    Prio3.Flp.JOINT_RAND_LEN, joint_rands)
+            verifiers_share += Prio3.Flp.query(meas_share,
+                                               proof_share,
+                                               query_rand,
+                                               joint_rand,
+                                               Prio3.SHARES)
 
         prep_state = (out_share, k_corrected_joint_rand)
-        prep_share = (verifier_share, k_joint_rand_part)
+        prep_share = (verifiers_share, k_joint_rand_part)
         return (prep_state, prep_share)
 
     @classmethod
@@ -115,16 +128,19 @@ class Prio3(Vdaf):
     @classmethod
     def prep_shares_to_prep(Prio3, _agg_param, prep_shares):
         # Unshard the verifier shares into the verifier message.
-        verifier = Prio3.Flp.Field.zeros(Prio3.Flp.VERIFIER_LEN)
+        verifiers = Prio3.Flp.Field.zeros(
+            Prio3.Flp.VERIFIER_LEN * Prio3.PROOFS)
         k_joint_rand_parts = []
-        for (verifier_share, k_joint_rand_part) in prep_shares:
-            verifier = vec_add(verifier, verifier_share)
+        for (verifiers_share, k_joint_rand_part) in prep_shares:
+            verifiers = vec_add(verifiers, verifiers_share)
             if Prio3.Flp.JOINT_RAND_LEN > 0:
                 k_joint_rand_parts.append(k_joint_rand_part)
 
-        # Verify that the proof is well-formed and the input is valid.
-        if not Prio3.Flp.decide(verifier):
-            raise ERR_VERIFY  # proof verifier check failed
+        # Verify that each proof is well-formed and accepts the measurement.
+        for _ in range(Prio3.PROOFS):
+            verifier, verifiers = front(Prio3.Flp.VERIFIER_LEN, verifiers)
+            if not Prio3.Flp.decide(verifier):
+                raise ERR_VERIFY  # proof verifier check failed
 
         # Combine the joint randomness parts computed by the
         # Aggregators into the true joint randomness seed. This is
@@ -158,7 +174,7 @@ class Prio3(Vdaf):
             k_helper_seeds[i]
             for i in range(0, (Prio3.SHARES-1) * 2, 2)
         ]
-        k_helper_proof_shares = [
+        k_helper_proofs_shares = [
             k_helper_seeds[i]
             for i in range(1, (Prio3.SHARES-1) * 2, 2)
         ]
@@ -172,13 +188,17 @@ class Prio3(Vdaf):
                 Prio3.helper_meas_share(j+1, k_helper_meas_shares[j]),
             )
 
-        # Generate the proof and shard it into proof shares.
-        prove_rand = Prio3.prove_rand(k_prove)
-        leader_proof_share = Prio3.Flp.prove(meas, prove_rand, [])
+        # Generate the proofs and shard it into proof(s) shares.
+        prove_rands = Prio3.prove_rands(k_prove)
+        leader_proofs_share = []
+        for _ in range(Prio3.PROOFS):
+            prove_rand, prove_rands = front(
+                Prio3.Flp.PROVE_RAND_LEN, prove_rands)
+            leader_proofs_share += Prio3.Flp.prove(meas, prove_rand, [])
         for j in range(Prio3.SHARES-1):
-            leader_proof_share = vec_sub(
-                leader_proof_share,
-                Prio3.helper_proof_share(j+1, k_helper_proof_shares[j]),
+            leader_proofs_share = vec_sub(
+                leader_proofs_share,
+                Prio3.helper_proofs_share(j+1, k_helper_proofs_shares[j]),
             )
 
         # Each Aggregator's input share contains its measurement share
@@ -186,13 +206,13 @@ class Prio3(Vdaf):
         input_shares = []
         input_shares.append((
             leader_meas_share,
-            leader_proof_share,
+            leader_proofs_share,
             None,
         ))
         for j in range(Prio3.SHARES-1):
             input_shares.append((
                 k_helper_meas_shares[j],
-                k_helper_proof_shares[j],
+                k_helper_proofs_shares[j],
                 None,
             ))
         return (None, input_shares)
@@ -204,7 +224,7 @@ class Prio3(Vdaf):
             k_helper_seeds[i]
             for i in range(0, (Prio3.SHARES-1) * 3, 3)
         ]
-        k_helper_proof_shares = [
+        k_helper_proofs_shares = [
             k_helper_seeds[i]
             for i in range(1, (Prio3.SHARES-1) * 3, 3)
         ]
@@ -230,14 +250,21 @@ class Prio3(Vdaf):
             0, k_leader_blind, leader_meas_share, nonce))
 
         # Generate the proof and shard it into proof shares.
-        prove_rand = Prio3.prove_rand(k_prove)
-        joint_rand = Prio3.joint_rand(
+        prove_rands = Prio3.prove_rands(k_prove)
+        joint_rands = Prio3.joint_rands(
             Prio3.joint_rand_seed(k_joint_rand_parts))
-        leader_proof_share = Prio3.Flp.prove(meas, prove_rand, joint_rand)
+        leader_proofs_share = []
+        for _ in range(Prio3.PROOFS):
+            prove_rand, prove_rands = front(
+                Prio3.Flp.PROVE_RAND_LEN, prove_rands)
+            joint_rand, joint_rands = front(
+                Prio3.Flp.JOINT_RAND_LEN, joint_rands)
+            leader_proofs_share += Prio3.Flp.prove(meas,
+                                                   prove_rand, joint_rand)
         for j in range(Prio3.SHARES-1):
-            leader_proof_share = vec_sub(
-                leader_proof_share,
-                Prio3.helper_proof_share(j+1, k_helper_proof_shares[j]),
+            leader_proofs_share = vec_sub(
+                leader_proofs_share,
+                Prio3.helper_proofs_share(j+1, k_helper_proofs_shares[j]),
             )
 
         # Each Aggregator's input share contains its measurement share,
@@ -246,13 +273,13 @@ class Prio3(Vdaf):
         input_shares = []
         input_shares.append((
             leader_meas_share,
-            leader_proof_share,
+            leader_proofs_share,
             k_leader_blind,
         ))
         for j in range(Prio3.SHARES-1):
             input_shares.append((
                 k_helper_meas_shares[j],
-                k_helper_proof_shares[j],
+                k_helper_proofs_shares[j],
                 k_helper_blinds[j],
             ))
         return (k_joint_rand_parts, input_shares)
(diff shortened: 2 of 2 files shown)


poc/flp_generic.py
```diff
@@ -658,6 +658,12 @@ class SumVec(Valid):
     JOINT_RAND_LEN = 1
     OUTPUT_LEN = None  # Set by constructor
 
+    @classmethod
+    def with_field(SumVec, field: field.FftField):
```
albertpl, line 662: I choose to this approach, instead of making `field` a `__init__` parameter  because
1. Follow existing patterns in `vdaf_prio3.py` where we using `Prio3` as a template to create `Prio3SumVec` etc.
2. `SumVec.Field` is closer to template/generic parameters than instance properties, IMHO

poc/vdaf_prio3.py
```diff
@@ -483,6 +510,51 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
 
 
+class Prio3SumVecWithMultiproof(Prio3SumVec):
+    ID = 0xFFFFFFFF  # To be set by `with_params`
+
+    # Operational parameters.
+    test_vec_name = 'Prio3SumVecWithMultiproof'
+
+    @staticmethod
+    def is_recommended(valid_cls,
```
albertpl, line 520: We might need to take other parameters from the underlying FLP circuits/gadgets into considerations? e.g. `Valid.GADGET_CALLS`, `GADGET.DEGREE`? What constrains shall we specify ?

poc/vdaf_prio3.py
```diff
@@ -483,6 +510,51 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+        if field == Field64:
+            return num_proofs >= 2
+        elif field == Field96:
+            return num_proofs >= 2
+        elif field == Field128:
+            return num_proofs >= 1
+        return False
+
+    @staticmethod
+    def algorithm_id_from(num_proofs: Unsigned,
```
albertpl, line 533: What algorithm IDs shall we allocate for this variants? 

Let $N$ be the next available algorithm ID. Maybe we can start with $N$ be Field64 with `PROOFS=2` and $N+1$ be Field64 with `PROOFS=3`? Then allocate more algorithm IDs whenever there is request?

poc/vdaf_prio3.py
```diff
@@ -302,7 +327,7 @@ def query_rand(Prio3, verify_key, nonce):
             verify_key,
```
cjpatton, line 327: ^ s/query_rand/query_rands/ for consistency with joint_rands

poc/vdaf_prio3.py
```diff
@@ -86,15 +89,24 @@ def prep_init(Prio3, verify_key, agg_id, _agg_param,
...
 
         # Query the measurement and proof share.
-        query_rand = Prio3.query_rand(verify_key, nonce)
-        verifier_share = Prio3.Flp.query(meas_share,
-                                         proof_share,
-                                         query_rand,
-                                         joint_rand,
-                                         Prio3.SHARES)
+        query_rands = Prio3.query_rand(verify_key, nonce)
+        verifier_share = []
```
cjpatton, line 96: Hmm, this is a bit odd; elsewhere we have "proof shares" and "query rands", but here we have just one "verifier share". WDYTA changing this to "verifier shares"? One downside is that, during prep_shares_to_prep we combine the "verifier shares" into the "verifier", so we'd need to change this to "list of verifier shares" into the "verifiers".

Let me know if I'm not making sense :)

poc/vdaf_prio3.py
```diff
@@ -483,6 +510,51 @@ class Prio3HistogramWithLength(Prio3Histogram):
         return Prio3HistogramWithLength
 
 
+class Prio3SumVecWithMultiproof(Prio3SumVec):
```
cjpatton, line 513: We want to make it clear that this is not (yet) on the standards track. We could prefix the class name with `Test`, or maybe just add a comment clarifying what this is intended for.

poc/vdaf_prio3.py
```diff
@@ -483,6 +510,51 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+    ID = 0xFFFFFFFF  # To be set by `with_params`
```
cjpatton, line 514: Why not just set it here? If this behavior is needed, then don't set the ID here.

poc/vdaf_prio3.py
cjpatton: yeah, for different choices of `PROOFS`, we'd need to distinct code points (i.e., algorithm IDs)

poc/vdaf_prio3.py
```diff
@@ -323,14 +348,14 @@ def joint_rand_seed(Prio3, k_joint_rand_parts):
         )
 
     @classmethod
-    def joint_rand(Prio3, k_joint_rand_seed):
+    def joint_rands(Prio3, k_joint_rand_seed):
```
cjpatton, line 351: As @divergentdave suggested, let's incorporate `Prio3.PROOFS` into the transcript for deriving the joint rands.

cjpatton (changes requested):
Looking good so far! Very clean for the most part. One high level ask: Please add more context to the commit message. A reference to the issue is somewhat useful, but it's helpful to have a summary of the change in the `git log` as well. (We rely on this heavily when preparing drafts.)

poc/vdaf_prio3.py
junyechen1996: If Prio3’s derivation of domain separation tag already includes algorithm ID, is it sufficient to make algorithm ID specify a distinct combination of `(Field, PROOFS)`, per thread [here](https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/298#discussion_r1344864217)?

poc/common.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, msg):
 ERR_ENCODE = Error('encode failure')
 ERR_INPUT = Error('invalid input parameter')
 ERR_VERIFY = Error('verification of the user\'s input failed')
+ERR_PARAMS_NOT_RECOMMENEDED = Error('parameters are not recommended')
```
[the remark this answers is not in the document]
albertpl: Sounds good. I will create a new class `ParamsNotRecommendedError` inherited from `Exception`.

poc/vdaf_prio3.py
cjpatton: That's one possibility, but it somewhat depends on how "configurable" we need the value of `PROOFS` to be. Ideally we can fix it for a given choice of circuit and field.

cjpatton: I think for now it makes sense to include it in the transcript.

poc/common.py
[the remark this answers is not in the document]
cjpatton: Actually I would follow what we do else where and just `raise ValueError('something bad happened')` like we do elsewhere.

poc/vdaf_prio3.py
albertpl: sorry to confuse. I want to convey *a share* of multiple verifier message.
I guess we can use `verifiers_share`, i.e. append **s** to verifier, instead of share, what do you think?

poc/vdaf_prio3.py
albertpl: OK. I will add a comment.

poc/vdaf_prio3.py
albertpl: Along the same line, (notice **s** appended to `proof`)
`leader_proof_share` -> `leader_proofs_share`
`helper_proof_share` -> `helper_proofs_share`

albertpl:
[Note, I will squash & summarize all changes in the commit, before merging this PR]

poc/vdaf_prio3.py
albertpl: >I think for now it makes sense to include it in the transcript.

Just to clarify. And do we want to use binder string?

poc/vdaf_prio3.py
cjpatton: Yeah I like that idea. It's grammatically a little hairy but it makes the variable names consistent.

poc/vdaf_prio3.py
cjpatton: Yup, at the step where we derive the joint rands from the `k_joint_rand_seed`.

poc/vdaf_prio3.py
```diff
@@ -323,14 +348,15 @@ def joint_rand_seed(Prio3, k_joint_rand_parts):
...
         """Derive the joint randomness from its seed."""
+        binder = b'' if Prio3.PROOFS == 1 else byte(Prio3.PROOFS)
```
albertpl, line 353: Choose to be backward compatible, i.e. only embedding `Prio3.PROOFS` when it is > 1.

poc/vdaf_prio3.py
albertpl: If we treat (`PROOFS, Field`) as parameter for the underlying algorithm, i.e. same as `SHARES`, it makes senses to add them into joint randomness derivation. 
1. If we do so, maybe we can allocate one algorithm ID/code point for Prio3SumVec variant with multiproofs? 
2. Shall we add field choice into joint randomness derivation too (binder) ?

cjpatton: > If we treat (`PROOFS, Field`) as parameter for the underlying algorithm, i.e. same as `SHARES`, it makes senses to add them into joint randomness derivation.
> 
>     1. If we do so, maybe we can allocate one algorithm ID/code point for Prio3SumVec variant with multiproofs?

I don't think we should actually allocate a codepoint until we know more about robustness. In the meantime, one is free to use any of the unreserved codepoints however they wish.


>     2. Shall we add field choice into joint randomness derivation too (binder) ?

We maybe don't need to unless we plan to use more than one field. Isn't `Field64` the only one that's needed?

poc/vdaf_prio3.py
cjpatton: Yup, good idea.

poc/flp_generic.py
```diff
@@ -829,6 +835,22 @@ def decode(self, output, num_measurements):
...
 
+# Test encoding, truncation, then decoding.
+def test_encode_truncate_decode(flp, measurements):
+    for measurement in measurements:
+        assert measurement == flp.decode(
+            flp.truncate(flp.encode(measurement)), 1)
+
+
+def test_encode_truncate_decode_with_fft_fields(cls, measurements, *args):
+    for _field in [field.Field64, field.Field96, field.Field128]:
```
cjpatton, line 846: Style question: Why prepend the underscore (`_`)? Is this because `field` is already defined? Is this following a specific Python convention?

I write a lot of rust these days, so I think of this convention as a way of differentiating multiple variables that I want to be anonymous.

poc/vdaf_prio3.py
cjpatton: For consistency with `shard()`
```suggestion
        verifiers_share = []
```

poc/vdaf_prio3.py
```diff
@@ -115,16 +127,19 @@ def prep_next(Prio3, prep, prep_msg):
     @classmethod
     def prep_shares_to_prep(Prio3, _agg_param, prep_shares):
         # Unshard the verifier shares into the verifier message.
-        verifier = Prio3.Flp.Field.zeros(Prio3.Flp.VERIFIER_LEN)
+        verifiers = Prio3.Flp.Field.zeros(
+            Prio3.Flp.VERIFIER_LEN * Prio3.PROOFS)
         k_joint_rand_parts = []
         for (verifier_share, k_joint_rand_part) in prep_shares:
```
cjpatton:
```suggestion
        for (verifiers_share, k_joint_rand_part) in prep_shares:
```

poc/vdaf_prio3.py
```diff
@@ -173,8 +188,12 @@ def shard_without_joint_rand(Prio3, meas, seeds):
             )
 
         # Generate the proof and shard it into proof shares.
-        prove_rand = Prio3.prove_rand(k_prove)
-        leader_proof_share = Prio3.Flp.prove(meas, prove_rand, [])
+        prove_rands = Prio3.prove_rands(k_prove)
+        leader_proof_share = []
```
cjpatton:
```suggestion
        leader_proofs_share = []
```

poc/vdaf_prio3.py
```diff
@@ -230,10 +249,16 @@ def shard_with_joint_rand(Prio3, meas, nonce, seeds):
             0, k_leader_blind, leader_meas_share, nonce))
 
         # Generate the proof and shard it into proof shares.
-        prove_rand = Prio3.prove_rand(k_prove)
-        joint_rand = Prio3.joint_rand(
+        prove_rands = Prio3.prove_rands(k_prove)
+        joint_rands = Prio3.joint_rands(
             Prio3.joint_rand_seed(k_joint_rand_parts))
-        leader_proof_share = Prio3.Flp.prove(meas, prove_rand, joint_rand)
+        leader_proof_share = []
```
cjpatton:
```suggestion
        leader_proofs_share = []
```

poc/vdaf_prio3.py
cjpatton, line 353: This imposes an upper bound on the value of `Prio3.PROOFS`. I think we should note this upper bound where `PROOFS` is first declared.

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,53 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+class Prio3SumVecWithMultiproof(Prio3SumVec):
+    # Operational parameters.
+    test_vec_name = 'Prio3SumVecWithMultiproof'
+
+    @staticmethod
+    def is_recommended(valid_cls,
+                       num_proofs: Unsigned,
+                       field: FftField) -> bool:
+        if num_proofs >= 0x100:
+            # Currently we encod this as one byte while deriving joing rands
```
cjpatton, line 524:
```suggestion
            # We encode this as one byte while deriving joing rands
```

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,53 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+            return False
+        # To be confirmed
```
cjpatton, line 526: If this comment is referring to a decision that still needs to be made, then I'd make it more clear.
```suggestion
        # TODO(issue#177) Decide how many proofs to use for each field type.
```

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,53 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+        if field == Field64:
+            return num_proofs >= 2
+        elif field == Field96:
```
cjpatton, line 529: Do you plan to potentially use this field? Before you do, I think we should figure out if we can find a better prime modulus. We'd like to find something that's closer to a power of 2. (cc/ @divergentdave)

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,53 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+            return num_proofs >= 2
+        elif field == Field128:
+            return num_proofs >= 1
+        return False
+
+    @staticmethod
+    def algorithm_id_from(num_proofs: Unsigned,
+                          field: FftField) -> Unsigned:
+        # To be confirmed
+        return 0xFFFFFFFF
```
cjpatton, line 539: Since this is hard-coded for now, I think we should get rid of this function and declare this the same way we do for the other Prio3 variants: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/poc/vdaf_prio3.py#L420

poc/vdaf_prio3.py
```diff
@@ -506,6 +581,56 @@ class TestPrio3AverageWithBits(TestPrio3Average):
         return TestPrio3AverageWithBits
 
 
+def _test_prio3sumvec(num_proofs: Unsigned, field: FftField):
```
cjpatton, line 584: Same question as above: What convention are you following by prefixing this function with an underscore?

poc/vdaf_prio3.py
```diff
@@ -506,6 +581,56 @@ class TestPrio3AverageWithBits(TestPrio3Average):
...
+    valid_cls = flp_generic.SumVec.with_field(field)
+    if not Prio3SumVecWithMultiproof.is_recommended(valid_cls, num_proofs, field):
+        return
```
cjpatton, line 587: It seems odd to silently skip the test if the number of proofs is not recommended. Also, should we test that it's greater than 0?

poc/vdaf_prio3.py
cjpatton: Reviewer note: I've confirmed the test vectors haven't changed.

cjpatton (commented):
Looking very good! High level ask: Make sure that variable naming conventions are consistent across sharding and preparation.

poc/flp_generic.py
albertpl: >Why prepend the underscore (_)?  Is this because field is already defined?

Yes.  Due to name collision. And we need to access `field.Field64` as well.
```
import field
```

poc/vdaf_prio3.py
albertpl: Do you mean that we should do negative tests? 

Like test with invalid configurations and expect it to throw exception? 
If so, I would like to introduce a dummy exception, like this
```
class MultiproofParamsNotRecommended(Exception):
    pass
```

This way we can throw & check this error. Let other `ValueError` to fail?

poc/vdaf_prio3.py
albertpl: I want to convey that this is an *internal* helper for `test_prio3sumvec`, not to used directly. What naming convention shall we follow?

poc/vdaf_prio3.py
albertpl: No. So we will limit to `Field64` only then?

poc/vdaf_prio3.py
albertpl: I think this is reflected in the latest version.

poc/vdaf_prio3.py
albertpl: I think this is reflected in the latest version.

poc/vdaf_prio3.py
albertpl: I think this is reflected in the latest version.

poc/vdaf_prio3.py
cjpatton: Yes. Better to not use `Field96` unless we have to :)

poc/vdaf_prio3.py
cjpatton: Sounds reasonable. I ask because I don't use Python a lot and I'm wondering about its conventions. I don't have an alternative in mind, so I'll just defer to you.

poc/vdaf_prio3.py
cjpatton: I'm suggesting that tests to enforce the conditions they expect rather than skip testing if the conditions aren't meant. I would `assert` instead of `return`ing.

poc/flp_generic.py
cjpatton: Let's maybe go with `f` here.

poc/vdaf_prio3.py
albertpl: Any chance you are not checking the [latest](i.e. 025c1c9) commit? Should be reflected, no?

poc/vdaf_prio3.py
albertpl: Yes. We check this upper bound in `Prio3SumVecWithMultiproof.is_recommended`.  I add comments at first declaration as well.
Having said that, I am not sure what `PROOFS` we will put in the draft? I suppose it depends on robustness analysis. In any cases, due to computation cost, I would imagine > 255 proofs is prohibitive for most use cases.

poc/vdaf_prio3.py
albertpl: Got it. Since we agree to allow only one field size, this is much easier now. I will update.

poc/vdaf_prio3.py
albertpl: OK. Only `Field64` for now.

poc/vdaf_prio3.py
albertpl: Will do. This does imply we have only one code point for this multiproof variant though

poc/vdaf_prio3.py
```diff
@@ -31,6 +32,7 @@ class Prio3(Vdaf):
     RAND_SIZE = None  # Computed from `Xof.SEED_SIZE` and `SHARES`
     ROUNDS = 1
     SHARES = None  # A number between `[2, 256)` set later
+    PROOFS = 1  # Number of proofs, default to 1, otherwise see `Prio3SumVecWithMultiproof.is_recommended`
```
cjpatton, line 35:
```suggestion
    
    # Operational parameters
    PROOFS = 1  # Number of proofs, in range `[1, 256)`
```

poc/vdaf_prio3.py
```diff
@@ -115,16 +126,19 @@ def prep_next(Prio3, prep, prep_msg):
...
-            verifier = vec_add(verifier, verifier_share)
+        for (verifiers_share, k_joint_rand_part) in prep_shares:
+            verifiers = vec_add(verifiers, verifiers_share)
             if Prio3.Flp.JOINT_RAND_LEN > 0:
                 k_joint_rand_parts.append(k_joint_rand_part)
 
-        # Verify that the proof is well-formed and the input is valid.
-        if not Prio3.Flp.decide(verifier):
-            raise ERR_VERIFY  # proof verifier check failed
+        # Verify that all the proofs are well-formed and the input is valid.
```
cjpatton, line 137:
```suggestion
        # Verify that each proof is well-formed and accepts the measurement.
```

poc/vdaf_prio3.py
```diff
@@ -172,27 +186,31 @@ def shard_without_joint_rand(Prio3, meas, seeds):
                 Prio3.helper_meas_share(j+1, k_helper_meas_shares[j]),
             )
 
-        # Generate the proof and shard it into proof shares.
-        prove_rand = Prio3.prove_rand(k_prove)
-        leader_proof_share = Prio3.Flp.prove(meas, prove_rand, [])
+        # Generate the proof(s) and shard it into proof(s) shares.
```
cjpatton, line 189:
```suggestion
        # Generate the proofs and shard it into proof(s) shares.
```

poc/vdaf_prio3.py
```diff
@@ -367,14 +393,15 @@ def test_vec_set_type_param(Prio3, test_vec):
...
+        (meas_share, proofs_share, k_blind) = input_share
         encoded = bytes()
-        if type(meas_share) == list and type(proof_share) == list:  # leader
+        if type(meas_share) == list and type(proofs_share) == list:  # leader
+            assert len(proofs_share) == Prio3.Flp.PROOF_LEN * Prio3.PROOFS
             encoded += Prio3.Flp.Field.encode_vec(meas_share)
-            encoded += Prio3.Flp.Field.encode_vec(proof_share)
-        elif type(meas_share) == bytes and type(proof_share) == bytes:  # helper
+            encoded += Prio3.Flp.Field.encode_vec(proofs_share)
+        elif type(meas_share) == bytes and type(proofs_share) == bytes:  # helper
```
cjpatton, line 402:
```suggestion
        elif type(meas_share) == bytes and type(proofs_share) == bytes:  # Helper
```

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,41 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+
+    @staticmethod
+    def is_recommended(valid_cls,
+                       num_proofs: Unsigned,
+                       field: FftField) -> bool:
+        # TODO(issue#177) Decide how many proofs to use.
+        if field == Field64:
+            # the upper bound is due to the fact
+            # we encode it using one byteb in `joint_rands`
+            return 2 <= num_proofs < 0x100
```
cjpatton, line 527:
```suggestion
            return 2 <= num_proofs < 256
```

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,41 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
 
+# Experimental multiproof variant of Prio3SumVec
+class Prio3SumVecWithMultiproof(Prio3SumVec):
+    # Operational parameters.
+    test_vec_name = 'Prio3SumVecWithMultiproof'
+
+    @staticmethod
+    def is_recommended(valid_cls,
+                       num_proofs: Unsigned,
+                       field: FftField) -> bool:
```
cjpatton, line 522: Avoid shadowing the imported `field` module
```suggestion
                       f: FftField) -> bool:
```

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,41 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+        # TODO(issue#177) Decide how many proofs to use.
+        if field == Field64:
+            # the upper bound is due to the fact
+            # we encode it using one byteb in `joint_rands`
+            return 2 <= num_proofs < 0x100
+        return False
```
cjpatton, line 528: Future proofing
```suggestion
        elif field == Field128:
            return 1 <= num_proofs < 256
        return False
```

poc/vdaf_prio3.py
```diff
@@ -506,6 +569,55 @@ class TestPrio3AverageWithBits(TestPrio3Average):
         return TestPrio3AverageWithBits
 
 
+def _test_prio3sumvec(num_proofs: Unsigned, field: FftField):
+    valid_cls = flp_generic.SumVec.with_field(field)
+    assert Prio3SumVecWithMultiproof.is_recommended(
+        valid_cls, num_proofs, field)
+
+    cls = Prio3SumVecWithMultiproof \
+        .with_params(10, 8, 9, num_proofs=num_proofs, field=field) \
```
cjpatton, line 578:
```suggestion
        .with_params(10, 8, 9, num_proofs, field) \
```

cjpatton (approved):
Looks good, just nit-picky comments left.

poc/vdaf_prio3.py
```diff
@@ -483,6 +511,41 @@ class Prio3HistogramWithLength(Prio3Histogram):
...
+    test_vec_name = 'Prio3SumVecWithMultiproof'
+
+    @staticmethod
+    def is_recommended(valid_cls,
+                       num_proofs: Unsigned,
+                       field: FftField) -> bool:
+        # TODO(issue#177) Decide how many proofs to use.
+        if field == Field64:
+            # the upper bound is due to the fact
+            # we encode it using one byteb in `joint_rands`
```
divergentdave, line 526: nit: typo
```suggestion
            # we encode it using one byte in `joint_rands`
```

divergentdave (approved):

Issue #287, opened 2023-08-23
Prio3 VDAF for vectors with a bounded number of 1s


divergentdave:
Section 4.5 of [Samplable Anonymous Aggregation for Private Federated Data Analysis](https://arxiv.org/pdf/2307.15017.pdf) notes that it's possible to efficiently validate that a vector consists of zeros and ones, with at most some constant number of ones. This could be done by generalizing Prio3Histogram. (Open question: does this belong in this document or a separate one?) Here's how I'd implement this:

* The parameters would be the length, $n$, the chunk length, and the maximum number of ones, $m$.
* Input length would be $n+\left\lfloor{log_2(m)}\right\rfloor+1$. Let $l_m=\left\lfloor{log_2(m)}\right\rfloor+1$, the number of bits needed to represent $m$.
* Encoding: The first $n$ field elements are just projected measurement values. Let $c=2^{l_m}-1-m$, a constant offset to let us achieve the sum limit we want. Then, for input values $x_i$, compute $c+\sum_i{x_i}$, encode it as a bit vector, and append it to the encoded measurement.
* Truncation: Take the first $n$ elements.
* Decoding: $id$
* Validity circuit: Evaluate the range check polynomial $x(x-1)$ on all `INPUT_LEN` elements, and combine the results in a random linear combination (using the parallel sum optimization). Separately, evaluate the following, where $x_i$ are the measurement values from the encoded measurement, and $s_j$ are the bits of the $c+\sum_i{x_i}$ from the encoded measurement. $$c + \sum_{i=0}^{n-1} {x_i} - \sum_{j=0}^{l_m-1} {2^j s_j}$$ Add the range check and sum check intermediate values together in another random linear combination, and return the result.

cjpatton:
If the generalization is straight-forward, then I think we can put it here.

Can you explain why this range check is sufficient? (Suppose $m$ is not a power of 2.)

Can we make the $m=1$ case wire-compatible with the current version of the circuit?

divergentdave:
[...]

PR #300, opened 2023-10-12
Implement Prio3MultiHotHistogram that checks bounded number of 1s

junyechen1996:
Add a reference implementation of the idea in issue #287, which needs a new FLP 'MultiHotHistogram' and a new Prio3 type 'Prio3MultiHotHistogram' that checks a Client's measurement has a bounded number of 1s.

### DIFF
--- a/poc/flp_generic.py
+++ b/poc/flp_generic.py
@@ -580,7 +580,7 @@ class Histogram(Valid):
 
     def __init__(self, length, chunk_length):
         """
-        Instantiate an instace of the `Histogram` circuit with the given
+        Instantiate an instance of the `Histogram` circuit with the given
         length and chunk_length.
         """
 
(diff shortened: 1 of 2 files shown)


wangshan (approved):

poc/vdaf_prio3.py
```diff
@@ -483,10 +483,34 @@ class Prio3HistogramWithLength(Prio3Histogram):
         return Prio3HistogramWithLength
 
 
+class Prio3MultiHotHistogram(Prio3):
+    # Generic types required by `Prio3`
+    Xof = xof.XofShake128
+
+    # Associated parameters.
+    VERIFY_KEY_SIZE = xof.XofShake128.SEED_SIZE
+    ID = 0x00000FFF  # private codepoint for Prio3 types
```
cjpatton, line 492: This codepoint space is intended for Prio3 codepoints that are standardized. Let's leave it out until we've decided to do so. (If it turns out that we do want this for the DP draft, and the DP draft gets adopted, then we will certainly add it to the spec!)
```suggestion
    ID = 0xFFFFFFFF  # private codepoint for Prio3 types
```

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
         return ['length', 'chunk_length']
 
 
+class MultiHotHistogram(Valid):
+    """
+    A validity circuit that checks each Client's measurement is a bit vector,
+    and has at most `max_num_bits` 1s.
```
cjpatton, line 647:
```suggestion
    A validity circuit that checks each Client's measurement is a bit vector
    with at most `max_num_bits` 1s.
```

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
...
+
+    In order to check the Client measurement `x` has at most `max_num_bits` 1s,
```
cjpatton, line 649: To match `valid()`
```suggestion
    In order to check the Client measurement `meas` has at most `max_num_bits` 1s,
```

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
...
+    we ask the Client to send an additional `num_bits_for_count` bits that
```
cjpatton, line 650: Consistency with `SumVec` and `Sum`
```suggestion
    we ask the Client to send an additional `bits_for_count` bits that
```

poc/flp_generic.py
cjpatton, line 649: Self-consistency: We call this the "count check" in the circuit.
```suggestion
    In order to check the Client measurement `x` has at most `max_count` 1s,
```

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
...
+      bits.
+    - We can naturally bound `count` as the following:
+      `0 <= count <= 2**num_bits_for_count - 1`, and therefore:
+      `-offset <= \sum_i x_i <= max_num_bits`.
+    - Since we also verify each `x_i` is a bit, we can lower bound the summation
+      by 0. Therefore, we will be able to verify
+      `0 <= \sum_i x_i <= max_num_bits`.
+    """
+    # Operational parameters
+    length = None  # Set by `MultiHotHistogram.with_params()`
```
cjpatton, line 664: Here and below.
```suggestion
    length = None  # Set by the constructor
```

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
...
+    max_num_bits = None  # Set by `MultiHotHistogram.with_params()`
+    chunk_length = None  # Set by `MultiHotHistogram.with_params()`
+
+    # Associated types
+    Measurement = Unsigned
```
cjpatton, line 669: What is this type? For `Histogram` it's a single bucket index, but for `MultiHotHistogram` we need to be able to represent multiple locations being set.

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
...
+            (length + self.num_bits_for_count + chunk_length - 1) //
+            chunk_length
+        ]
+        self.MEAS_LEN = self.length + self.num_bits_for_count
+        self.OUTPUT_LEN = self.length
+
+    def eval(self, meas, joint_rand, num_shares):
+        self.check_valid_eval(meas, joint_rand)
+
+        # Check that each bucket is one or zero.
```
cjpatton, line 720: Is this sufficient for the DP use case? Suppose our algorithm for adding noise is something like this:
```python3
for bucket_index in len(meas):
   meas[bucket_index] += sample_independent_noise()
```
Depending on what `sample_independent_noise()` does, `meas[bucket_index]` might be something other than `0` or `1`.

poc/flp_generic.py
```diff
@@ -640,6 +641,140 @@ def test_vec_set_type_param(self, test_vec):
...
+                    meas_elem = self.Field(0)
+
+                inputs[j * 2] = r_power * meas_elem
+                inputs[j * 2 + 1] = meas_elem - shares_inv
+
+                r_power *= r
+
+            range_check += self.GADGETS[0].eval(self.Field, inputs)
+
+        # Check that `offset + \sum_{i=0}^{length-1} meas[i]` is equal to the
```
cjpatton, line 741: Prefer simpler language in comments
```suggestion
        # Check that `offset` plus the some of the buckets is equal to the
```

cjpatton (commented):
Great work! One clarifying question about the requirements, see inline comment.

poc/flp_generic.py
junyechen1996: Fixed at all sites for `x`.

poc/flp_generic.py
junyechen1996: Changed it at all locations.

junyechen1996 (commented):
@cjpatton Please also feel free to add David to review, since it's his idea. I don't have permission to add him.

poc/flp_generic.py
cjpatton: Perhaps it should be a set of bucket indices, i.e., `set[Unsigned]`? That would line up better with the measurement type for `Histogram`.

poc/flp_generic.py
cjpatton: Great, thanks for clarifying.

poc/flp_generic.py
```diff
@@ -640,6 +641,137 @@ def test_vec_set_type_param(self, test_vec):
...
+      summation by 0. Therefore, we will be able to verify
+      `0 <= \sum_i meas_i <= max_count`.
+    """
+    # Operational parameters
+    length = None  # Set by the constructor
+    max_count = None  # Set by constructor
+    chunk_length = None  # Set by constructor
+
+    # Associated types
+    Measurement = Vec[Unsigned]
```
cjpatton, line 669: Idea: the measurement is the set of bucket indices.
```suggestion
    Measurement = set[Unsigned]
```

poc/flp_generic.py
```diff
@@ -640,6 +641,137 @@ def test_vec_set_type_param(self, test_vec):
...
+        # Check that `offset` plus the sum of the buckets is equal to the
+        # value claimed by the Client.
+        count_check = self.offset * shares_inv
+        for i in range(self.length):
+            count_check += meas[i]
+        count_check -= self.Field.decode_from_bit_vector(meas[self.length:])
+
+        out = joint_rand[1] * range_check + \
+            joint_rand[1] ** 2 * count_check
+        return out
```
cjpatton, line 747:
```suggestion
        return joint_rand[1]    * range_check + \
               joint_rand[1]**2 * count_check
```

cjpatton (commented):
One minor suggestion for the measurement type.

poc/flp_generic.py
```diff
@@ -640,6 +641,137 @@ def test_vec_set_type_param(self, test_vec):
...
+        """
+        if length <= 0:
+            raise ValueError('invalid length')
+        if max_count <= 0 or max_count > length:
+            raise ValueError('invalid max_count')
+        if chunk_length <= 0:
+            raise ValueError('invalid chunk_length')
+
+        # Compute the number of bits to represent `max_count`.
+        self.bits_for_count = int(math.log2(max_count)) + 1
```
divergentdave, line 693: We may as well avoid floating point math when computing this, so we don't have to worry about precision and rounding.
```suggestion
        self.bits_for_count = max_count.bit_length()
```

poc/flp_generic.py
divergentdave: We have `autopep8` in our linter CI job, and it won't like the irregular number of spaces around operators. We could add some extra parentheses if readability is of concern.

divergentdave (approved):

poc/flp_generic.py
cjpatton: :(((

No no, let's do what the robot tells us to do. The main thing is to be consistent I think.

junyechen1996 (commented):
Rebased on main. Also squashed commits because I have to resolve conflicts.

poc/flp_generic.py
cjpatton: I don't have a strong preference.
* Even if you go with `set`, I don't think there's a need to change the API in the DP draft to match: just map one type to the other.
* If you stick with what you have please (1) change `Vec` to `list` and (2) add a comment describing the measurement type.

junyechen1996: Sounds good. I changed it to `list[Unsigned]` and added comments next to `Measurement` type and `AggResult` type.

cjpatton (approved):

junyechen1996:
Squashed.

PR #303, opened 2023-10-19
Reduce number of AES calls needed by using the LSB of s for t in IdpfPoplar.extend

schoppmp:


divergentdave (approved):
Looks reasonable to me. FWIW, note that the seed correction word will always have its low bit cleared.

cjpatton (approved):
Great idea 👍

PR #308, opened 2023-11-09
Replace SHAKE128 with TurboSHAKE128

### LINKED ISSUE #299: Replace SHAKE128 with TurboSHAKE128
As discussed on the CFRG mailing list: https://mailarchive.ietf.org/arch/msg/cfrg/zzpG6CWQHG2hvQuzmfmkZcZ2AzM/

cjpatton:
Closes #299.

The reference code uses the reference implementation of TurboSHAKE128. This code is unoptimized, so care is needed to ensure our tests run in a reasonable amount of time.

Each time `XofTurboShake128` is constructed we call `TurboSHAKE128()` once and fill a buffer with the output stream. The size of the buffer is a constant, `MAX_XOF_OUT_STREAM_BYTES`, chosen to be sufficiently long for every test that we have. So that we don't have to make this value too large, some of tests in `vdaf_poplar1.py` have been modified.

### DIFF
--- a/poc/xof.py
+++ b/poc/xof.py
@@ -2,12 +2,23 @@
 
 from __future__ import annotations
 
+import os
+import sys
+
 from Cryptodome.Cipher import AES
-from Cryptodome.Hash import SHAKE128
+
+assert os.path.isdir('draft-irtf-cfrg-kangarootwelve/py')  # nopep8
+sys.path.append('draft-irtf-cfrg-kangarootwelve/py')  # nopep8
+from TurboSHAKE import TurboSHAKE128
 
 from common import (TEST_VECTOR, VERSION, Bytes, Unsigned, concat, format_dst,
                     from_le_bytes, gen_rand, next_power_of_2,
                     print_wrapped_line, to_le_bytes, xor)
+
+# Maximum XOF output length that will be requested by any test in this package.
+# Each time `XofTurboShake128` is constructed we call `TurboSHAKE128()` once
+# and fill a buffer with the output stream.
+MAX_XOF_OUT_STREAM_BYTES = 2000
 
 
 class Xof:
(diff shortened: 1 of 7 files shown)


poc/xof.py
```diff
@@ -2,13 +2,24 @@
 
 from __future__ import annotations
 
+import os
+import sys
+
 from Cryptodome.Cipher import AES
-from Cryptodome.Hash import SHAKE128
+
+assert os.path.isdir('draft-irtf-cfrg-kangarootwelve/py')  # nopep8
```
cjpatton, line 10: CI is failing here, apparently because it's not fetching the `draft-irtf-cfrg-kangarootwelve` submodule. How do we tell CI to fetch this?

Note that I've included this implementation because there appears to be no implementation of TurboSHAKE available in pip. As a last resort, we could just copy-paste the reference code for TurboSHAKE here.

divergentdave: This is controlled by the default settings of the `actions/checkout@v3` reusable action. We just need the following in the test job:

```yaml
- name: Checkout
  uses: actions/checkout@v3
  with:
    submodules: true
```

There is [`Crypto.Hash.KangarooTwelve.K12_XOF`](https://pycryptodome.readthedocs.io/en/latest/src/hash/k12.html#Crypto.Hash.KangarooTwelve.K12_XOF) in PyCryptodome, but if it's using a different name, I wonder if the algorithm might be from an old enough draft version that the algorithm is different.

cjpatton: I played around with this a bit and I don't think it complies with the draft. (See https://github.com/Legrandin/pycryptodome/issues/780.) Hint: This takes in an arbitrary-length customization string, whereas the draft wants just a byte for customization.

cjpatton: What do you mean by "test job"? I'm not well versed on GitHub CI and am not sure where to put this action.

cjpatton: That fixed CI! Thanks @divergentdave. I'll leave the TurboSHAKE128 implementation where it is ... it would be great to use PyCryptodome if/when it is compatible with the draft.

divergentdave:
FWIW, Datatracker is returning a quoted-printable encoding of "Benoît Viguier", so xml2rfc's reference cache sees `<author initials="" surname="=?utf-8?q?Beno=C3=AEt_Viguier?=" fullname="=?utf-8?q?Beno=C3=AEt_Viguier?=">`. (Both the bibtex and bibxml files have the quoted-printable version) The encoded strings unfortunately show up in the [rendered document](https://cfrg.github.io/draft-irtf-cfrg-vdaf/cjpatton/299/draft-irtf-cfrg-vdaf.html#TurboSHAKE). This is showing up on the [KangarooTwelve Datatracker page](https://datatracker.ietf.org/doc/draft-irtf-cfrg-kangarootwelve/) as well. The [upstream source](https://github.com/cfrg/draft-irtf-cfrg-kangarootwelve/blame/master/draft-irtf-cfrg-kangarootwelve.xml) looks reasonable, as it's using an XML character entry reference. There's an issue filed about this particular issue already, see ietf-tools/datatracker#6455.

divergentdave (approved):

PR #309, opened 2023-11-10
Prio3: Improve domain separation for multi-proof mode

cjpatton:
Partially addresses #177.

The query randomness is correlated for different values of `Prio3.PROOFS`. To provide defense in depth in case the same verification key is used for different versions, prepend the binder with `bytes(Prio3.PROOFS)`.

For consistency, do the same for the prove randomness and each Helper's share of the proofs. Also, do the same for joint randomness in case `Prio3.PROOFS == 1`.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2608,7 +2608,7 @@ def helper_proofs_share(Prio3, agg_id, k_share):
       Prio3.Flp.Field,
       k_share,
       Prio3.domain_separation_tag(USAGE_PROOF_SHARE),
-      byte(agg_id),
+      byte(Prio3.PROOFS) + byte(agg_id),
       Prio3.Flp.PROOF_LEN * Prio3.PROOFS,
   )
 
@@ -2624,7 +2624,7 @@ def prove_rands(Prio3, k_prove):
       Prio3.Flp.Field,
       k_prove,
       Prio3.domain_separation_tag(USAGE_PROVE_RANDOMNESS),
-      b'',
+      byte(Prio3.PROOFS),
       Prio3.Flp.PROVE_RAND_LEN * Prio3.PROOFS,
   )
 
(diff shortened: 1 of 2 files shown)


cjpatton:
@divergentdave please let me know if you'd be happy implementing multi-proof mode in libprio-rs. (Previously it was optional, but this PR would make this feature wire breaking.)

@albertpl please review as well and make sure you're happy.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2608,7 +2608,7 @@ def helper_proofs_share(Prio3, agg_id, k_share):
       Prio3.Flp.Field,
       k_share,
       Prio3.domain_separation_tag(USAGE_PROOF_SHARE),
-      byte(agg_id),
+      byte(Prio3.PROOFS) + byte(agg_id),
```
albertpl, line 2611: Note to myself. `byte` actually returns `bytes`, which is a sequence. So `+` is concatenation operation.

albertpl (commented):
Good catch!

So this change would break all the randomness related test vectors , right? If we generate test vectors with `PROOFS=1`, it wouldn't be the same again. Is that a problem? Shall we document this? 

Wait. So `wire change` is for this purpose?

cjpatton:
> Good catch!
> 
> So this change would break some of the Prio3 related test vectors , right? If we generate test vectors with `PROOFS=1`, it wouldn't be the same again. Is that a problem? Shall we document this?

Yes, we'll break the test vectors, but we will break the test vectors anyway in the next version (when we bump `VERSION`). All we need to do is re-generate them, so this is not a problem. The break will be documented in the Change Log section of the draft.

> Wait. So `wire change` is for this purpose?

Yes.

albertpl (approved):
Thank you for catching this.

divergentdave (approved):
LGTM. We can implement it in libprio-rs too. I figure the main thing we'll have to watch out for is performance regressions if bound checks stop being elided, but we can address that with appropriate use of iterators.

PR #313, opened 2023-11-16
Prepare VDAF-08

cjpatton:
* Update change log
* Bump `VERSION`
* Generate test vectors


schoppmp (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -354,6 +354,23 @@ security considerations for VDAFs.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+08:
+
+* Poplar1: Bind the report nonce to the authenticator vector programmed into
+  the IDPF. (\*)
```
divergentdave, line 360: This is for my own edification, not a request to change the wording: since k_shard is randomly generated, and not used anywhere else, this binding isn't checked/enforced by any other protocol participant. While this isn't preventing any replay attack, I can see that it could simplify security proofs, and maybe improve bounds, by eliminating the impact of k_shard colliding. Is it typical to still refer to incorporating identifiers as "binding" in situations like this?

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: "Binding" is a very abused term. I'm using the phrase "bind to nonce" here to mean that the value is derived from the nonce: if the changes, then so does the value. I don't mean to imply any particular security property. In certain situations, this sort of thing may indeed be necessary for security.

PR #315, opened 2023-11-28
Remove test vectors for non-standard Prio3 variants

cjpatton:
The reference code has implementations of Prio3 variants that have not been standardized. The test code generates test vectors for these, which we checked in to the repo when we cut the last draft.

To avoid confusion, remove the test vectors for these non-standardized variants.


divergentdave:
One related note: when `test_prio3sumvec_with_multiproof()` calls `_test_prio3sumvec()` three times, this results in clobbering the test vector file twice. Before we eventually standardize this, we'll want to pass different numbers for `test_vec_instance` to fix this.

divergentdave (approved):
Looks good. Should we take out the `print_test_vec=TEST_VECTOR` arguments that generated these as well?

cjpatton:
> Looks good. Should we take out the `print_test_vec=TEST_VECTOR` arguments that generated these as well?

Done.

Issue #311, opened 2023-11-16
Prio3: Add guidance for choosing `PROOFS` and `Flp.Field` 


cjpatton:
If a small field is used, then it may be necessary to use a larger number of proofs in order to achieve the desired level of robustness. (Privacy is not significantly impacted by the choice of these parameters.) The draft needs to provide some guidance for choosing this.

We may also consider re-parameterizing the existing Prio3 variants.

See #177 for initial discussion and analysis.

cc/ @albertpl

cjpatton:
I've written a sage script (#318) for plotting robustness of Prio3SumVec for various choices of field and number of proofs. The generated plot is below. What it shows is an upper bound of the probability that $1$ in $1,000,000,000$ accepted reports is invalid, given a amount of computational resources ($2^{80}$ random oracle queries, which is a pretty comfortable margin). The x-axis shows the input length; the number of bits is $1$; and the chunk length is set to the asymptotic optimal (square root of the length).

![prio3_sum_vec](https://github.com/cfrg/draft-irtf-cfrg-vdaf/assets/3453007/49442bc9-91c8-4c87-9203-72b5df267d7b)

* The baseline is Field128/1 (one proof). The probability of a robustness error is definitely non-negligible (where "negligible" is approximately the probability of correctly guessing an AES key), but it's still quite unlikely to happen. For instance, for input size = $100,000$, the probability that $1$ in every $1$ billion accepted reports is invalid is at most $2^{-36}$. This is the same as the probability of ruining $1$ in every $100,000$ batches assuming a batch size of $10,000$.
* Field64/1 is quite terrible and should not be deployed.
* Field64/2 is significantly worse than Field128/1, but may be tolerable depending on the application.
* Field64/3 is comfortably better than Field128/1.
* Field64/4 is much more robust than we'd ever need.

Based on this analysis, I'd suggest the following text change. Top of Section 7.4:

> For circuits for which the number of gadget calls grows sub-linearily in the parameters (e.g., `Sum`, `SumVec`, or `Histogram`), a smaller field can be used as long as additional proofs are generated and verified. When replacing `Field128` with `Field64` it is RECOMMENDED to use `3` proofs. However, some applications may tolerate as few as `2`.

albertpl:
This assumes the adversary donates 1000000000 reports? either control the number of devices or break authentication / rate limit.

cjpatton:
[...]

PR #318, opened 2023-12-05
Add a tool for plotting Prio3 robustness bounds

cjpatton:
Partially addresses #311.

### DIFF
--- a/poc/plot_prio3_multiproof_robustness.py
+++ b/poc/plot_prio3_multiproof_robustness.py
@@ -1,0 +1,101 @@
+# prio3_multiproof_robustness.py - Plot robustness bounds for various parameters.
+# Use `sage -python prio3_multiproof_robustness.py`
+import math
+
+import matplotlib.pyplot as plt
+
+from field import Field64, Field128
+from vdaf_prio3 import Prio3SumVec
+
+NUM_REPORTS = 1000000000
+
+
+def soundness(gadget_calls, gadget_degree, field_size):
+    '''
+    ia.cr/2019/188, Theorem 4.3
+
+    gadget_calls - number of times the gadget is called
+
+    gadget_degree - arithmetic degree of the gadget
+
+    field_size - size of the field
+    '''
+    return gadget_calls * gadget_degree / (field_size - gadget_calls)
+
+
+def robustness(epsilon, ro_queries, prep_queries, num_proofs, seed_bits):
+    '''
+    ia.cr/2023/130, Theorem 1, assuming the bound can be modified by raising
+    `epsilon` to the power of the number of FLPs.
+
+    epsilon - soundness of the base FLP
+
+    ro_queries - random oracle queries, a proxy for the amount of precomputation
+                 done by the adversary
+
+    prep_queries - number of online attempts, a proxy for the batch size
+
+    num_proofs - number of FLPs
+
+    seed_bits - the size of the XOF seed in bits
+    '''
+    return (ro_queries + prep_queries) * epsilon**num_proofs + \
+           (ro_queries + prep_queries**2) / 2**(seed_bits - 1)
+
+
+def sum_vec(field_size, num_proofs, length):
+    '''
+    Prio3SumVec (draft-irtf-cfrg-vdaf-08, Section 7.4.3): Probability of
+    accepting one report in a batch of NUM_REPORTS. Assuming the asymptotically
+    optimal chunk length.
+    '''
+    bits = 1
+    chunk_length = max(1, length**(1/2))
+    vdaf = Prio3SumVec.with_params(length, bits, chunk_length)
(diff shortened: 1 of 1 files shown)


misc/prio3_multiproof_robustness.py
```diff
@@ -0,0 +1,100 @@
...
+    gadget_degree - arithmetic degree of the gadget
+
+    field_size - size of the field
+    '''
+    return gadget_calls * gadget_degree / (field_size - gadget_calls)
+
+
+def robustness(epsilon, ro_queries, prep_queries, num_proofs):
+    '''
+    ia.cr/2023/130, Theorem 1, ignoring negligible terms and assuming the bound
```
albertpl, line 27: btw, I believe when `num_proofs` increase,  these negligible may become more significant since the first term is exponentially shrinking. Any harm to include it anyway?

misc/prio3_multiproof_robustness.py
```diff
@@ -0,0 +1,100 @@
...
+    '''
+    ia.cr/2019/188, Theorem 4.3
+
+    gadget_calls - number of times the gadget is called
+
+    gadget_degree - arithmetic degree of the gadget
+
+    field_size - size of the field
+    '''
+    return gadget_calls * gadget_degree / (field_size - gadget_calls)
```
junyechen1996, line 22: Don't you also need to account for the soundness error of taking random linear combination, when one of the coefficients multiplied with x^2-x circuits is 0? proved by Theorem 5.3 in the same paper.

misc/prio3_multiproof_robustness.py
```diff
@@ -0,0 +1,100 @@
...
+    optimal chunk length.
+    '''
+
+    # Table 11
+    def gadget_calls(length, bits, chunk_length):
+        return (length * bits + chunk_length - 1) // chunk_length
+
+    return robustness(
+        soundness(
+            gadget_calls(length, 1, max(1, math.sqrt(length))),
```
junyechen1996, line 55: nit: `math.sqrt` will produce a float, better to use round down or up? since `chunk_length` should be an integer.

misc/prio3_multiproof_robustness.py
cjpatton: Yes, good catch.

misc/prio3_multiproof_robustness.py
cjpatton: Sure, I can add it.

misc/prio3_multiproof_robustness.py
albertpl: That is excellent point. And this new term is probably the dominant one.

poc/plot_prio3_multiproof_robustness.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def sum_vec(field_size, num_proofs, length):
+    '''
+    Prio3SumVec (draft-irtf-cfrg-vdaf-08, Section 7.4.3): Probability of
+    accepting one report in a batch of NUM_REPORTS. Assuming the asymptotically
+    optimal chunk length.
+    '''
+    chunk_length = max(1, int(math.sqrt(length)))
+    vdaf = Prio3SumVec.with_params(length, 1, chunk_length)
```
junyechen1996, line 52: I know we are assuming `bit_length = 1` in this function, but in case you want to extend this function in the future, it might help to make this explicit (e.g. defining this variable, or set that as one of the function arguments), since you also need `bit_length` when computing `circuit_soundness = length * bit_length / field_size`, so we don't forget about this assumption.

albertpl (approved):

poc/plot_prio3_multiproof_robustness.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def sum_vec(field_size, num_proofs, length):
+    '''
+    Prio3SumVec (draft-irtf-cfrg-vdaf-08, Section 7.4.3): Probability of
+    accepting one report in a batch of NUM_REPORTS. Assuming the asymptotically
+    optimal chunk length.
+    '''
+    bits = 1
+    chunk_length = max(1, length**(-1/2))
```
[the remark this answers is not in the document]
cjpatton: Whoops, the negative sign isn't supposed to be there.
The `int()` isn't necessary because sage represents numbers exactly:
```
[cjpatton/311][~/github.com/cfrg/draft-irtf-cfrg-vdaf/poc]$ sage
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 9.8, Release Date: 2023-02-11                     │
│ Using Python 3.11.1. Type "help()" for help.                       │
└────────────────────────────────────────────────────────────────────┘
sage: 100^(1/2)
10
sage: 102^(1/2)
sqrt(102)
sage:
```

PR #320, opened 2023-12-17
Infer relative path of kangerootwelve

junyechen1996:
This will resolve two problems:
- When another repo tries to submodule this repo, we won't hit problems when trying to import kangerootwelve
- We don't always have to run unit tests from the "poc" folder, e.g., I can still run unit tests from the root directory of this repo, sage -python poc/vdaf_prio3.py


cjpatton (approved):
👍

PR #322, opened 2023-12-20
poc: Reduce test runtime by making TurboSHAKE128 stateful

### LINKED ISSUE #321: poc: TurboSHAKE128: Replace reference implementation with a stateful version
We currently impose a limit on the number of bytes output by `XofTurboShake128` due to the inefficiency of the current implementation. Replace the stateless implementation with a stateful one and remove the limit.

cjpatton:
Closes #321.

The reference implementation of TurboSHAKE128 provides a one-shot API. To use this efficiently, we restrict the number of bytes required for tests so that we can pre-compute the entire output buffer. This is controlled by MAX_XOF_OUT_STREAM_BYTES.

This restriction is confusing and has side-effects for specs that use XofTurboShake128. However using the one-shot API without this optimization makes the unit tests prohibitively slow, about a minute on my machine.

Implement a stateful API for TurboSHAKE128 and use it in XofTurboShake128. This reduces the runtime to 20 seconds on my machine.

Accordingly, restore the unit tests for Poplar1 to what they were before we made this optimization.

cc/ @junyechen1996

### DIFF
--- a/poc/Makefile
+++ b/poc/Makefile
@@ -2,6 +2,7 @@ test:
 	sage -python common.py
 	sage -python field.py
 	sage -python xof.py
+	sage -python turboshake.py
 	sage -python flp.py
 	sage -python flp_generic.py
 	sage -python idpf.py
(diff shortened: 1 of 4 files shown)


cjpatton:
Reviewer note: Eventually we'll be able to use PyCryptodome instead: https://github.com/Legrandin/pycryptodome/issues/780

PR #324, opened 2024-01-11
Use PyCryptodomex's implementation of TurboSHAKE128

### LINKED ISSUE #323: poc: Use PyCryptodome for TurboSHAKE
As of this writing, PyCryptodome does not support TurboSHAKE. However the next release will: https://github.com/Legrandin/pycryptodome/issues/780

In the meantime, we are using our own implementation: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/322

Once a PyCryptodome is released with support TurboSHAKE, use it instead.

cjpatton:
Closes #323.

Support for TurboSHAKE was recently (as of version 3.20.0) added to PyCryptodome. Use it instead of our own implementation and remove our implementation.

Accordingly, remove the draft-irtf-cfrg-kangarootwelve submodule, as we no longer need it for interop testing with our code.

Note that PyCryptodome can be upgraded with:

$ sage -pip install --upgrade pycryptodome


.github/workflows/test.yml
```diff
@@ -25,7 +25,8 @@ jobs:
     - name: Install Sage
       run: |
         sudo apt-get update
-        sudo apt-get install -y sagemath python3-pycryptodome python3-cffi
+        sudo apt-get install -y sagemath python3-cffi
+        sage -pip install pycryptodome
```
divergentdave, line 29: This should install `pycryptodomex` instead, per `poc/README.md`. The two PyPI packages get imported differently.

cjpatton: What about just importing AES from Crypto instead of Cryptodome. I've updated the README to install pycryptodome instead of pycryptodomex.

Does that work for you, or might there be an enviernment where this doesn't work? CI seems happy, as well as my Mac.

divergentdave: Yeah, this should be fine, so long as we're consistent. This would preclude someone from installing PyCrypto in the same Sage install, but hopefully that's not a common occurrence.

divergentdave (approved):

.github/workflows/test.yml
cjpatton: Hmm actually that's a good call. In fact I think that's what we intended when we decided to use PyCryptodomex in the first place.

cjpatton: Replaced Crypto with Cryptodome.

divergentdave (approved):
LGTM, test vectors are unchanged to boot.

PR #325, opened 2024-01-11
Add guidance for choosing the field size

### LINKED ISSUE #311: Prio3: Add guidance for choosing `PROOFS` and `Flp.Field`

cjpatton:
Closes #311.

For Prio3, when joint randomness is used:

* RECOMMEND Field128
* Field64 MAY be used, but `PROOFS` MUST be `3`

The latter is motivated by issue #177. The recommendation is based on the upper bound given by {{DPRS23}}, Theorem 1 and a matching attack described by @bwesterb (see issue #311).

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2134,7 +2134,15 @@ def run_flp(flp, meas: Vec[Flp.Field], num_shares: Unsigned):
 The proof system is constructed so that, if `meas` is valid, then `run_flp(Flp,
 meas, 1)` always returns `True`. On the other hand, if `meas` is invalid, then
 as long as `joint_rand` and `query_rand` are generated uniform randomly, the
-output is `False` with overwhelming probability.
+output is `False` with high probability. False positives are possible: there is
+a small probability that a verifier accepts an invalid input as valid. An FLP
+is said to be "sound" if this probability is sufficiently small. The soundness
+of the FLP depends on a variety of parameters, like the length of the
+input and the size of the field. See {{flp-generic}} for details.
+
+Note that soundness of an FLP system is not the same as robustness for a VDAF
+In particular, soundness of the FLP is necessary, but insufficient for
+robusntess of Prio3 ({{prio3}}). See {{security-multiproof}} for details.
 
 We remark that {{BBCGGI19}} defines a much larger class of fully linear proof
 systems than we consider here. In particular, what is called an "FLP" here is
@@ -2176,21 +2184,25 @@ post-processing.
 We remark that, taken together, these three functionalities correspond roughly
 to the notion of "Affine-aggregatable encodings (AFEs)" from {{CGB17}}.
 
-### Multiple proofs {#multiproofs}
+### Multiple Proofs {#multiproofs}
 
-To improve soundness, the prover can construct multiple unique proofs for
-its measurement such that the verifier will only accept the measurement once
-all proofs have been verified. Notably, several proofs using a smaller
-field can offer the same level of soundness as a single proof using a large field.
+It is sometimes desirable to generate and verify multiple independent proofs
+for the same input. First, this improves the soundness of the proof system
+without having to change any of its parameters. Second, it allows a smaller
+field to be used (e.g., replace Field128 with Field64, see {{flp-generic}})
+without sacrificing soundness. Generally, choosing a smaller field can
+significantly reduce communication cost. (This is a trade-off, of course, since
+generating and verifying more proofs requires more time.) Given these benefits,
+this feature is implemented by Prio3 ({{prio3}}).
 
 To generate these proofs for a specific measurement, the prover calls
 `Flp.prove` multiple times, each time using an independently generated prover
-and joint randomness string. The verifier checks each proof independently,
-each time with an independently generated query randomness string. It accepts
-the measurement only if all the decision algorithm accepts on each proof.
+and joint randomness string. The verifier checks each proof independently, each
+time with an independently generated query randomness string. It accepts the
+measurement only if all the decision algorithm accepts on each proof.
 
 See {{security-multiproof}} below for discussions on choosing the right number
- of proofs.
+of proofs.
 
 ## Construction {#prio3-construction}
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -4634,11 +4642,32 @@ We also stress that even if the Idpf is not extractable, Poplar1 guarantees
...
+{{DPRS23}}, Theorem 1.) Note that privacy is not susceptible to such attacks.
+
+Another way to mitigate this issue (or improve robustness in general) is to
+generate and verify multiple, independent proofs. (See {{multiproofs}}.) For
+Prio3, the `PROOFS` parameter controls the number of proofs (at least one) that
+are generated and verified.
+
+In general, Field128 is RECOMMENDED for use in Prio3 when the circuit uses
+joint randomness (`JOINT_RAND_LEN > 0`) and `PROOFS == 1`. Field64 MAY be used
+instead, but `PROOFS` MUST be set to `3`. Breaking robustness for `PROOFS == 2`
```
bwesterb, line 4668: Perhaps ≥3 instead of =3?

bwesterb: For the rest, lgtm.

draft-irtf-cfrg-vdaf.md
```diff
@@ -2134,7 +2134,11 @@ def run_flp(flp, meas: Vec[Flp.Field], num_shares: Unsigned):
 The proof system is constructed so that, if `meas` is valid, then `run_flp(Flp,
 meas, 1)` always returns `True`. On the other hand, if `meas` is invalid, then
 as long as `joint_rand` and `query_rand` are generated uniform randomly, the
-output is `False` with overwhelming probability.
+output is `False` with high probability. False positives are possible: there is
+a small probability that a verifier accepts an invalid input as valid. An FLP
+is said to be "sound" if this probability is sufficiently small. The soundness
+of the FLP depends on a variety of parameters, like like the length of the
```
wangshan, line 2140:
```suggestion
of the FLP depends on a variety of parameters, like the length of the
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4634,11 +4642,32 @@ We also stress that even if the Idpf is not extractable, Poplar1 guarantees
...
+flexible circuit design (see {{flp-generic-overview-extensions}}), but opens up
+Prio3 to precomputation attacks, which the larger field mitigates. (See
+{{DPRS23}}, Theorem 1.) Note that privacy is not susceptible to such attacks.
+
+Another way to mitigate this issue (or improve robustness in general) is to
+generate and verify multiple, independent proofs. (See {{multiproofs}}.) For
+Prio3, the `PROOFS` parameter controls the number of proofs (at least one) that
+are generated and verified.
+
+In general, Field128 is RECOMMENDED for use in Prio3 when the circuit uses
```
wangshan, line 4666: should we add a section to quantify the risk of using 2 PROOFS vs 3?

wangshan: here soundness is not mentioned, while it was in #multiproofs. Should we mention the relationship between soundness and robustness (which is the overall goal) here?

cjpatton: Can you clarify: 2 proofs for Field128 or Field64?

Connecting soundness and robustness is a good idea, will do.

divergentdave (approved):
Looks good. I think we should also add `| PROOFS | 1 |` to tables 9, 10, 11, and 12, right after `Field`, to be clear about the parameter values for the four concrete Prio3 VDAFs (and algorithm IDs) we define.

junyechen1996 (approved):

draft-irtf-cfrg-vdaf.md
wangshan: 2 proofs for Field64.
There were many discussions in issues for robustness analysis, especially 2 proofs vs 3. Would be good to tell reader what are the risks involved using PROOFS == 2. So we can justify why "`PROOFS` MUST be set to at least `3`", instead of "SHOULD be"

cjpatton: Below we say "breaking robustness for `PROOFS == 2` is feasible, if impractical".

I mean to use the terms "feasible" and "impractical" the way they are generally meant in the crypto literature:  An attack is "feasilbe' if it can be pulled off, given enough resources. However the amount of resources it takes to pull off means we may not see the attack happen in practice (costs too much money, too much memory, or too much time).

wangshan (approved):
Approved assuming "Connecting soundness and robustness" is addressed either in this PR or a future one.

cjpatton:
> Approved assuming "Connecting soundness and robustness" is addressed either in this PR or a future one.

Right, thanks for the reminder! I'll ping you for a quick review if you don't mind.

draft-irtf-cfrg-vdaf.md
cjpatton: Added note about robustness versus soundness.

cjpatton:
> @divergentdave  Looks good. I think we should also add `| PROOFS | 1 |` to tables 9, 10, 11, and 12, right after `Field`, to be clear about the parameter values for the four concrete Prio3 VDAFs (and algorithm IDs) we define.

These tables specify parameters of the circuit; the number of proofs is not a parameter of the circuit. In fact, the gudance here is about modifying one of the circuits, say `SumVec` by changing the field from `Field64` to `Field128`.

What do you think about decoupling `Field` from the circuit and moving it to the `Prio3` variant. There, `PROOFS` is (or should, at least) be specified.

divergentdave:
Good point, I think that sounds better

cjpatton:
I'll lift to an issue and handle it after this PR.

Issue #327, opened 2024-01-23
Testing improvements


cjpatton:
1. Adopt `unittest` for unit tests.
2. Decouple test vector generation from unit tests. Requirements:
     * We would like the code to set the path for the test vector (e.g., `test_vec/08`).
     * Unit tests should always generate fresh randomness; for test vectors, either: (1) use the a fixed sequence of coins, as the draft suggests; or (2) publish the generated randomness with test vectors.
     * We need a separate function for generating test vectors and running the VDAF.

cc/ @junyechen1996

cjpatton:
Another issue here: Some tests, when they fail, merely dump something to stdout without triggering unittest to have a fatal exit status. This means these failures won't be noticed in CI.

PR #330, opened 2024-02-29
Use unittest for tests

divergentdave:
This moves tests from `if __name__ == "__main__":` blocks in each file to a separate directory, using the standard library's `unittest` framework. Aside from better code organization, this also speeds up test execution, because we only need to load Sage's libraries once. This is part of #327.


poc/flp_generic.py
```diff
@@ -647,7 +647,7 @@ def test_vec_set_type_param(self, test_vec):
 
 
 class MultiHotHistogram(Valid):
-    """
+    r"""
```
cjpatton, line 650: Why make this change?

poc/tests/__init__.py
cjpatton, line 1: Why is this file needed?

cjpatton (approved):
Beautiful, thanks so much for putting the time in on this.

poc/flp_generic.py
divergentdave: I saw a message after I added an import for this file, I think it was a deprecation warning. The issue is that `\s` is not a supported escape character sequence. Since backslashes are only used in LaTeX syntax here, I just turned the docstring into a raw string literal to fix it.

poc/tests/__init__.py
divergentdave: This indicates that Python should treat the folder hierarchy as a package. The unittest auto-discovery code needs to see this, and it's also required to enable intra-package imports.

PR #331, opened 2024-02-29
Use Sage container image in CI

divergentdave:
This uses a Sage Docker image in lieu of installing Sage with apt-get. This should hopefully speed up CI significantly. #330 enables this because our Makefile is now only two lines long, so we can duplicate the command in it, and use this container image that lacks GNU Make.


cjpatton (approved):

divergentdave:
Now twice as fast!

PR #332, opened 2024-03-07
Add checks that prefix counts are consistent across multiple VDAF executions

### LINKED ISSUE #316: Clarify that Poplar-like protocols may leak sensitive information if the collector misbehaves
The collector decides which parts of the IDPF tree to explore. This not only leaks prefixes of the heavy hitter values but, if the collector decides to explore other branches, possibly the values of individual clients.

This is pointed out in BBCGGI21:

> A malicious server’s only strategy to learn extra information
in Protocol 5 is to manipulate answers to the prefix-count oracle
queries using an “additive attack.” [...] Intuitively:
the adversary can essentially control which strings are heavy
hitters (and can thus learn how many honest clients hold strings
in a small set), [..]

In the current DAP implementation this is attack is trivial to execute for the collector since it even sees the values in plain. This risk should be clearly pointed out in the standard.

schoppmp:
Fixes #316


draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,41 @@ parameter with the same level has been used with the same input share before.
...
+
+    # Check that level increased by one, and the threshold is the same
+    if level != last_level + 1:
+      return False
+    if threshold != last_threshold:
+      return False
+
+    # Check that prefixes are suffixes of the last level's prefixes,
+    # and that the counts of siblings sum up to at most the parent's count.
+    for i in range(len(prefixes))
```
cjpatton, line 4004:
```suggestion
    for i in range(len(prefixes)):
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,41 @@ parameter with the same level has been used with the same input share before.
...
+      if not parent_index.contains(prefixes[i] >> 1):
```
cjpatton, line 4005: `parent_index` is not defined.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,41 @@ parameter with the same level has been used with the same input share before.
...
+        return False
+
+    # Check that level increased by one, and the threshold is the same
+    if level != last_level + 1:
+      return False
+    if threshold != last_threshold:
+      return False
+
+    # Check that prefixes are suffixes of the last level's prefixes,
+    # and that the counts of siblings sum up to at most the parent's count.
```
cjpatton, line 4003: This code is hard to follow. Let's try to encapsulate some of the bit fiddling details with high level funtions, like `is_parent_of(parent, child)`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4603,8 +4633,15 @@ perhaps unlikely) for a large set of non-heavy-hitter values to share a common
 prefix, which would be leaked by a prefix tree with a sufficiently small
 threshold.
 
-The only known, general-purpose approach to mitigating this leakage is via
-differential privacy.
+A malicious adversary controlling the Collector and one of the Aggregators can
+further turn arbitrary non-heavy prefixes into heavy ones by tampering with the
+DPF output at any position. While our construction ensures that the sum of all
```
cjpatton, line 4638:
```suggestion
IDPF output at any position. While our construction ensures that the sum of all
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,41 @@ parameter with the same level has been used with the same input share before.
 The whole preparation phase MUST NOT be run more than once for a given
 combination of input share and level.
 
+<!-- TODO: Text about counts -->
```
cjpatton, line 3980: Here we should list the invariants we should expect to hold for each aggregation.

WDYTA enforcing that the sum of the prefix counts is equal to the number of valid reports in the batch?

schoppmp: Added some text.

> WDYTA enforcing that the sum of the prefix counts is equal to the number of valid reports in the batch?

This is enforced in `aggregate`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,41 @@ parameter with the same level has been used with the same input share before.
...
+
+    (level, prefixes, threshold, prefix_counts) = agg_param
+    (last_level, last_prefixes, last_threshold, last_prefix_counts) = previous_agg_params[-1]
+
+    # Check that prefix_count for prefixes are above threshold.
+    for count in prefix_counts:
+      if count < threshold:
+        return False
+
+    # Check that level increased by one, and the threshold is the same
```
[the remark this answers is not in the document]
schoppmp: We could, is this needed somewhere? This would require changing the checks below.

draft-irtf-cfrg-vdaf.md
schoppmp: Fixed.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,41 @@ parameter with the same level has been used with the same input share before.
...
+    if level != last_level + 1:
+      return False
+    if threshold != last_threshold:
```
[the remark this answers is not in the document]

draft-irtf-cfrg-vdaf.md
schoppmp: Added a helper `get_prefix` that contains the bit shift.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3885,15 +3885,15 @@ def prep_init(Poplar1, verify_key, agg_id, agg_param,
         corr_xof = Poplar1.Xof(
             corr_seed,
             Poplar1.domain_separation_tag(USAGE_CORR_INNER),
-            byte(agg_id) + nonce,
+            byte(agg_id) + nonce + to_be_bytes(self.threshold),
```
cjpatton, line 3888: Note that we'll need to make the same change to `shard()`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3952,8 +3952,8 @@ def prep_next(Poplar1, prep_state, prep_msg):
 
 def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
     if len(prep_shares) != 2:
-        raise ERR_INPUT  # unexpected number of prep shares
-    (level, _) = agg_param
+        raise ERR_INPUT # unexpected number of prep shares
```
cjpatton, line 3955: I believe our Python linters like two spaces for in-line comments
```suggestion
        raise ERR_INPUT  # unexpected number of prep shares
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,48 @@ parameter with the same level has been used with the same input share before.
...
+
 ~~~
-def is_valid(agg_param, previous_agg_params):
-    (level, _) = agg_param
-    return all(
-        level != other_level
-        for (other_level, _) in previous_agg_params
-    )
+def get_prefix(input, this_level, last_level):
+    # Helper function to determin the prefix of `input` at `last_level`.
```
cjpatton, line 3986:
```suggestion
    # Helper function to determine the prefix of `input` at `last_level`.
```

draft-irtf-cfrg-vdaf.md
cjpatton, line 3888: `prep_init()` is a classmethod; the class is called `Poplar1` in this scope.

(Sorry, in retrospect, class factories was probably the wrong paradigm for this project. cc/ @divergentdave.)

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,48 @@ parameter with the same level has been used with the same input share before.
...
+counts.
+
 ~~~
-def is_valid(agg_param, previous_agg_params):
-    (level, _) = agg_param
-    return all(
-        level != other_level
-        for (other_level, _) in previous_agg_params
-    )
+def get_prefix(input, this_level, last_level):
```
cjpatton, line 3985: WDYT?
```suggestion
def get_ancestor(input, this_level, last_level):
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3977,13 +3977,48 @@ parameter with the same level has been used with the same input share before.
...
+
+    # Check that prefixes are suffixes of the last level's prefixes,
+    # and that the counts of siblings sum up to at most the ancestor's count.
+    for (i, prefix) in enumerate(prefixes):
+      last_prefix = get_prefix(prefix, level, last_level)
+      if last_prefix not in last_prefix_count_dict:
+        return False  # Current prefix not a suffix of last level's prefixes.
+      if last_prefix_count_dict[last_prefix] < prefix_counts[i]:
+        return False  # Count exceeds last level's count.
+      last_prefix_count_dict[last_prefix] -= prefix_counts[i]
```
cjpatton, line 4019: I think we also need to check that value in this dict is 0?

cjpatton (commented):
This looks to be on the right track, modulo our conversation today about compatibility with DP. I think the only thing that would change is enforcing count consistency, in which case we would take some version of this PR no matter what the outcome is.

I'd suggest implementing in the reference code so that we can unit test is_valid() and make sure the changes to the rest of the construction still pass our current tests. If you like you can leave out the count checks.

draft-irtf-cfrg-vdaf.md
schoppmp: Removed the dict.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4603,8 +4626,15 @@ perhaps unlikely) for a large set of non-heavy-hitter values to share a common
...
-The only known, general-purpose approach to mitigating this leakage is via
-differential privacy.
+A malicious adversary controlling the Collector and one of the Aggregators can
+further turn arbitrary non-heavy prefixes into heavy ones by tampering with the
+DPF output at any position. While our construction ensures that the sum of all
+nodes evaluated at a level cannot exceed the value at the parent nodes, this
+still may allow an adversary to discover individual non-heavy strings.
+
+The only practical, general-purpose approach to mitigating these leakages is via
+differential privacy, which is RECOMMENDED for all protocols using Poplar1 for
```
schoppmp, line 4636: Should this become a MUST?

schoppmp:
Following the discussion in [this doc](https://docs.google.com/document/d/1ZjXz-1kGsTDf2Vn2u-fwYqR8BSc3tOYlELVHAYvAfjk), it seems clear that differential privacy is required to implement heavy hitters privately. For now, this PR only ensures that the tree is traversed correctly (i.e., not on arbitrary leaves; see mitigation 1 [here](https://docs.google.com/document/d/1ZjXz-1kGsTDf2Vn2u-fwYqR8BSc3tOYlELVHAYvAfjk/edit#bookmark=id.yyi83665dnuy)).

The details of how to add differential privacy to a heavy hitters protocol based on Poplar should be specified in the higher-level protocol, as VDAF doesn't have any syntax for DP (like privacy parameters). There are also different ways to implement DP here, so this is not something we want to restrict in VDAF.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3975,15 +3975,36 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+
+def is_valid(Poplar1, agg_param, previous_agg_params):
+    # Exit early if this is the first time.
+    if len(previous_agg_params) < 1:
+      return True
+
+    (level, prefixes) = agg_param
+    (last_level, last_prefixes) = previous_agg_params[-1]
+    # The empty prefix 0 is always there.
+    last_prefixes_set = set(list(last_prefixes) + [0])
```
divergentdave, line 3995: `0` is not the empty prefix here, but the leftmost prefix. The length of all these prefixes is implicitly `last_level`, so there's no way to represent the empty prefix. Rather, bootstrapping prefixes is handled by `if len(previous_agg_params) < 1:` above.
```suggestion
    last_prefixes_set = set(last_prefixes)
```

poc/vdaf_poplar1.py
```diff
@@ -151,16 +151,39 @@ def shard(Poplar1, measurement, nonce, rand):
...
+        the prefixes at each level are suffixes of the previous level's
+        prefixes.
+        """
+        if len(previous_agg_params) < 1:
+            return True
+
+        (level, prefixes) = agg_param
+        (last_level, last_prefixes) = previous_agg_params[-1]
+        # The empty prefix 0 is always there.
+        last_prefixes_set = set(list(last_prefixes) + [0])
```
divergentdave, line 174: See the related suggestion on the specification.
```suggestion
        last_prefixes_set = set(last_prefixes)
```

poc/vdaf_poplar1.py
```diff
@@ -151,16 +151,39 @@ def shard(Poplar1, measurement, nonce, rand):
         input_shares = list(zip(keys, corr_seed, corr_inner, corr_leaf))
         return (public_share, input_shares)
 
-    def is_valid(agg_param, previous_agg_params):
+    @classmethod
+    def get_ancestor(Poplar1, input, this_level, last_level):
```
cjpatton, line 155: This doesn't need to be a class method. How about defining this below the class somewhere, just as a helper function?

draft-irtf-cfrg-vdaf.md
```diff
@@ -4603,8 +4626,15 @@ perhaps unlikely) for a large set of non-heavy-hitter values to share a common
 prefix, which would be leaked by a prefix tree with a sufficiently small
 threshold.
 
-The only known, general-purpose approach to mitigating this leakage is via
-differential privacy.
+A malicious adversary controlling the Collector and one of the Aggregators can
+further turn arbitrary non-heavy prefixes into heavy ones by tampering with the
+DPF output at any position. While our construction ensures that the sum of all
```
cjpatton, line 4631:
```suggestion
IDPF output at any position. While our construction ensures that the sum of all
```

draft-irtf-cfrg-vdaf.md
cjpatton: I don't think we can make it a MUST without specifying a mechanism for achieving it.

cjpatton (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -3995,6 +4016,8 @@ Aggregation involves simply adding up the output shares.
 ~~~
 def aggregate(Poplar1, agg_param, out_shares):
     (level, prefixes) = agg_param
+    if sum(prefix_counts) > len(out_shares):
```
tgeoghegan, line 4019: I don't think I understand what this check does. Where is `prefix_counts` defined?

draft-irtf-cfrg-vdaf.md
tgeoghegan: I don't think that's true. For instance, OHTTP doesn't specify how clients obtain the configuration of gateways, but still has a MUST about it ([source](https://datatracker.ietf.org/doc/html/rfc9458#section-7)):

> A specific method for a Client to acquire configurations is not included in this specification. Applications using this design MUST provide accommodations to mitigate tracking using Client configurations.

So I think we could include a MUST about using DP, since the VDAF's privacy guarantees don't hold unless you do.

cjpatton: I'm fine with it being a MUST.

draft-irtf-cfrg-vdaf.md
cjpatton: Bump

poc/vdaf_poplar1.py
cjpatton: Bump

draft-irtf-cfrg-vdaf.md
schoppmp: I'm leaving it at RECOMMENDED for now, since I'm not sure prescribing DP but not any particular DP parameters would be meaningful. In particular, implementers could satisfy the requirement by implementing DP with $\delta=1$ or similar insecure settings.

draft-irtf-cfrg-vdaf.md
schoppmp: This was from an earlier version, removed it.

poc/vdaf_poplar1.py
schoppmp: Done.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3975,15 +3960,38 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+    return input >> (this_level - last_level)
+
+def is_valid(Poplar1, agg_param, previous_agg_params):
+    # Exit early if this is the first time.
+    if len(previous_agg_params) < 1:
+      return True
+
+    (level, prefixes) = agg_param
+    (last_level, last_prefixes) = previous_agg_params[-1]
+    # The empty prefix 0 is always there.
```
divergentdave, line 3979: This comment is outdated.
```suggestion
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3975,15 +3960,38 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+    last_prefixes_set = set(list(last_prefixes))
```
divergentdave, line 3980: nit: I think the intermediate list is unnecessary, as `set()` can take any iterable
```suggestion
    last_prefixes_set = set(last_prefixes)
```

poc/vdaf_poplar1.py
```diff
@@ -151,16 +151,34 @@ def shard(Poplar1, measurement, nonce, rand):
...
-        return all(
-            level != other_level
-            for (other_level, _) in previous_agg_params
-        )
+        if len(previous_agg_params) < 1:
+            return True
+
+        (level, prefixes) = agg_param
+        (last_level, last_prefixes) = previous_agg_params[-1]
+        # The empty prefix 0 is always there.
```
divergentdave, line 166: See above
```suggestion
```

poc/vdaf_poplar1.py
```diff
@@ -151,16 +151,34 @@ def shard(Poplar1, measurement, nonce, rand):
...
+        last_prefixes_set = set(list(last_prefixes))
```
divergentdave, line 167: See above
```suggestion
        last_prefixes_set = set(last_prefixes)
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3975,15 +3960,38 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+
+    # Check that level increased.
+    if level <= last_level
+      return False
+
+    # Check that prefixes are suffixes of the last level's prefixes,
+    # unless the last level was 0 (and therefore had no prefixes).
+    if last_level > 0:
```
divergentdave, line 3988: I think this conditional is unnecessary, or at least the explanation in the comment is incorrect, since we already handle the first aggregation separately at the top of the function. If the last level was 0, then we would expect the last prefixes to be '0' and '1' (unless the collector had some reason to skip either). Thus, the checks below should still work, though we are high up enough in the tree that the attacker has nothing to gain in its choice of prefixes.

poc/vdaf_poplar1.py
```diff
@@ -151,16 +151,34 @@ def shard(Poplar1, measurement, nonce, rand):
...
+
+        # Check that level increased.
+        if level <= last_level:
+            return False
+
+        # Check that prefixes are suffixes of the last level's prefixes,
+        # unless the last level was 0 (and therefore had no prefixes).
+        if last_level > 0:
+            for (i, prefix) in enumerate(prefixes):
```
divergentdave, line 176: nit: `i` is unused
```suggestion
            for prefix in prefixes:
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4603,8 +4611,15 @@ perhaps unlikely) for a large set of non-heavy-hitter values to share a common
 prefix, which would be leaked by a prefix tree with a sufficiently small
 threshold.
 
-The only known, general-purpose approach to mitigating this leakage is via
-differential privacy.
+A malicious adversary controlling the Collector and one of the Aggregators can
+further turn arbitrary non-heavy prefixes into heavy ones by tampering with the
+IDPF output at any position. While our construction ensures that the sum of all
+nodes evaluated at a level cannot exceed the value at the parent nodes, this
+still may allow an adversary to discover individual non-heavy strings.
```
divergentdave, line 4618: This is incorrect since we backed out passing the previous aggregate result to `is_valid()`.

draft-irtf-cfrg-vdaf.md
schoppmp: The issue is the following. In the first run, there are no prefixes being passed (i.e., last_prefixes_set is empty). In the second run, either 0 or 1 or both are passed in. `get_ancestor` below will turn both of these to 0 (shifting right by 1 bit), but the membership check fails because the set is empty. The previous version incorrectly always added 0 to the set, here we just skip the check since it's supposed to always pass even on the second run (any string has the empty string as prefix).

draft-irtf-cfrg-vdaf.md
schoppmp: Fixed.

draft-irtf-cfrg-vdaf.md
divergentdave: In the second run, "0, 1, or both" would be the prefixes from the past aggregation parameter, while the current aggregation parameter would have a level of 1 and a list of prefixes that are two bits long. The current level's prefixes get passed to `get_ancestor()` so those would be mapped to 0 or 1, which is correct.

I think the existing unit tests of `is_valid()` are misleading. For starters, they show an aggregation parameter with an empty list of candidate prefixes. This doesn't make sense, as it would produce an empty output share if the VDAF were actually run. The candidate prefixes in later aggregation parameters are shorter than they need to be, see the tests `test_poplar1` and test_generate_test_vectors` for comparison.

schoppmp: Oh, I see. I guess `prefixes` confused me, I thought it refers to the nodes being expanded (the parent), but it's the nodes after the expansion. I'll fix that tomorrow, and the tests as well.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3975,15 +3960,35 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+def is_valid(Poplar1, agg_param, previous_agg_params):
+    # Exit early if this is the first time.
+    if len(previous_agg_params) < 1:
+      return True
+
+    (level, prefixes) = agg_param
+    (last_level, last_prefixes) = previous_agg_params[-1]
+    last_prefixes_set = set(last_prefixes)
+
+    # Check that level increased.
```
cjpatton, line 3981:
```suggestion
    # Check that the level increased.
```

poc/tests/test_vdaf_poplar1.py
cjpatton, line 1: Could we also add some unit tests for `get_ancestor()`?

cjpatton (approved):
LGTM, though I think we could increase test coverage a bit.

poc/tests/test_vdaf_poplar1.py
schoppmp: Done.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3975,15 +3960,35 @@ def prep_shares_to_prep(Poplar1, agg_param, prep_shares):
...
+
+    # Check that the level increased.
+    if level <= last_level
+      return False
+
+    # Check that prefixes are suffixes of the last level's prefixes.
+    for prefix in prefixes:
+      last_prefix = get_ancestor(prefix, level, last_level)
+      if last_prefix not in last_prefixes_set:
+        return False  # Current prefix not a suffix of last level's prefixes.
```
divergentdave, line 3989: This is closer to the Sage, and doesn't need as many columns in the text rendering.
```suggestion
        # Current prefix not a suffix of last level's prefixes.
        return False
```

divergentdave (approved):

PR #336, opened 2024-04-30
Add IDPF extractability to security considerations

### LINKED ISSUE #12: VDAF for quantiles?
With `prio3` it's straightforward to roughly estimate quantiles using a histogram (cf https://docs.rs/prio/0.6.1/prio/vdaf/prio3/type.Prio3Histogram64.html). But what if you wanted to precisely compute, say, the 50th percentile (i.e., the median) or the 95th percentile? Prio isn't well-suited for this (cf [[CGB17], Appendix G]).

If your application allows for multiple rounds of interaction with the clients, maybe you could do a kind of binary search to find the quantile you're after. Suppose the client measurements are arbitrary 32-bit integers and you want to compute the median. Suppose there are `N` measurements in all. Let `buckets = [2^31]`. Have the clients shard their measurements using `buckets` as the bucket boundaries. Execute the VDAF over the set of input shares, getting bucket counts `agg` as the aggregate result. `agg[0]` is equal to the number of measurements less than or equal to `2^31`, and `agg[1] <= N/2`, then the median must be less than `2^31`. On the other hand, if `agg[1] > N/2` then you know the median must greater than `2^31`. You then repeat the process, adjusting the bucket boundaries based on the previous result: if the median is less than `2^31`, then let `buckets = [2^31 - 2^30]`; otherwise let `buckets = [2^31 + 2^30]`.

I'm not sure how many iterations would be needed, or how precisely you could compute quantiles this way. Regardless, ideally there would be a VDAF that allows for computing quantiles without interacting with clients more than once.

[CGB17]: https://crypto.stanford.edu/prio/paper.pdf

### LINKED ISSUE #273: Poplar1 security considerations: explain importance of "extractability"
A recent [email](https://mailarchive.ietf.org/arch/msg/ppm/6DEXt2PS3IGOGWcg5hHzgsIGYyg/) on the PPM list mentioned using Poplar1 and making some optimizations in the traversal of the IDPF tree. One natural way to alter the tree traversal algorithm would be to stop early before reaching the leaf level, but I'm concerned that doing so would lose the benefit of the "extractability" property, which depends on the leaf level field size being larger, and thus it could impact the whole scheme's robustness. Currently, we only mention extractability twice, in the context of the public share and the leaf level field size, but we do not explain its relevance. Is it the case that maliciously constructed reports could evaluate to "1" on multiple prefixes at some inner level, but it would be computationally infeasible to do the same on the leaf level? If so, we should outline the limitations on robustness, and possibly provide advice about skipping tree levels, i.e. to always do one last subset histogram at the leaf level. In what way do the secure sketching scheme and the extractability property contribute to the overall VDAF's robustness? Would the attack on an inner level be brute-forcing keys that produce values that pass secure sketching by pure chance, inside a ~2^-64 soundness error?

cjpatton:
Closes #12.
Closes #273.

First, clarify that the Aggregators need to check that they have the same public share.

Second, clarify which IDPF tree traversals are safe. Since we only guarantee extractability at the last level, it is not safe to use intermediate nodes for any purpose except the heavy hitters tree traversal.


draft-irtf-cfrg-vdaf.md
```diff
@@ -4609,6 +4615,37 @@ differential privacy.
...
+non-zero entry is `1`. Indeed, for an honestly generated Poplar1 report, we
+expect the IDPF tree has exactly one non-zero path from the root to the leaves,
+but this is not guaranteed by the secure sketch alone.
+
+IDPFs are designed to prevent a malicious Client from producing key shares with
+more than one non-zero path in the tree. This property is formalized as
+"extractability" in Appendix D of {{BBCGGI21}}: for any pair of key shares and
+public share produced by a malicious Client and a given level of the tree, one
+can efficiently extract from the Client the underlying prefix (i.e., a prefix
+of `alpha`), in particular from its queries to the random oracle.
```
cjpatton, line 4632: @schoppmp a question: I wonder if we should be worried about using `XofFixedKeyAes128` for generating leaf nodes, since the Poplar paper wants this to be a random oracle?

draft-irtf-cfrg-vdaf.md
```diff
@@ -4574,7 +4574,13 @@ exposing" a measurement by including it too many times in a single batch or
 across multiple batches. It is RECOMMENDED that the nonce generated by the
 Client be used by the Aggregators for replay protection.
 
-## Requirements for the Aggregation Parameters
+## Requirements for the Public Share
+
+The Aggregators MUST verify that have both received the same public share from
+the Client. It is sufficient, for example, to exchange a hash of the pubic
```
branlwyd, line 4580:
```suggestion
the Client. It is sufficient, for example, to exchange a hash of the public
```

(typo nit)

draft-irtf-cfrg-vdaf.md
```diff
@@ -4574,7 +4574,13 @@ exposing" a measurement by including it too many times in a single batch or
...
+share over a secure channel.
```
tgeoghegan, line 4581: Do we satisfy this requirement in DAP?

cjpatton: Yes, since public share is bound to HPKE decryption.

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -4609,6 +4615,37 @@ differential privacy.
...
+
+The secure sketch described in {{poplar1}} is used by the Aggregators to check
+that the shares of the vector obtained by evaluating a Client's IDPF at a
+sequence of candidate prefixes has at most one non-zero entry, and that the
+non-zero entry is `1`. Indeed, for an honestly generated Poplar1 report, we
+expect the IDPF tree has exactly one non-zero path from the root to the leaves,
+but this is not guaranteed by the secure sketch alone.
+
+IDPFs are designed to prevent a malicious Client from producing key shares with
+more than one non-zero path in the tree. This property is formalized as
```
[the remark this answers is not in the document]
cjpatton: Good catch. This needs a bit of thinking. The way I understand it is that a malicious client shouldn't be able to _undetectably_ produce a Poplar1 report that evaluates to 1 at two leaves.

schoppmp (approved):

draft-irtf-cfrg-vdaf.md
tgeoghegan: Maybe I'm missing something: in DAP , the client separately HPKE encrypts input shares to either aggregator's respective key. A malicious could therefore use distinct public shares and construct distinct AADs, no?

divergentdave: The risk that the Poplar paper talks about when the aggregators don't have a consistent view of the public parameters is that the malicious client could control the output of the DPF at two inputs, instead of one, like we want. This happens because choosing different group element correction words for one level, and giving each aggregator a different one, gives the client an extra degree of freedom it shouldn't have.

Since we're considering a malicious client when we consider attempts to break extractability in this manner, we don't need to protect this report's privacy, and thus the only goal that matters is robustness. For robustness, we get to assume that both aggregators are honest, and thus we can assume that the leader will distribute the public share to the helper correctly, without considering the HPKE AAD.

Including the public share in the HPKE AAD is probably more relevant for protecting against privacy attacks, as we wouldn't want to let a rogue leader pass an honest client's input share with a tampered public share.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4581,6 +4587,40 @@ threshold.
...
+level of the tree, it is infeasible for an attacker to find a non-trivial IDPF
+that has a non-zero value at more than one node. (A trivial IDPF is one for
+which the attacker cannot predict the value or position of non-zero nodes. For
+example, it is trivial to produce IDPF key shares and public share whose output
+is always psuedorandom, but this would be detected by the secure sketch.)
+
+The IDPF specified in {{idpf}} only guarantees extractability at the last level
+of the tree. (This is by virtue of using a larger field for the leaves than for
+the inner nodes.) For intermediate levels, it is feasible for a client to
+produce IDPF shares with two non-zero nodes.
```
divergentdave, line 4612: (same idea as above)
```suggestion
produce IDPF shares with two controlled non-zero nodes.
```

PR #339, opened 2024-05-01
Clarify that VDAFs aren't sufficient for DP

### LINKED ISSUE #94: Accommodations for differential privacy
We wish for the VDAF draft to have first-class support for schemes that are amendable to differential privacy (DP). The task for this issue is to identify what changes need to be made, if any, to accommodate DP. In addition, we should consider adding guidance to the draft about best practices.

cjpatton:
Closes #94.

As written, the intro seems to frame MPC as an alternative to DP, which is not true. Rather, we want to frame MPC as a necessary component of a data aggregation system that provides DP.

~Indeed, this change goes a step forward: VDAFs SHOULD be composed with a mechanism for DP.~

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -149,6 +149,16 @@ informative:
     seriesinfo: S&P 2020
     target: https://eprint.iacr.org/2019/074
 
+  MPRV09:
+    title: "Computational Differential Privacy"
+    author:
+      - ins: I. Mironov
+      - ins: O. Pandey
+      - ins: O. Reingold
+      - ins: S. Vadhan
+    seriesinfo: CRYPTO 2009
+    target: https://link.springer.com/chapter/10.1007/978-3-642-03356-8_8
+
   MRH04:
     title: "Indifferentiability, impossibility results on reductions, and applications to the random oracle methodology"
     seriesinfo:
@@ -175,13 +185,6 @@ informative:
     title: "Origin Telemetry"
     date: 2020
     target: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/collection/origin.html
-
-  #Vad16:
-  #  title: "The Complexity of Differential Privacy"
-  #  author:
-  #    - ins: S. Vadhan
-  #  date: 2016
-  #  target: https://link.springer.com/chapter/10.1007/978-3-319-57048-8_7
 
 --- abstract
 
@@ -217,47 +220,42 @@ fraction of people have experienced a given disease.  Thus systems that provide
 aggregate statistics while protecting individual measurements can deliver the
 value of the measurements while protecting users' privacy.
 
-Most prior approaches to this problem fall under the rubric of "differential
-privacy (DP)" {{Dwo06}}. Roughly speaking, a data aggregation system that is
-differentially private ensures that the degree to which any individual
-measurement influences the value of the aggregate result can be precisely
-controlled. For example, in systems like RAPPOR {{EPK14}}, each user samples
-noise from a well-known distribution and adds it to their measurement before
-submitting to the aggregation server. The aggregation server then adds up the
-noisy measurements, and because it knows the distribution from whence the noise
-was sampled, it can estimate the true sum with reasonable precision.
+This problem is often formulated in terms of differential privacy (DP)
+{{Dwo06}}. Roughly speaking, a data aggregation system that is differentially
+private ensures that the degree to which any individual measurement influences
+the value of the aggregate result can be precisely controlled. For example, in
+systems like RAPPOR {{EPK14}}, each user samples noise from a well-known
+distribution and adds it to their measurement before submitting to the
+aggregation server. The aggregation server then adds up the noisy measurements,
+and because it knows the distribution from which the noise was sampled, it can
+estimate the true sum with reasonable accuracy.
 
-Differentially private systems like RAPPOR are easy to deploy and provide a
-useful guarantee. On its own, however, DP falls short of the strongest privacy
-property one could hope for. Specifically, depending on the "amount" of noise a
-client adds to its measurement, it may be possible for a curious aggregator to
-make a reasonable guess of the measurement's true value. Indeed, the more noise
-the clients add, the less reliable will be the server's estimate of the output.
-Thus systems employing DP techniques alone must strike a delicate balance
+However, even when noise is added to the measurements, collecting them in the
+clear still reveals a significant amount of information to the collector. On
+the one hand, depending on the "amount" of noise a client adds to its
+measurement, it may be possible for a curious collector to make a reasonable
+guess of the measurement's true value. On the other hand, the more noise the
+clients add, the less reliable will be the server's estimate of the output.
+Thus systems relying solely on a DP mechanism must strike a delicate balance
 between privacy and utility.
 
 The ideal goal for a privacy-preserving measurement system is that of secure
 multi-party computation (MPC): No participant in the protocol should learn
 anything about an individual measurement beyond what it can deduce from the
-aggregate. In this document, we describe Verifiable Distributed Aggregation
-Functions (VDAFs) as a general class of protocols that achieve this goal.
+differentially private aggregate {{MPRV09}}. In this document, we describe
+Verifiable Distributed Aggregation Functions (VDAFs) as a general class of
+delegated MPC protocols that can be used to achieve this goal.
 
 VDAF schemes achieve their privacy goal by distributing the computation of the
 aggregate among a number of non-colluding aggregation servers. As long as a
 subset of the servers executes the protocol honestly, VDAFs guarantee that no
 measurement is ever accessible to any party besides the client that submitted
-it. At the same time, VDAFs are "verifiable" in the sense that malformed
+it. VDAFs can also be composed with various DP mechanisms, thereby ensuring the
+aggregate result does not leak too much information about any one measurmment.
+At the same time, VDAFs are "verifiable" in the sense that malformed
 measurements that would otherwise garble the result of the computation can be
 detected and removed from the set of measurements. We refer to this property as
 "robustness".
-
-In addition to these MPC-style security goals of privacy and robustness, VDAFs
-can be composed with various mechanisms for differential privacy, thereby
-providing the added assurance that the aggregate result itself does not leak
-too much information about any one measurement.
-
-> TODO(issue #94) Provide guidance for local and central DP and point to it
-> here.
 
 The cost of achieving these security properties is the need for multiple servers
 to participate in the protocol, and the need to ensure they do not collude to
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -4508,12 +4507,11 @@ picks bogus measurements for the remaining Clients.  Applications can guard
...
-VDAFs do not inherently provide differential privacy {{Dwo06}}.  The VDAF approach
-to private measurement can be viewed as complementary to differential privacy,
-relying on non-collusion instead of statistical noise to protect the privacy of
-the inputs.  It is possible that a future VDAF could incorporate differential
-privacy features, e.g., by injecting noise before the sharding stage and
-removing it after unsharding.
+VDAFs do not inherently provide differential privacy {{Dwo06}}: depending on
+the distribution of the measurements, the aggregate result itself can still
+leak a significant amount of information about the measurements. Thus
+applications SHOULD execute a VDAF with a compatible mechanism for differential
```
OR13, line 4513: How does an implementer know what makes a mechanism compatible?

When is this a MUST?

OR13 (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: > How does an implementer know what makes a mechanism compatible?

I don't think we can answer this question without being concrete about what DP mechanisms are available. There is an I-D for this, but its in early stages and doesn't have a home at IETF yet: https://github.com/wangshan/draft-wang-ppm-differential-privacy/

> When is this a MUST?

It really depends on the application: for Poplar1, we think it's always going to be necessary to stop attacks (https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/336); for Prio3, it doesn't seem strictly necessary in all cases.

cjpatton: For the record, @ekr suggests on list that a SHOULD is inappropriate and the sentence should just be deleted.

draft-irtf-cfrg-vdaf.md
```diff
@@ -217,48 +227,43 @@ fraction of people have experienced a given disease.  Thus systems that provide
...
-Thus systems employing DP techniques alone must strike a delicate balance
-between privacy and utility.
+This problem is often formulated in terms of "differential privacy (DP)"
+{{Dwo06}}. Roughly speaking, a data aggregation system that is differentially
+private ensures that the degree to which any individual measurement influences
+the value of the aggregate result can be precisely controlled. For example, in
+systems like RAPPOR {{EPK14}}, each user samples noise from a well-known
+distribution and adds it to their measurement before submitting to the
+aggregation server. The aggregation server then adds up the noisy measurements,
+and because it knows the distribution from whence the noise was sampled, it can
```
bifurcation, line 237: "from whence" seems fancy, when "from which" would do.

draft-irtf-cfrg-vdaf.md
```diff
@@ -217,48 +227,43 @@ fraction of people have experienced a given disease.  Thus systems that provide
...
+estimate the true sum with reasonable accuracy.
+
+However even when the measurements are noised, collecting them in the clear
```
bifurcation, line 240:
```suggestion
However, even when the measurements are noised, collecting them in the clear
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4508,12 +4507,11 @@ picks bogus measurements for the remaining Clients.  Applications can guard
 against these risks by adding additional controls on report submission,
 such as Client authentication and rate limits.
 
-VDAFs do not inherently provide differential privacy {{Dwo06}}.  The VDAF approach
-to private measurement can be viewed as complementary to differential privacy,
-relying on non-collusion instead of statistical noise to protect the privacy of
-the inputs.  It is possible that a future VDAF could incorporate differential
-privacy features, e.g., by injecting noise before the sharding stage and
-removing it after unsharding.
+VDAFs do not inherently provide differential privacy {{Dwo06}}: depending on
```
martinthomson, line 4510:
```suggestion
VDAFs do not inherently provide differential privacy {{Dwo06}}. Depending on
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -217,48 +227,43 @@ fraction of people have experienced a given disease.  Thus systems that provide
...
+distribution and adds it to their measurement before submitting to the
+aggregation server. The aggregation server then adds up the noisy measurements,
+and because it knows the distribution from whence the noise was sampled, it can
+estimate the true sum with reasonable accuracy.
+
+However even when the measurements are noised, collecting them in the clear
+still reveals a significant amount of information to the collector.
+Specifically, depending on the "amount" of noise a client adds to its
+measurement, it may be possible for a curious collector to make a reasonable
+guess of the measurement's true value. Indeed, the more noise the clients add,
```
bifurcation, line 244: Instead of "Specifically, ... Indeed...", I might use "On the one hand... On the other hand..."

draft-irtf-cfrg-vdaf.md
```diff
@@ -4508,12 +4507,11 @@ picks bogus measurements for the remaining Clients.  Applications can guard
...
 
-VDAFs do not inherently provide differential privacy {{Dwo06}}.  The VDAF approach
-to private measurement can be viewed as complementary to differential privacy,
-relying on non-collusion instead of statistical noise to protect the privacy of
-the inputs.  It is possible that a future VDAF could incorporate differential
-privacy features, e.g., by injecting noise before the sharding stage and
-removing it after unsharding.
+VDAFs do not inherently provide differential privacy {{Dwo06}}: depending on
+the distribution of the measurements, the aggregate result itself can still
+leak a significant amount of information about the measurements. Thus
```
martinthomson, line 4512:
```suggestion
leak a significant amount of information about individual measurements or measurements
that relate to a single person or group. Thus
```

draft-irtf-cfrg-vdaf.md
martinthomson, line 4513:
```suggestion
applications might compose a VDAF with a compatible mechanism for differential
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -217,48 +227,43 @@ fraction of people have experienced a given disease.  Thus systems that provide
...
+the less reliable will be the server's estimate of the output. Thus systems
+relying solely on a DP mechanism must strike a delicate balance between privacy
+and utility.
 
 The ideal goal for a privacy-preserving measurement system is that of secure
 multi-party computation (MPC): No participant in the protocol should learn
 anything about an individual measurement beyond what it can deduce from the
-aggregate. In this document, we describe Verifiable Distributed Aggregation
-Functions (VDAFs) as a general class of protocols that achieve this goal.
+differntially private aggregate {{MPRV09}}. In this document, we describe
```
martinthomson, line 252:
```suggestion
differentially private aggregate {{MPRV09}}. In this document, we describe
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -217,48 +227,43 @@ fraction of people have experienced a given disease.  Thus systems that provide
...
-
-Differentially private systems like RAPPOR are easy to deploy and provide a
-useful guarantee. On its own, however, DP falls short of the strongest privacy
-property one could hope for. Specifically, depending on the "amount" of noise a
-client adds to its measurement, it may be possible for a curious aggregator to
-make a reasonable guess of the measurement's true value. Indeed, the more noise
-the clients add, the less reliable will be the server's estimate of the output.
-Thus systems employing DP techniques alone must strike a delicate balance
-between privacy and utility.
+This problem is often formulated in terms of "differential privacy (DP)"
```
martinthomson, line 230:
```suggestion
This problem is often formulated in terms of differential privacy (DP)
```

draft-irtf-cfrg-vdaf.md
martinthomson, line 240:
```suggestion
However even when noise is added to measurements, collecting them in the clear
```

draft-irtf-cfrg-vdaf.md
bifurcation: Might I suggest one of the additional options in RFC 6919, e.g., [SHOULD CONSIDER](https://rfcs.online/rfcs/rfc6919.html#section-2)?

More seriously, the idea of composition seems worth mentioning, but I agree that RFC 2119 language probably isn't appropriate.

bifurcation (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: @martinthomson also concurs on list, I'll remove it!

draft-irtf-cfrg-vdaf.md
cjpatton: Changed to "an individual measurement or the person that generated it", PTAL.

draft-irtf-cfrg-vdaf.md
cjpatton: I deleted the sentence per @ekr's suggestion ... I can add it back with "might" if you like.

draft-irtf-cfrg-vdaf.md
cjpatton: Thanks for crushing my dreams, Richard.

draft-irtf-cfrg-vdaf.md
```diff
@@ -149,6 +149,16 @@ informative:
...
+  MPRV09:
+    title: "Computational Differential Privacy"
+    author:
+      - ins: I. Mironov
+      - ins: O. Pandey
+      - ins: O. Reingold
+      - ins: S. Vadhan
+    seriesinfo: CRYPTO 2009
+    target: https://link.springer.com/chapter/10.1007/978-3-642-03356-8_8
+
```
divergentdave, line 161: We can also delete the commented-out "Vad16" reference below. That was just used [here](https://github.com/cjpatton/ppm/blob/1d38aacd713ec7f23a3de3a41c5166e29ef43be2/draft-patton-multi-ppm.md?plain=1#L377) early in the history of the document, and the other references we have are sufficient.

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -217,48 +220,43 @@ fraction of people have experienced a given disease.  Thus systems that provide
...
 between privacy and utility.
 
 The ideal goal for a privacy-preserving measurement system is that of secure
 multi-party computation (MPC): No participant in the protocol should learn
 anything about an individual measurement beyond what it can deduce from the
-aggregate. In this document, we describe Verifiable Distributed Aggregation
-Functions (VDAFs) as a general class of protocols that achieve this goal.
+differentially private aggregate {{MPRV09}}. In this document, we describe
+Verifiable Distributed Aggregation Functions (VDAFs) as a general class of
+delegated MPC protocols that can be used to achieve this goal.
```
tgeoghegan, line 247: What does "delegated" MPC mean? Is the distinction useful in this document?

tgeoghegan (commented):
~I agree with Richard that a SHOULD CONSIDER is appropriate.~ Oh I get the joke

draft-irtf-cfrg-vdaf.md
cjpatton: "Delegated" means a small set of servers carry out the computation on behalf of a large number of clients providing the inputs.

martinthomson: Is that a term of art?  It's the first I've heard it.

chris-wood (approved):

PR #340, opened 2024-05-01
Tweak nonce requirements

tgeoghegan:
- Clarify that aggregators must verify that nonces are never re-used. Since VDAF aims to provide privacy in the face of malicious clients, it doesn't suffice to say clients MUST generate nonces using a CSPRNG; we have to account for malicious clients by adding a MUST for the aggregator. This lines up with the behavior DAP has specified for a long time now.

- In the second paragraph, clarify that over exposing a *report* is the risk, not a *measurement*. It's always possible for the same measurement to occur many times (for instance, in `Prio3Count`, most measurements are 1), but we want the enclosing *report* to be unique.

See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/558 for discussion


cjpatton (approved):

cjpatton:
For the record: it's not clear this is strictly necessary, but it does correspond to the threat model described in DPRS23. And since we already have the requirement in DAP, we might as well adopt it here.

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -4563,16 +4563,17 @@ scheme would not allow key rotation over the lifetime of a task.
...
 associated with the Client's report. To ensure privacy of the underlying
-measurement, the Client MUST generate this nonce using a CSPRNG. This is
-required in order to leverage security analysis for the privacy definition of
-{{DPRS23}}, which assumes the nonce is chosen at random prior to generating the
-report.
+measurement, honest Clients MUST generate this nonce using a CSPRNG and the
+Aggregators MUST verify that nonces are never re-used, to account for malicious
+clients. This is required in order to leverage security analysis for the privacy
+definition of {{DPRS23}}, which assumes the nonce is chosen at random prior to
+generating the report.
```
cjpatton, line 4570: Actually, I'm a little hesitant to create this requirement for non-DAP applications. And I'm not sure it's strictly necessary.

```suggestion
measurement, the Client MUST generate this nonce using a CSPRNG. This is
required in order to leverage security analysis for the privacy definition of
{{DPRS23}}, which assumes the nonce is chosen at random prior to generating the
report.
```

First, let me to clarify my claim on the thread in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/558.  First, the security game in {{DPRS23}} models honest clients that always generate nonces at random. It does not say anything about privacy once a collision occurs, but if the nonce space is sufficiently large, then the probability of a collision is negligible.

Now, let's consider an attacker that sends the honest aggregator a report with the nonce generated by an honest client. For Prio3, what it learns is a verifier share computed with the same query randomness, but over an input and proof it already knows. The question then is what does the honest client's input as a result? Observe that the attacker can compute this verifier share on its own; there is nothing more it can learn by interacting with the honest aggregator.

This makes me think that replay protection is not necessary for the _base_ privacy goal of VDAF. (It would be necessary for DP.)

@hannahdaviscrypto I would appreciate your view here.

cjpatton (changes requested):

draft-irtf-cfrg-vdaf.md
hannahdaviscrypto: I agree with this change. It's true that for Prio3, the attacker can't learn anything by submitting an input share with a repeated nonce and known input/proof to the honest aggregator, and in the {{DPRS23}} model, this is the only way it can repeat a nonce.

But implicitly, the {{DPRS23}} model is relying on non-malleability of the channel between the prover and the honest aggregator, which some VDAFs might not want to assume. If the attacker can submit a report with the same nonce and input share as an honest client but a tampered proof, I think we've lost security in a way that isn't covered by the {{DPRS23}} model (but would be by the {{MPDST24}} model).

In DAP, of course, we should have non-malleability and don't need to worry about it. But if we're specifically talking about non-DAP applications, it seems wise to keep the requirements on them as light as possible. 

More to the point, is there a downside to asking this of aggregators? Is there a target application where it would substantially help to allow repeated nonces?

tgeoghegan: The downside of this requirement is that checking for nonce uniqueness is expensive and so relaxing the requirement can make protocols easier to implement or at least to operate (fewer expensive database lookups).

cjpatton: By {{MPDST24}} you mean https://eprint.iacr.org/2024/221, correct?

This non-malleability property is important and needs to be called out. I believe this is provided by HPKE, as long as it is IND-CCA. Do you agree? 

> In DAP, of course, we should have non-malleability and don't need to worry about it. But if we're specifically talking about non-DAP applications, it seems wise to keep the requirements on them as light as possible.
>
> More to the point, is there a downside to asking this of aggregators? Is there a target application where it would substantially help to allow repeated nonces?

You bring up a very interesting point. On the one hand, we would like security requirements to be as minimal as possible, i.e., not requiring non-malleability of the channel would be good. On the other hand, detecting replayed nonces is a rather stringent _operational_ requirement. In fact, this is what spurred @tgeoghegan's PR in the first place: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/556

To answer your question: anti-replay has a high operational cost, and while it already exists for DAP, I wouldn't want to impose it on every application.

cjpatton: On balance, I would prefer to spell out that security consideration than impose the operational consideration. What do you two think?

cjpatton: > Thanks for highlighting the operational cost. In that case I think the question is this: are there any VDAFs we're interested in where, unlike Prio3, privacy could be compromised by submitting a known report with a repeated nonce to an honest aggregator? If so, I think that the security concern would probably outweigh the cost concern and we should make this change. If not, letting applications pick between using HPKE and unique nonces makes sense.

Let me see if I understand the attack:
1. Honest client uploads a report; the attacker interacts with the honest aggregator to process it
2. Attacker uploads a report with the same nonce; the attacker interacts with the honest aggregator to process it
3. Attacker guesses a bit of the honest measurement based on its share of the honest report and the messages exchanged

hannahdaviscrypto: In step 3, possibly also the final result (i.e., whether the honest aggregator accepted the fake report). Otherwise, yes.

cjpatton: Thanks. At the moment I can't imagine a VDAF for which this attack is possible. It would require preparation of the honest report to leak something it's not supposed to, correct? In which case the repeated nonce doesn't matter.

Perhaps we should try to flesh out your counter example?

cjpatton (approved):

cjpatton:
Ah, I forgot that the draft already reacommended using the nonce for replay protection, and that this just adds clarity to this. Looks like we're good to go! I'm gonna merge, and I don't think there's a need for a second PR.

PR #345, opened 2024-06-05
Add guidance for encoding byte strings as IDPF indices

### LINKED ISSUE #101: Guidance for encoding IDPF index
IDPFs (and Poplar1 by extension) take in an "index" `alpha` represented as an `Unsigned`. @bifurcation pointed out in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/97#discussion_r918280713 that we also call this a "bit-string", which is ambiguous. The draft needs guidance on how applications should represent bit strings as IDPF indexes.

cjpatton:
Closes #101.

Big-endian encoding is the most natural choice as it aligns most naturally with our definition of index. We also should note that a reversible padding scheme is needed when the inputs might be variable length.

While at it, pick up a few minor editorial improvements:

- "set of candidate preixes" -> "sequence of candidate prefixes"
- Use one free variable in defining index prefixes
- "Vec" -> "list"
- "Client" -> "Sharding" for consistency with the Prio3 section

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3519,11 +3519,11 @@ measurement is sharded. This is provided to the FLP by Prio3.
 
 # Poplar1 {#poplar1}
 
-This section specifies Poplar1, a VDAF for the following task. Each Client holds
-a string of length `BITS` and the Aggregators hold a set of `l`-bit strings,
-where `l <= BITS`. We will refer to the latter as the set of "candidate
-prefixes". The Aggregators' goal is to count how many measurements are prefixed
-by each candidate prefix.
+This section specifies Poplar1, a VDAF for the following task. Each Client
+holds a bit-string of length `BITS` and the Aggregators hold a sequence of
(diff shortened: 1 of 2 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -3711,10 +3734,10 @@ subsections. These methods make use of constants defined in {{poplar1-const}}.
 | USAGE_VERIFY_RAND: Unsigned | 4     |
 {: #poplar1-const title="Constants used by Poplar1."}
 
+### Sharding
 
-### Client
-
-The Client's measurement is interpreted as an IDPF index, denoted `alpha`. The
+The Client's measurement is as an IDPF index, denoted `alpha`. (See
```
divergentdave, line 3739: nit: extra word
```suggestion
The Client's measurement is an IDPF index, denoted `alpha`. (See
```

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: Done.

PR #349, opened 2024-06-06
Add Prio3MultihotCountVec to the draft

### LINKED ISSUE #287: Prio3 VDAF for vectors with a bounded number of 1s

cjpatton:
Closes #287.

Assign Prio3MultiHotHistogram a codepoint (0x00000004) and add it to the draft under the name Prio3MultihotCountVec. Also, rename "count" to "weight", as in the "weight of the vector" (i.e., the number of non-zero entries).

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3616,6 +3616,115 @@ measurement is sharded. This is provided to the FLP by Prio3.
 | `Field`          | `Field128` ({{fields}})                         |
 {: title="Parameters of validity circuit Histogram."}
 
+### Prio3MultihotCountVec
+
+For this instance of Prio3, each measurement is a vector of ones and zeros,
+where the number of ones is bounded. This provides a functionality similar to
+Prio3Histogram except that more than one entry may be non-zero. This allows
+Prio3MultihotCountVec to be composed with a randomized response mechanism, like
+{{EPK14}}, for providing differential privacy. (For example, each Client would
+set each entry to one with some small probability.)
+
+Prio3MultihotCountVec uses XofTurboShake128 ({{xof-turboshake128}}) as its XOF.
+Its validity circuit, denoted `MultihotCountVec`, uses `Field128` ({{fields}})
+as its default finite field. It has three parameters: `length`, the number of
+of entries in the count vector; `max_weight`, the maximum number of non-zero
+entries (i.e., the weight must be at most `max_weight`); and `chunk_length`,
+used the same way as in {{prio3sumvec}} and {{prio3histogram}}.
+
+Validation works as follows. Let
+
+* `bits_for_weight = max_weight.bit_length()`
+* `offset = 2**bits_for_weight - 1 - max_weight`
+
+The Client reports the weight of the count vector by adding `offset` to it and
+bit-encoding the result. Observe that only a weight of at most `max_weight` can
+be encoded with `bits_for_weight` bits.
+
+The verifier checks that each entry of the encoded measurement is a bit (i.e.,
+either one or zero). It then decodes the reported weight and subtracts it from
+`offset + sum(count_vec)`, where `count_vec` is the count vector. The result is
+zero if and only if the reported weight is equal to the true weight.
+
+Encoding, truncation, and decoding are defined as follows:
+
+~~~
+def encode(self, measurement):
+    if len(measurement) != self.length:
+        raise ValueError('invalid Client measurement length')
+
+    # The first part is the vector of counters.
+    count_vec = list(map(self.Field, measurement))
+
+    # The second part is the reported weight.
+    weight_reported = sum(count_vec, self.Field(0))
+
+    encoded = []
+    encoded += count_vec
+    encoded += self.Field.encode_into_bit_vector(
+        (self.offset + weight_reported).as_unsigned(),
+        self.bits_for_weight)
+    return encoded
+
+def truncate(self, meas):
+    return meas[:self.length]
+
+def decode(self, output, _num_measurements):
+    return [bucket_count.as_unsigned() for bucket_count in output]
+~~~
+
+Circuit evaluation is defined as follows:
+
+~~~
+def eval(self, meas, joint_rand, num_shares):
+    # Check that each entry in the input vector is one or zero.
+    range_check = self.Field(0)
+    r = joint_rand[0]
+    r_power = r
+    shares_inv = self.Field(num_shares).inv()
+    for i in range(self.GADGET_CALLS[0]):
+        inputs = [None] * (2 * self.chunk_length)
+        for j in range(self.chunk_length):
+            index = i * self.chunk_length + j
+            if index < len(meas):
+                meas_elem = meas[index]
+            else:
+                meas_elem = self.Field(0)
+
+            inputs[j * 2] = r_power * meas_elem
+            inputs[j * 2 + 1] = meas_elem - shares_inv
+
+            r_power *= r
+
+        range_check += self.GADGETS[0].eval(self.Field, inputs)
+
+    # Check that the weight `offset` plus the sum of the counters
+    # is equal to the value claimed by the Client.
+    count_vec = meas[:self.length]
+    weight = sum(count_vec, self.Field(0))
+    weight_reported = \
+        self.Field.decode_from_bit_vector(meas[self.length:])
+    weight_check = self.offset*shares_inv + weight - \
+        weight_reported
+
+    out = joint_rand[1] * range_check + \
+        joint_rand[1] ** 2 * weight_check
+    return out
+~~~
+
+| Parameter        | Value                                           |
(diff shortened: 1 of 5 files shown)


cjpatton:
Changed base to main.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3593,6 +3593,101 @@ measurement is sharded. This is provided to the FLP by Prio3.
 | `Field`          | `Field128` ({{fields}})                         |
 {: title="Parameters of validity circuit Histogram."}
 
+### Prio3MultihotCountVec
+
+For this instance of Prio3, each measurement is a vector of ones and zeros,
+where the number of ones is bounded. This provides a functionality similar to
+Prio3Histogram except that more than one entry may be set. This allows
+Prio3MultihotCountVec to be composed with a randomized respjonse mechanism,
```
junyechen1996, line 3601:
```suggestion
Prio3MultihotCountVec to be composed with a randomized response mechanism,
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3593,6 +3593,101 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+like {{EPK14}}, for providing differential privacy. (For example, each Client
+would set each entry to one with some small probability.)
+
+Prio3MultihotCountVec uses XofTurboShake128 ({{xof-turboshake128}}) as its XOF.
+Its validity circuit, denoted `MultihotCountVec`, uses `Field128` ({{fields}})
+as its default finite field. It has three parameters: `length`, the number of
+of entries in the count vector; `max_weight`, the maximum weight of the vector
+(i.e., the number of non-zero entries); and `chunk_length`, which is used by by
```
junyechen1996, line 3609:
```suggestion
(i.e., the number of non-zero entries); and `chunk_length`, which is used by
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3593,6 +3593,101 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+~~~
+def encode(self, measurement):
+    if len(measurement) != self.length:
+        raise ValueError('invalid Client measurement length')
+
+    # The first part is the vector of counters.
+    encoded = list(map(self.Field, measurement))
+
+    # The second part is the bit-encoded, offset weight.
+    offset_weight = self.offset + sum(encoded, self.Field(0))
```
junyechen1996, line 3625: Should we quickly explain what `offset` is in the draft itself?

draft-irtf-cfrg-vdaf.md
```diff
@@ -3593,6 +3593,101 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+def truncate(self, meas):
+    return meas[:self.length]
+
+def decode(self, output, _num_measurements):
+    return [bucket_count.as_unsigned() for bucket_count in output]
+~~~
+
+The `MultihotCountVec` circuit checks the encoded measurement has two
+properties. First, each entry is either a zero or a one. Second, the weight of
+the vector, when offset by `max_weight` is equal to the offset weight reported
```
junyechen1996, line 3639: Related comment about `offset`: I don't think the circuit is offsetting by `max_weight`? The `offset` is the difference between `max_weight` and `next_power_of_2(max_weight) - 1`.

junyechen1996 (commented):
Have one comment about the definition of `offset`, otherwise LGTM.

cjpatton:
Addressed Junye's comments and rebased.

junyechen1996 (approved):

draft-irtf-cfrg-vdaf.md
```diff
@@ -3595,6 +3595,118 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+Its validity circuit, denoted `MultihotCountVec`, uses `Field128` ({{fields}})
+as its default finite field. It has three parameters: `length`, the number of
+of entries in the count vector; `max_weight`, the maximum weight of the vector
+(i.e., the number of non-zero entries); and `chunk_length`, used the same way
+as in {{prio3sumvec}} and {{prio3histogram}}. It is RECOMMENDED to set
+`chunk_length` to an integer near the square root of `length` (see
+{{parallel-sum-chunk-length}}).
+
+The reported weight is encoded in a way that ensures it is in
+`range(max_weight)`. Specifically:
```
divergentdave, line 3617: We shouldn't use `range(max_weight)` Python-style notation here (and elsewhere), as `max_weight` should be an inclusive upper bound, but `range()` has an exclusive upper bound.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3595,6 +3595,118 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+
+* Let `offset = 2**bits_for_weight - 1 - max_weight` The reported weight is
```
divergentdave, line 3619: We should define `bits_for_weight` first, before this definition.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3595,6 +3595,118 @@ measurement is sharded. This is provided to the FLP by Prio3.
...
+        joint_rand[1] ** 2 * weight_check
+    return out
+~~~
+
+| Parameter        | Value                                           |
+|:-----------------|:------------------------------------------------|
+| `GADGETS`        | `[ParallelSum(Mul(), chunk_length)]`            |
+| `GADGET_CALLS`   | `[(length + bits_for_weight + chunk_length - 1) // chunk_length]` |
+| `MEAS_LEN`       | `length + bits_for_weight`                      |
+| `OUTPUT_LEN`     | `length + bits_for_weight`                      |
```
divergentdave, line 3702: This is incorrect, as we remove some of these elements during `truncate()`.
```suggestion
| `OUTPUT_LEN`     | `length`                                        |
```

poc/flp_generic.py
```diff
@@ -705,44 +711,49 @@ class MultiHotHistogram(Valid):
...
-        Instantiate an instance of the `MultiHotHistogram` circuit with the
-        given length, max_count, and chunk_length.
+        Instantiate an instance of the this circuit with the given
+        `length`, `max_weight`, and `chunk_length`.
+
+        Pre-conditions:
+
+            - `length > 0`
+            - `0 < max_weight` and `max_weight <= length`
+            - `chunk_Length > 0`
```
divergentdave, line 723: nit: typo
```suggestion
            - `chunk_length > 0`
```

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
divergentdave: Hmm, I think that making the VDAF's parameter an exclusive bound on the measurement's weight, rather than an inclusive bound, is confusing to the end user. I'd rather say "I want between zero and five ones, so I'll construct `Prio3MultihotCountVec` with `max_weight = 5`", than "I want between zero and five ones, so I'll construct `Prio3MultihotCountVec` with `weight_bound = 6`". Plus, the reader would have to be familiar with Python to catch the nuance.

cjpatton: Hmm, I figured aligning with Python would make things clearer. I don't feel strongly about this though, and since it's what we we already doing, I'm happy to revert.

cjpatton: Done - PTAL

divergentdave (approved):

PR #351, opened 2024-06-06
Enforce type hints in field.py, xof.py

cjpatton:
Partially addresses #59.

Add type hint enforcement with mypy to CI, starting with field.py and xof.py. This checks that the variables passed into a function match the function's type hints.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,8 +24,12 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Install dependencies
-      run: sage --pip install pycryptodomex
+      run: sage --pip install pycryptodomex mypy
 
     - name: Run tests
       working-directory: poc
       run: sage -python -m unittest
+
+    - name: Enforce type hints
+      working-directory: poc
+      run: sage -python -m mypy xof.py field.py
--- a/poc/field.py
+++ b/poc/field.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from sage.all import GF, PolynomialRing
+from sage.all import GF, PolynomialRing  # type: ignore
 
 from common import ERR_DECODE, from_le_bytes, to_le_bytes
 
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -24,8 +24,12 @@ jobs:
...
-      run: sage --pip install pycryptodomex
+      run: sage --pip install pycryptodomex mypy
 
     - name: Run tests
       working-directory: poc
       run: sage -python -m unittest
+
+    - name: Enforce type hints
+      working-directory: poc
+      run: sage -python -m mypy xof.py field.py
```
cjpatton, line 35: @divergentdave I added this to the test workflow simply becuase we already have sage as a dependency.  Do you think it would be better to add this as a different workflow?

Also, I'm starting with xof.py and field.py because these were straightforward to fix. We'll add more as we go.

divergentdave: Sounds good, this is fast enough to install and run that splitting it wouldn't be worth the parallelism.

poc/field.py
```diff
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from sage.all import GF, PolynomialRing
+from sage.all import GF, PolynomialRing  # type: ignore
```
divergentdave, line 5: It looks like we could improve upon this situation by generating some `.pyi` files with `stubgen`, or writing them by hand, but I think that should wait until after this first pass.

`sage.all` is kind of terrifying, because it is responsible for importing modules in the right order (and this is load bearing!), it imports all of Sage's libraries, which is wasteful from our perspective, and it does some runtime reflection tricks that probably make things harder for static analysis tools.

divergentdave (approved):

divergentdave:
BTW, it would be good to run mypy on unit tests as well, to help catch mistakes.

poc/field.py
cjpatton: I wonder if we should even bother. We have already encapsulated most of the sage stuff behind abstractions like `Field`.

PR #354, opened 2024-06-08
Enforce type hints for flp.py, flp_generic.py, and tests

cjpatton:
Partially addresses #59.

The "associated type" pattern we're trying to follow is not enforceable.
For example, we'll do something like this:

```
class Foo:
    Measurement = None
    def bar(self, m: Measurement): pass
```

`mypy` complains because `None` is not a valid type. We instead could do
the following:

```
class Foo:
    Measurement: Any = None
    def bar(self, m: Measurement): pass
```

A concrete sub-class would override this in the natural way:

```
class FooForInt(Foo):
    Measurement = int
    def bar(self, m: int): pass

Foo().bar(23)
Foo().bar("23")
FooForInt().bar(23)
FooForInt().bar("23") # fails
```

### DIFF
--- a/poc/flp.py
+++ b/poc/flp.py
@@ -1,17 +1,18 @@
 """Fully linear proof (FLP) systems."""
 
+from typing import Any
+
 import field
 from common import vec_add, vec_sub
-from field import Field
 
 
 class Flp:
     """The base class for FLPs."""
 
     # Generic paraemters
-    Measurement = None
-    AggResult = None
-    Field: field.Field = None
+    Measurement: Any = None
+    AggResult: Any = None
+    Field: Any = None
 
     # Length of the joint randomness shared by the prover and verifier.
     JOINT_RAND_LEN: int
@@ -111,9 +112,9 @@ class Flp:
         return []
 
 
-def additive_secret_share(vec: list[Field],
+def additive_secret_share(vec: list[field.Field],
                           num_shares: int,
-                          field: type) -> list[list[Field]]:
+                          field: type[field.Field]) -> list[list[field.Field]]:
     shares = [
         field.rand_vec(len(vec))
         for _ in range(num_shares - 1)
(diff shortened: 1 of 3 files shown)


poc/flp_generic.py
```diff
@@ -458,11 +455,8 @@ class ParallelSum(Gadget):
     """
 
     # Operational parameters
-    subcircuit = None
-    count = None
-
-    ARITY = None  # Set by constructor
-    DEGREE = None  # Set by constructor
```
cjpatton, line 465: Reviewer note: this is an error because `ARITY` and `DEGREE` are declared to have type `int`. Similarly below.

poc/flp.py
```diff
@@ -1,17 +1,18 @@
...
 class Flp:
     """The base class for FLPs."""
 
     # Generic paraemters
-    Measurement = None
-    AggResult = None
-    Field: field.Field = None
+    Measurement: Any = None
+    AggResult: Any = None
+    Field: Any = None
```
cjpatton, line 15: Reviewer note: I tried this because I thought it would be more natural:

```suggestion
    Field: type[field.Field]
```

This doesn't work: ttps://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

Another alternative is to make this generic. For instance we could do:
```python
F = TypeVar('F', bound=field.Field)
class Flp(Generic[F]):
   field: F

class FlpField64(Flp[Field64]):
   field = Field64
```

Generic may in fact be the way to go in the long run, but it requires more refactoring.

Note that we still we don't get to use `F` as a type: it exists only to support type hints. Thus it is still necessary to define `field` as an attribute.

Also, note that as of python 3.12, there is some syntactic sugar that makes this look nicer:
```python
class Flp[F: Field]:
   field: F
```

divergentdave: I think the problem here is fundamentally one of partial initialization, due to how we use class factories. Instead of `Any`, we could hypothetically use something like `Union[type[field.Field], NoneType]`. Note that `None` would still be a problem even if we had generic type variables, as the type hint would then be `Union[type[F], NoneType]`.

Moreover, the final class we construct from a class factory only really exists at runtime, so static analysis-based type checking will run into some fundamental limitations with this pattern.

divergentdave (approved):

poc/flp.py
cjpatton: This is indeed messy. I find I often end up in trying to over-engineer things like this in Python. I think to address this issue I want to more or less follow what we've already done. Using `Any` for the "abstract base class" solves things without too many changes.

Issue #306, opened 2023-10-25
Generalize range check in Prio3Sum


cjpatton:
In Prio3Sum and Prio3SumVec we do a simple $[0, 2^k)$ range check by encoding the integer as a $k$-bit string and checking each bit is $0$ or $1$. In #287 we devised a simple strategy for generalizing this check to $[0, C)$ for any $C > 0$, even if $C$ is not a power of $2$. What's more, this check is concretely efficient in that it doesn't change the input length or the number of gadget calls. (The affine part of the computation is a bit more complex.)

Could we, and would it be worth it to, do the same for Prio3Sum and Prio3SumVec? It seems to me this would be a valuable improvement.

cc/ @divergentdave, @junyechen1996

divergentdave:
I don't immediately see how we could do this without increasing the input size of Prio3Sum. In the other case, we relied on the already-present range checks on histogram bits for the ≥0 end of the range.

One unoptimized way we could extend Prio3Sum to do different range checks would be to double the input length, encode the measurement as a bit vector, and then encode the measurement plus some constant as a bit vector. The validity circuit would range check all the bits, and then do a sum check on the two halves of the input, confirming that the latter encoded integer is equal to the former plus the constant. This effectively lets us slide two [0, 2^n) range checks over each other, and build a [0, m) range check from their intersection.

cjpatton:
:face_palm: You are absolutely right, we'd need to double the input length to apply the same trick.

I think it's worth doing for Prio3Sum since input is fairly small.

For Prio3SumVec, we definitely don't want to blow up the common "bit-vector" use case ($k=1$). If we wanted to change Prio3SumVec this way, then we probably would want to split out the bit-vector circuit into a different variant, e..g, Prio3CountVec or something.

cjpatton:
We may end up with a breaking change for Prio3Sum as part of #262. If so, we should also consider taking this change as well.

PR #355, opened 2024-06-08
flp: Make the output of `Valid` a vector

### LINKED ISSUE #262: Flp: Let the circuit output be a vector?
While looking at #258, I was considering how one might implement the Prio2 or Prio3CountVec AFE while using Prio3 without joint randomness. No joint randomness means we can't use random linear combinations. Thus, I think we'd be left with taking the conjunction of multiple range check sub-circuits by raising each sub-circuit output to the power $p-1$, which would be pretty expensive. (I think you'd have to split the exponentiation into multiple gadgets for it to be feasible)

This conjunction could be avoided entirely if we let the circuit output a vector instead of a single field element, and require that it be equal to $\vec{0}$ instead of $0$. Circuits that take advantage of such a change would have a bigger verifier message, but it would give us more power/expressiveness without joint randomness.

I'm not sure if this change is actually worth making in the document, but I figure it would be a useful counterfactual in Prio2 vs. Prio3 benchmark conversations, and in the broader joint randomness discussion.

cjpatton:
Closes #262.
Partially addresses #306.

If the input is valid, then each output should be zero. To save bandwidth, we generate an extra element of query randomness and use it to reduce the result.

To exercise this feature, implement a new circuit, `Sum2`, that generalizes `Sum` to support any bound, including non-powers-of-two.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3030,7 +3030,7 @@ the gadget polynomial, are described in detail in {{flp-generic-construction}}.
 #### Extensions {#flp-generic-overview-extensions}
 
 The FLP described in the next section extends the proof system of {{BBCGGI19}},
-Section 4.2 in three ways.
+Section 4.2 in a few ways.
 
 First, the validity circuit in our construction includes an additional, random
 input (this is the "joint randomness" derived from the measurement shares in
@@ -3071,12 +3071,19 @@ inputs are in range `[0,2)` and the last `N-L` inputs are in range `[0,3)`. Of
 course, the same circuit can be expressed using a sub-component that the
 gadgets have in common, namely `Mul`, but the resulting proof would be longer.
 
-Finally, {{BBCGGI19}}, Theorem 4.3 makes no restrictions on the choice of the
+Third, {{BBCGGI19}}, Theorem 4.3 makes no restrictions on the choice of the
 fixed points `alpha[0], ..., alpha[M-1]`, other than to require that the points
 are distinct. In this document, the fixed points are chosen so that the gadget
 polynomial can be constructed efficiently using the Cooley-Tukey FFT ("Fast
 Fourier Transform") algorithm. Note that this requires the field to be
 "FFT-friendly" as defined in {{field-fft-friendly}}.
+
+Finally, the validity circuit in our FLP may have any number of outputs (at
+least one). The input is said to be valid if each of the outputs is zero. To
+save bandwidth, we take a random linear combination of the outputs. If each of
+the outputs is zero, then the reduced output will be zero; but if one of the
+outputs is non-zero, then the reduced output will be non-zero with high
+probability.
 
 ### Validity Circuits {#flp-generic-valid}
 
@@ -3085,16 +3092,17 @@ validity circuit `Valid` that implements the interface described here.
 
 A concrete `Valid` defines the following parameters:
 
-| Parameter        | Description                           |
-|:-----------------|:--------------------------------------|
-| `GADGETS`        | A list of gadgets                     |
-| `GADGET_CALLS`   | Number of times each gadget is called |
-| `MEAS_LEN`       | Length of the measurement             |
-| `OUTPUT_LEN`     | Length of the aggregatable output     |
-| `JOINT_RAND_LEN` | Length of the random input            |
-| `Measurement`    | The type of measurement               |
-| `AggResult`      | Type of the aggregate result          |
-| `Field`          | An FFT-friendly finite field as defined in {{field-fft-friendly}} |
+| Parameter         | Description                           |
+|:------------------|:--------------------------------------|
+| `GADGETS`         | A list of gadgets                     |
+| `GADGET_CALLS`    | Number of times each gadget is called |
+| `MEAS_LEN`        | Length of the measurement             |
+| `OUTPUT_LEN`      | Length of the aggregatable output     |
+| `JOINT_RAND_LEN`  | Length of the random input            |
+| `EVAL_OUTPUT_LEN` | Length of the circuit output          |
+| `Measurement`     | The type of measurement               |
+| `AggResult`       | Type of the aggregate result          |
+| `Field`           | An FFT-friendly finite field as defined in {{field-fft-friendly}} |
 {: title="Validity circuit parameters."}
 
 Each gadget `G` in `GADGETS` defines a constant `DEGREE` that specifies the
@@ -3132,7 +3140,10 @@ def prove_rand_len(self):
 
 def query_rand_len(self):
     """Length of the query randomness."""
-    return len(Valid.GADGETS)
+    query_rand_len = len(self.GADGETS)
+    if self.EVAL_OUTPUT_LEN > 1:
+        query_rand_len += 1
+    return query_rand_len
 
 def proof_len(self):
     """Length of the proof."""
@@ -3243,8 +3254,14 @@ is generated as follows:
    of each gadget in the corresponding table. This step is similar to the
    prover's step (3.) except the verifier does not evaluate the gadgets.
    Instead, it computes the output of the `k`th call to `G_i` by evaluating
-   `poly_gadget_i(alpha_i^k)`. Let `v` denote the output of the circuit
+   `poly_gadget_i(alpha_i^k)`. Let `out` denote the output of the circuit
    evaluation.
+
+1. Next, reduce `out` as follows. If `EVAL_OUTPUT_LEN > 1`, then consume the
+   first element of `query_rand` by letting `[r], query_rand = front(1,
+   query_rand)`. Then let `v = r*out[0] + r**2*out[1] + r**3*out[2] + ...`.
+   That is, interpret the outputs as coefficients of a polynomial `f(x)` and
+   evaluate polynomial `f(x)*x` at a random point `r`.
 
 1. Compute the wire polynomials just as in the prover's step (4.).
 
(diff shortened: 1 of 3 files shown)


cjpatton:
I'm going to update this to speak in terms of "max_measurement" rather than "measurement_bound" per @divergentdave's comment here: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/349#discussion_r1635055085

poc/flp_generic.py
```diff
@@ -888,3 +918,82 @@ def test_vec_set_type_param(self, test_vec):
         test_vec['bits'] = self.bits
         test_vec['chunk_length'] = self.chunk_length
         return ['length', 'bits', 'chunk_length']
+
+
+class Sum2(Valid):
```
junyechen1996, line 923: Can this be more descriptive?

draft-irtf-cfrg-vdaf.md
```diff
@@ -3598,22 +3617,23 @@ def eval(self, meas, joint_rand, num_shares):
...
+| Parameter         | Value                                           |
+|:------------------|:------------------------------------------------|
+| `GADGETS`         | `[ParallelSum(Mul(), chunk_length)]`            |
+| `GADGET_CALLS`    | `[(length + chunk_length - 1) // chunk_length]` |
+| `MEAS_LEN`        | `length`                                        |
+| `OUTPUT_LEN`      | `length`                                        |
+| `JOINT_RAND_LEN`  | `2`                                             |
+| `EVAL_OUTPUT_LEN` | `1`                                             |
+| `Measurement`     | `int`                                           |
+| `AggResult`       | `list[int]`                                     |
```
junyechen1996, line 3635:
```suggestion
| `AggResult`       | `list[int]`, each element in `range(2)`                                     |
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3369,19 +3386,20 @@ def eval(self, meas, joint_rand, _num_shares):
...
-| `MEAS_LEN`       | `bits`                    |
-| `OUTPUT_LEN`     | `1`                       |
-| `JOINT_RAND_LEN` | `1`                       |
-| `Measurement`    | `int` in `range(2**bits)` |
-| `AggResult`      | `int`                     |
-| `Field`          | `Field128` ({{fields}})   |
+    return [out]
+~~~
+
+| Parameter         | Value                     |
```
junyechen1996, line 3392: Will there be a spec for `Sum2` now that we've added a reference implementation?

cjpatton: for now `Sum2` is meant to exercise the new feature in tests. It may eventually replace the circuit in `Prio3Sum`, but that's a question for #306.

poc/flp_generic.py
cjpatton: Renamed to `SumOfRangeCheckedInputs`. This is probably overly verbose; we may consider renaming it to `Sum` as part of #306.

poc/flp_generic.py
```diff
@@ -287,11 +297,25 @@ def query(self, meas, proof, query_rand, joint_rand, num_shares):
...
+        if len(out) != self.Valid.EVAL_OUTPUT_LEN:
+            raise ValueError('circuit has unexpected output length')
 
         if len(query_rand) != self.Valid.query_rand_len():
             raise ValueError('incorrect query randomness length')
 
+        # Reduce the output.
+        if self.Valid.EVAL_OUTPUT_LEN > 1:
+            ([r], query_rand) = front(1, query_rand)
+            r_power = self.Field(1)
```
divergentdave, line 311: FWIW, this is inconsistent with how we reduce vectors inside various sectors, where we multiply by r, r^2, r^3, ..., instead of 1, r, r^2, ...

In this case it's fine as far as security is concerned, since there's no addition after this linear combination happens, so we don't need to worry about the term with coefficient 1 cancelling with anything. If we unified on starting with r, that could help implementers avoid mistakes by being consistent.

poc/flp_generic.py
```diff
@@ -888,3 +918,83 @@ def test_vec_set_type_param(self, test_vec):
...
+        return out
+
+    def encode(self, measurement):
+        x = self.Field(measurement)
+        encoded = [x]
+        encoded += self.Field.encode_into_bit_vector(
+            (x + self.offset).as_unsigned(),
+            self.bits
+        )
+        return encoded
```
divergentdave, line 990: I don't think this encoding and validity circuit achieve their goal. If the parameters are max_measurement=5, bits=3, offset=2, then the malicious encoding [p-1, 0, 0, 1] would produce a circuit output of [0, 0, 0, shares * 2 * shares_inv + p - 1 - 1] = [0, 0, 0, 0].

We could fix this if the encoded measurement length were self.bits * 2, and the encoding routine concatenated encode_into_bit_vector(x) and encode_into_vector(x + offset). I'm not sure whether we can do better.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3450,7 +3468,7 @@ def eval(self, Field, inp):
         start_index = i * self.subcircuit.ARITY
         end_index = (i + 1) * self.subcircuit.ARITY
         out += self.subcircuit.eval(Field, inp[start_index:end_index])
-    return out
+    return [out]
```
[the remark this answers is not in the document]
cjpatton: You're right, that was sloppy.

poc/flp_generic.py
cjpatton: Good point. So the draft language in this PR is "interpret the outputs as coefficients of a polynomial and evaluate that polynomial at a random point", which is what's implemented here. I think this is a bit more intuitive than starting at $r^1=r$.

However I agree we should be consistent. We could either:
1. Start at $r^0=1$ everywhere (like this change does). This has the advantage of not being a breaking change.
2. Start at $r^1=r$ everywhere (like the current circuits do). This would require a breaking change.

I'm on the fence because I think 2 is more intuitive. On the other hand, breaking changes are expensive at this stage. On the the other _other_ hand, we're already considering a breaking change for Prio3Sum (#306).

I don't want to make this change in draft-10, but I think we should consider doing it before RGLC. What do you think?

poc/flp_generic.py
cjpatton: Hmmm, perhaps you can try express this attack as a test? The circuit seems to detect this particular malformed input:
```python
    def test_sum_of_range_checked_inputs_malformed(self):
        flp = FlpGeneric(SumOfRangeCheckedInputs(5))
        self.assertEqual(flp.Valid.bits, 3)
        self.assertEqual(flp.Valid.offset.as_unsigned(), 2)
        test_flp_generic(flp, [
            (
                [
                    flp.Field(flp.Field.MODULUS - 1),
                    flp.Field(0),
                    flp.Field(0),
                    flp.Field(1),
                ],
                False,
            ),
        ])
```
The output of the circuit is `[0, 0, 0, 18446744069414584318]`.

Alternatively, is there a fundamental flaw I'm not seeing?

draft-irtf-cfrg-vdaf.md
```diff
@@ -3059,30 +3059,38 @@ inputs are in range `[0,2)` and the last `N-L` inputs are in range `[0,3)`. Of
...
+Third, {{BBCGGI19}}, Theorem 4.3 makes no restrictions on the choice of the
 fixed points `alpha[0], ..., alpha[M-1]`, other than to require that the points
 are distinct. In this document, the fixed points are chosen so that the gadget
 polynomial can be constructed efficiently using the Cooley-Tukey FFT ("Fast
 Fourier Transform") algorithm. Note that this requires the field to be
 "FFT-friendly" as defined in {{field-fft-friendly}}.
 
+Finally, the validity circuit in our FLP may have any number of outputs (at
+least one). The input is said to be valid if each of the outputs is zero. To
+save bandwidth, we interpret the outputs as coefficients of a polynomial and
```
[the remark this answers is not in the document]
cjpatton: "Linear combination" is a bit more general than this I think? For instance, we could have done $r_1 x_1, + \cdots r_n x_n$ where the $x_i$s are the outputs and the $r_i$s are independent, uniform random field elements.

poc/flp_generic.py
cjpatton: Do you prefer 1 or 2?

divergentdave: Changing the circuits using joint randomness to start with 1 makes me a little nervous. We'd have to carefully analyze circuit soundness if we did so, to make sure we don't end up with two monomials with the same powers of joint randomness values, and different input elements. This is doable, but having an extra r factor makes it easier to justify.

I'm torn between the two options. On the one hand, starting with r has the benefit of inertia, and it's easier to convince oneself of circuit soundness, as discussed above. On the other hand, starting with 1 could be more intuitive to people reading the spec for the first time; in fact this was one of the compatibility bugs we worked through when initially implementing Prio3 in divviup-ts. Maybe if we start with r we should rephrase the prose description.

poc/flp_generic.py
divergentdave: Ah, I encoded the bitvector the wrong way in my head. This fails:

```python
    def test_sum_of_range_checked_inputs_malformed(self):
        flp = FlpGeneric(SumOfRangeCheckedInputs(5))
        self.assertEqual(flp.Valid.bits, 3)
        self.assertEqual(flp.Valid.offset.as_unsigned(), 2)
        test_flp_generic(flp, [
            (
                [
                    flp.Field(flp.Field.MODULUS - 1),
                    flp.Field(1),
                    flp.Field(0),
                    flp.Field(0),
                ],
                False,
            ),
        ])
```

poc/flp_generic.py
cjpatton: > Maybe if we start with r we should rephrase the prose description.

Definitely.

poc/flp_generic.py
cjpatton: Ah, I think I see what's going on here: we need to prevent the offset measurement from wrapping around the field modulus. Bit encoding it is sufficient. Really great catch :)

cjpatton: Fixed.

poc/flp_generic.py
cjpatton: > Changing the circuits using joint randomness to start with 1 makes me a little nervous. We'd have to carefully analyze circuit soundness if we did so, to make sure we don't end up with two monomials with the same powers of joint randomness values, and different input elements. This is doable, but having an extra r factor makes it easier to justify.

Can you elaborate here? I don't know what you mean by "monomial".

Starting with $1$: My intuition here is that if you have a polynomial $f$ of degree $n > 0$, then there are at most $n$ roots, i.e., points $x$ for which $f(x) = 0$. Hence  $Pr[ f(x) = 0 ] \leq n / p$, where $p$ is the field modulus and where $x$ is uniform random.

When we start with $r$ rather $1$, we're doing basically the same thing except that the polynomial we're evaluating is $f(x) \cdot x$ which has degree $n+1$.

divergentdave: "Monomial" here refers to the terms that get added together in the circuit output, after distributing multiplication and expanding everything.

Off the top of my head, I think Histogram would be fine to switch to starting with 1. The output would look like bit_check_0 + bit_check_1 * joint_rand[0] + bit_check_2 * joint_rand[0] ** 2 + sum_check * joint_rand[1].

Starting with r would be robust to things like forgetting joint_rand[1] in the second combination step, because there's already a power of joint_rand[0] on every term from one side, and none on the other side. Thus, we're safe from cancellation even before we multiply by joint_rand[1] and joint_rand[1] ** 2 respectively.

cjpatton: So to your thinking, this is more about defense in depth? Always a good idea, I think. We want to establish patterns that are more likely to be secure when copy-pasted without thinking.

cjpatton: Done! Please take a look.

poc/flp_generic.py
```diff
@@ -888,3 +918,88 @@ def test_vec_set_type_param(self, test_vec):
...
+        encoded = []
+        encoded += self.Field.encode_into_bit_vector(
+            x.as_unsigned(),
+            self.bits
+        )
+        encoded += self.Field.encode_into_bit_vector(
+            (x + self.offset).as_unsigned(),
+            self.bits
+        )
+        return encoded
```
divergentdave, line 995: This could be simplified a little now
```suggestion
        encoded = []
        encoded += self.Field.encode_into_bit_vector(
            measurement,
            self.bits
        )
        encoded += self.Field.encode_into_bit_vector(
            measurement + self.offset.as_unsigned(),
            self.bits
        )
        return encoded
```

divergentdave (approved):

cjpatton:
Rebased and squashed.

PR #356, opened 2024-06-10
Clarify network assumptions and limits on robustness

### LINKED ISSUE #304: Clarify threat model for robustness
Hello, I had a discussion with @cjpatton and @pratiksarkar93 about the VDAF security considerations. The draft states:

```
 VDAFs have two essential security goals:

 1.  Privacy: An attacker that controls the network, the Collector,
     and a subset of Clients and Aggregators learns nothing about the
     measurements of honest Clients beyond what it can deduce from the
     aggregate result.

 2.  Robustness: An attacker that controls the network and a subset of
     Clients cannot cause the Collector to compute anything other than
     the aggregate of the measurements of honest Clients.
```

I was wondering if the notion of **robustness** is referring only to **correctness** or to both correctness and **guaranteed output delivery (GOD)**. Would it make sense to split robustness into these two properties to use more formal MPC terms (i.e., [pages 2 & 3 from Lindell's paper](https://eprint.iacr.org/2020/300.pdf))? 

For example, the draft states "break robustness" or "attack on robustness" but it is not clear if that means to cause the Collector to output an incorrect result (i.e., attack on correctness) or to prevent honest parties from receiving their output (i.e., attack on GOD).

cjpatton:
Closes #304.

Add the following clarifications to our thread model:

- Privacy assumes the attacker does not control transmission of messages between the honest Clients and Aggregators.

- Robustness assumes the attacker eavesdrops on the network but does not control transmission.

This implies that guaranteed output delivery and malicious robustness are not guaranteed.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -126,6 +126,16 @@ informative:
     seriesinfo: EUROCRYPT 2014
     target: https://link.springer.com/chapter/10.1007/978-3-642-55220-5_35
 
+  GSZ20:
+    title: "Guaranteed Output Delivery Comes Free in Honest Majority MPC"
+    author:
+      - ins: V. Goyal
+      - ins: Y. Song
+      - ins: C. Zhu
+    date: 2020
+    seriesinfo: CRYPTO 2020
+    target: https://link.springer.com/chapter/10.1007/978-3-030-56880-1_22
+
   GKWWY20:
     title: Better concrete security for half-gates garbling (in the multi-instance setting)
     author:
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -126,6 +126,16 @@ informative:
...
 
+  GSZ20:
+    title: "Guaranteed Output Delivery Comes Free in Honest Majority MPC"
+    author:
+      - ins: V. Goyal
+      - ins: Y. Song
+      - ins: C. Zhu
+    date: 2020
+    seriesinfo: CRYPTO 2020
+    target: https://link.springer.com/chapter/10.1007/978-3-030-56880-1_22
```
cjpatton, line 137: @jimouris, @pratiksarkar93 let me know if you have a preferred reference for guaranteed output delivery.

cjpatton: I decided against https://eprint.iacr.org/2020/300 because it's not published anywhere. However RFCs do have references to unpublished work, so we could include it if you think it's better.

jimouris (approved):
Looks good to me, thanks for updating! Agreed that it's better to cite published papers.

divergentdave (approved):

PR #362, opened 2024-06-12
Enforce type hints for vdaf_*.py and all remaining tests

### LINKED ISSUE #59: poc: Enforce types

cjpatton:
Closes #59.

### DIFF
--- a/poc/vdaf_prio3.py
+++ b/poc/vdaf_prio3.py
@@ -1,8 +1,5 @@
 """The prio3 VDAF."""
 
-from typing import Optional, Union
-
-import flp
 import flp_generic
 import xof
 from common import byte, concat, front, vec_add, vec_sub, zeros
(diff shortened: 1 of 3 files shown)


poc/vdaf_prio3.py
```diff
@@ -540,13 +540,10 @@ class Prio3SumVecWithMultiproof(Prio3SumVec):
     # Operational parameters.
     test_vec_name = 'Prio3SumVecWithMultiproof'
 
+    # TODO(issue #361) Enforce that each parameter is an `int` except `field`,
+    # which should be a `type[Field]`.
```
divergentdave, line 544: FYI `issubclass()` is the class-level counterpart of `isinstance()`.

divergentdave (approved):

PR #373, opened 2024-07-02
Use parentheses around notes to RFC editor

divergentdave:
Using square brackets around these notes causes Markdown issues. Rather than try to escape the brackets, I decided to use parentheses instead, because a quick Google search suggests parentheses are more common for these sorts of remarks.


cjpatton (approved):
Completness: There are may also be inlined TODOs or NOTEs for which we'd want to apply the same change.

divergentdave:
All the existing `TODO`s and `NOTE`s just use blockquotes, so I think they're good as-is.

PR #375, opened 2024-07-02
Rename FlpGeneric to FlpBBCGGI19 and IdpfPoplar to IdpfBBCGGI21

cjpatton:
The rationale has two parts.

First, the names follow different conventions: "Generic" in the sense that it applies to any language that can be expressed as an arithmetic circuit, and "Poplar" simply because it was the IDPF from the Poplar paper. This is a bit awkward.

Second, neither scheme is named, so the names that apply best are the papers in which the schemes were described (BBCGGI19 and BBCGGI21 respectively).

The new names only appear in the reference code. IN the document we just refer to them as the "FLP of {{BBCGGI19}}" and IDPF of {{BBCGGI21}} respectively.


poc/tests/test_idpf_bbcggi21.py
```diff
@@ -3,56 +3,58 @@
...
 from common import TEST_VECTOR, from_be_bytes, gen_rand, vec_add
 from field import Field
-from idpf_poplar import IdpfPoplar
+from idpf_bbcggi21 import IdpfBBCGGI21
 from tests.idpf_util import gen_test_vec, test_idpf, test_idpf_exhaustive
 
 
-class TestIdpfPoplar(unittest.TestCase):
+class TestIdpfBBCGGI21(unittest.TestCase):
     def test_idpfpoplar(self) -> None:
```
divergentdave, line 11: nit: rename this too

poc/idpf_bbcggi21.py
```diff
@@ -27,12 +27,11 @@
 CorrectionWordTuple: TypeAlias = tuple[bytes, tuple[Field2, Field2], FieldVec]
 
 
-class IdpfPoplar(Idpf[Field64, Field255]):
+class IdpfBBCGGI21(Idpf[Field64, Field255]):
```
divergentdave, line 30: We should change `test_vec_name` too, below

draft-irtf-cfrg-vdaf.md
```diff
@@ -2078,8 +2082,7 @@ class XofFixedKeyAes128(Xof):
         # Use TurboSHAKE128 to derive a key from the binder string
         # and domain separation tag. Note that the AES key does not
         # need to be kept secret from any party. However, when used
-        # with IdpfPoplar, we require the binder to be a random
-        # nonce.
+        # with an IDPF, we require the binder to be a random # nonce.
```
divergentdave, line 2085: nit: This got rewrapped wrong
```suggestion
        # with an IDPF, we require the binder to be a random nonce.
```

divergentdave (approved):

PR #376, opened 2024-07-03
Prio3: Move field spec from FLP to VDAF

### LINKED ISSUE #328: Have Prio3 specify the field rather than the FLP circuit
> @divergentdave  Looks good. I think we should also add `| PROOFS | 1 |` to tables 9, 10, 11, and 12, right after `Field`, to be clear about the parameter values for the four concrete Prio3 VDAFs (and algorithm IDs) we define.

These tables specify parameters of the circuit; the number of proofs is not a parameter of the circuit. In fact, the gudance here is about modifying one of the circuits, say `SumVec` by changing the field from `Field64` to `Field128`.

What do you think about decoupling `Field` from the circuit and moving it to the `Prio3` variant. There, `PROOFS` is (or should, at least) be specified.

_Originally posted by @cjpatton in https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/325#issuecomment-1906952751_

cjpatton:
Closes #328.

If one wants to use a different field for a circuit other than the default, they must define a subclass with the `field` parameter overwritten. This is a bit clunky, and the need to configure the field is common enough that it's worth reducing the friction.

This commit moves the specification of the field from the validity circuit up to the VDAF. Accordingly, add a table to each subsection of {{prio3-instantiations}} that defines the circuit, field, number of proofs, and XOF.

Other changes:

- Align the FLP syntax with the poc: use `F` as the generic field parameter rather than `Field`.

- In the specification of `FlpBBCGGI19`, use `field` instead of `Field` as a reference to the field object.


draft-irtf-cfrg-vdaf.md
```diff
@@ -2416,10 +2416,10 @@ of proofs.
 ## Construction {#prio3-construction}
 
 This section specifies `Prio3`, an implementation of the `Vdaf` interface
-({{vdaf}}). It has two generic parameters: an `Flp` ({{flp}}) and a `Xof`
-({{xof}}). It also has an associated constant, `PROOFS`,  with a value within
-the range of `[1, 256)`, denoting the number of FLPs generated by the Client
-({{multiproofs}}). The value of `PROOFS` is `1` unless explicitly specified.
+({{vdaf}}). It has three generic parameters: a `Field ({{field}}), an `Flp`
```
divergentdave, line 2419: I think we should mention the requirement that the field be FFT-friendly, and link to the relevant section, since the link was removed from the validity circuit's table of parameters.

divergentdave (approved):

PR #381, opened 2024-07-23
Rename binder string to nonce, and add a length requirement

### LINKED ISSUE #352: Clarify requirements for generating the IDPF binder string
We require the binder string to be generated at random, but we don't prescribe its length or say why it needs to be random:

https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/3ac03623c4680b28d1517af0a16dd513e1d0c310/draft-irtf-cfrg-vdaf.md?plain=1#L3710-L3714

I believe this has to do with how `XofFixedKeyAes128` affects the concrete security of `Poplar1`, but I don't remember the details. Let's make sure this is documented. Furthermore, we need to define the length so that the user knows how many bytes to sample. I'd suggest renaming the binder to `nonce` and adding `Idpf.NONCE_SIZE` as a constraint on `Idpf`. (`Poplar1` may need to be modified accordingly.)

schoppmp:
Fixes #352.


cjpatton (approved):
Works for me, thanks Phillipp!

PR #382, opened 2024-07-30
Make the reference code installable via pip

### LINKED ISSUE #359: Make proof of concept code pip-installable
We should rearrange the directory and module structure of the proof of concept code, and add metadata files (i.e. pyproject.toml or setup.py) to make it installable as a package. This would remove the need for other related drafts to use git submodules and Python import path hacks. I figure we'd want to provide just one top-level module, probably called `vdaf`, and then move module paths under that. Note that this would require moving around files and rewriting a lot of internal import statements. The goal is that other projects should be able to do `pip install git+https://github.com/cfrg/draft-irtf-cfrg-vdaf.git@draft-irtf-cfrg-vdaf-NN#subdirectory=poc`, and then `import vdaf`.

cjpatton:
Closes #359.

Projects can now install the reference implementation via

```
sage -pip install git+https://github.com/cfrg/draft-irtf-cfrg-vdaf@draft-irtf-cfrg-vdaf-11#subdirectory=poc
```

This avoids the need to hack a submodule into the project's import path.

This change also moves `test_vec` to the root project directory so that `pip install` can disambiguate this folder from the actual package, which is called `vdaf_poc` and exists in the same directory. We could alternatively put the package in a distinguished `src` directory, but this would require the user to install the package in order to run tests.


poc/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+requires = ["setuptools >= 61.0.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "vdaf_poc"
+description = "Reference implementation of draft-irtf-cfrg-vdaf. This code is not optimzed for performance or resistance to side-channel attacks."
```
cjpatton, line 6: nit: @divergentdave is this too long / should it have periods in it? [The example I'm following](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata) does not have periods.

.gitignore
```diff
@@ -20,3 +20,4 @@ sagelib
 *.pyc
 *.pyo
 __pycache__
+*.egg-info
```
cjpatton, line 23: @divergentdave please confirm we don't want to check this in.

poc/README.md
```diff
@@ -6,30 +6,47 @@ themesleves.
 
 ## Installation
 
-This code is compatible with SageMath version 9.6.
+This code is compatible with SageMath version 9.6. To install:
 
-In order to run the code you will need to install
-[PyCryptodome](https://pycryptodome.readthedocs.io/en/latest/index.html).
+```
+sage -pip install git+https://github.com/cfrg/draft-irtf-cfrg-vdaf@draft-irtf-cfrg-vdaf-11#subdirectory=poc
```
cjpatton, line 12: Reviewer note: I've confirmed this works at cf1c27aaae0d33007e0211b6fc5b5816b98e5ed8.

poc/vdaf_poc/common.py
```diff
@@ -11,7 +11,7 @@
 TEST_VECTOR = os.environ.get('TEST_VECTOR', 'false').lower() == 'true'
 # The path where test vectors are generated.
 TEST_VECTOR_PATH = os.environ.get('TEST_VECTOR_PATH',
-                                  'test_vec/{:02}'.format(VERSION))
+                                  '../test_vec/{:02}'.format(VERSION))
```
cjpatton, line 14: @divergentdave do you think this will cause anyone any trouble? I  did this to help `pip` disambiguate `vdaf_poc` from `test_vec` as the package to install. Perhaps there's a better way to do this?

poc/pyproject.toml
divergentdave: It looks like this is the tagline that appears on PyPI project pages, and in search results. Thus, I think this might be on the line of "too long". If we want to include something longer, we could use `readme` to point at a file. I wonder if the setup.py `long_description` argument would work here as well, if we wanted to put the bigger blurb inline.

.gitignore
divergentdave: Yes, LGTM.

poc/pyproject.toml
cjpatton: Shortened to "Reference implementation of draft-irtf-cfrg-vdaf" and added a pointer to the README.

poc/tools/plot_prio3_multiproof_robustness.py
```diff
@@ -6,9 +6,9 @@
 
 import matplotlib.pyplot as plt
 
-from field import Field64, Field128
-from flp_bbcggi19 import FlpBBCGGI19
-from vdaf_prio3 import Prio3SumVec
+from vdaf_poc.field import Field64, Field128
+from vdaf_poc.flp_bbcggi19 import FlpBBCGGI19
+from vdaf_poc.vdaf_prio3 import Prio3SumVec
```
divergentdave, line 11: From the `poc` directory, I tried running `sage -python tools/plot_prio3_multiproof_robustness.py`, and it didn't work. Running `sage -pip install -e .` first fixed it, though. I tried uninstalling the editable package, then copying the script up a directory, into `poc`, and running that worked as well. I suppose getting rid of the tools directory would be the simplest solution.

poc/vdaf_poc/common.py
divergentdave: I think this should be fine.

poc/README.md
```diff
@@ -1,35 +1,59 @@
...
+This directory contains a reference implementation of the VDAFs specified in
+[draft-irtf-cfrg-vdaf](https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/).
+It is not intended for production use: the code is not optimized for
+performance or resistance to side-channel attacks. Its primary purpose is to
+generate test vectors.
+
+
+SageMath implementations of VDAFs. This code is used to
 generate test vectors as well as the algorithm definitions in the document
 themesleves.
```
divergentdave, line 12: This paragraph is largely duplicative now

poc/README.md
```diff
@@ -1,35 +1,59 @@
...
+TEST_VECTOR=TRUE sage -python -m unittest --quiet
 ```
 
 Users can also specify a custom path to generate the test vectors in
 environment variable `TEST_VECTOR_PATH`. For example, to generate
 test vectors for Prio3 VDAFs into path `test_vec/00`:
 
 ```
-TEST_VECTOR=TRUE TEST_VECTOR_PATH=test_vec/00 sage -python -m unittest --quiet tests/test_vdaf_prio3.py
+TEST_VECTOR=TRUE TEST_VECTOR_PATH=test_vec/11 sage -python -m unittest --quiet tests/test_vdaf_prio3.py
```
divergentdave, line 58: This doesn't align with the sentence above.
```suggestion
TEST_VECTOR=TRUE TEST_VECTOR_PATH=test_vec/00 sage -python -m unittest --quiet tests/test_vdaf_prio3.py
```

divergentdave (approved):

poc/tools/plot_prio3_multiproof_robustness.py
cjpatton: Moved the plotter to the poc directory. There you can run it without installing the package.

PR #384, opened 2024-08-01
Prio3: Add an implementation note about the output share

### LINKED ISSUE #380: Prio3: Store compressed helper input share in PrepState
Currently, the PrepState for Prio3 is `tuple[list[F], Optional[bytes]]`, consisting of the output share and the recomputed joint randomness seed. We could improve the storage requirements for the helper if the helper instead stored its compressed input share seed, and then re-expanded and re-truncated it into an output share during prep_next(). I think the leader should still store its output share in the PrepState, as it may be shorter than the leader input share, depending on the Prio3 instance. This would be quasi-wire breaking, in that it would change the format of an aggregator's persistent storage, but wouldn't affect messages between parties.

cjpatton:
Closes #380.

It may be desirable to store the measurement share seed rather than the output share.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -2746,6 +2746,14 @@ same joint randomness seed before accepting their output shares. To do so, they
 exchange their parts of the joint randomness along with their shares of
 verifier(s).
 
+Implementation note: the preparation state for Prio3 includes the output share
+that will be released once preparation is complete. In some situations, it may be
+necessary for the Aggregator to encode this state as bytes and store it for
+retrieval later on. For all but the first Aggregator, it is possible to save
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -2746,6 +2746,14 @@ same joint randomness seed before accepting their output shares. To do so, they
 exchange their parts of the joint randomness along with their shares of
 verifier(s).
 
+Implementation note: the preparation state for Prio3 includes the output share
+will be released once preparation is complete. In some situations, it may be
```
divergentdave, line 2750:
```suggestion
that will be released once preparation is complete. In some situations, it may be
```

divergentdave (approved):

PR #386, opened 2024-08-02
poc: Ensure all test failures are enforced by CI

### LINKED ISSUE #327: Testing improvements

cjpatton:
Closes #327.

For some of our tests, when the test fails, it outputs something to stdout:

```python
if not test_succeeded:
    print('test failed!')
```

This does not cause `sage python -m unittest` to fail, and so will be missed by CI. To address this, replace these with an assertion:

```python
assert test_succeeded
```

In fact, this change goes one step further: encapsulate the test in a `unittest.TestCase` so that the test runner can provide more useful feedback:

```python
class MyTest(unittest.TestCase):
    def test(self):
       ...
       self.assertTrue(test_succeeded)
```

This required a bit of refactoring:

1. Create a new module, `vdaf_poc/test_utils.py`, that defines a test harness for instances of `Vdaf` and `FlpBBCGGI19`.

2. Likewise, move `vdaf_poc/idpf_util.py` to a test harness in `tests/test_idpf_bbcggi21.py`.

Finally, replace assertions in unit tests with `unittest`-style assertions.


poc/tests/test_daf.py
cjpatton, line 1: Reviewer note: this is definitely easier to review with whitespace hidden.

divergentdave (approved):
I checked the code coverage as well, and it looks good. (to make sure all the test code is properly discovered and run, after being rearranged)

PR #399, opened 2024-08-13
Add diagrams for FLP and rework overview/extensions

### LINKED ISSUE #257: A diagram to show the relationship of different terms in FLP
There are many terms discussed in the generic FLP section (https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/draft-irtf-cfrg-vdaf.md#a-general-purpose-flp-flp-generic):
Gadgets, Validity Circuits, Proof, Prover, Verifier, etc. A diagram shows the relationship between them would make the text easier to follow.

cjpatton:
Closes #257.
Partially addresses #226.

This commit includes editorial improvements for `FlpBBCGGI19`:

1. Add diagrams to show concepts relate to one another: validity circuit, gadgets, encoded measurement, joint, prove, and query randomness, and so on.

2. Improve the overview section, building up each of the concepts one by one.

3. Do an editorial pass of the extensions section, some of which is a little out of date. Also, add a reference for NTT.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -3157,107 +3157,107 @@ struct {
 
 ## The FLP of {{BBCGGI19}} {#flp-bbcggi19}
 
-This section describes an FLP based on the construction from in {{BBCGGI19}},
-Section 4.2. We begin in {{flp-bbcggi19-overview}} with an overview of their proof
-system and the extensions to their proof system made here. The construction is
-specified in {{flp-bbcggi19-construction}}.
-
-> OPEN ISSUE Chris Wood points out that the this section reads more like a paper
-> than a standard. Eventually we'll want to work this into something that is
-> readily consumable by the CFRG.
+This section describes an FLP based on the construction from {{BBCGGI19}},
+Section 4.2. We begin in {{flp-bbcggi19-overview}} with an overview of their
+proof system and some extensions to it. The construction is specified in
+{{flp-bbcggi19-construction}}.
 
 ### Overview {#flp-bbcggi19-overview}
 
-In the proof system of {{BBCGGI19}}, validity is defined via an arithmetic
-circuit evaluated over the encoded measurement: if the circuit output is zero,
-then the measurement is deemed valid; otherwise, if the circuit output is
-non-zero, then the measurement is deemed invalid. Thus the goal of the proof
-system is merely to allow the verifier to evaluate the validity circuit over
-the measurement. For our application ({{prio3}}), this computation is
-distributed among multiple Aggregators, each of which has only a share of the
-measurement.
+Conventional zero-knowledge proof systems involve two parties:
+
+* The prover, who holds a measurement and generates a proof of the
+  measurement's validity
+
+* The verifier who holds an encryption of, or commitment to, the measurement
+  and checks the proof
+
+Our proof system is much the same, except the verifier is split across multiple
+Aggregators, each of which has a secret share of the measurement rather than an
+encryption of it.
+
+Validity is defined in terms of an arithmetic circuit evaluated over the
+measurement. The inputs to this circuit are elements of a finite field that
+comprise the encoded measurement; the gates of the circuit are multiplication,
+addition, and subtraction operations; and the output of the circuit is a single
+field element. If the value is zero, then the measurement is deemed valid;
+otherwise, if the output is non-zero, then the measurement is deemed invalid.
+
+For example, the simplest circuit specified in this document is the following
+({{prio3count}}):
+
+~~~
+C(x) = x * (x-1)
+~~~
+
+This circuit contains one subtraction gate (`x - 1`) and one multiplication
+gate (`x * (x - 1)`). Observe that `C(x) = 0` if and only if `x in range(2)`.
+
+Our goal is to allow each Aggregator, who holds a secret share of `x`, to
+correctly compute a secret share of `C(x)`. This allows the Aggregators to
+determine validity by combining their shares of the output.
 
 Suppose for a moment that the validity circuit `C` is affine, meaning its only
-operations are addition and multiplication-by-constant. In particular, suppose
-the circuit does not contain a multiplication gate whose operands are both
-non-constant. Then to decide if a measurement `x` is valid, each Aggregator
-could evaluate `C` on its share of `x` locally, broadcast the output share to
-its peers, then combine the output shares locally to recover `C(x)`. This is
-true because for any `SHARES`-way secret sharing of `x` it holds that
+operations are addition, subtraction, and multiplication-by-constant. (The
+circuit above is non-affine because it contains a multiplication gate with
+two non-constant inputs.) Then each Aggregator can compute its share locally, since
 
 ~~~
 C(x_shares[0] + ... + x_shares[SHARES-1]) =
     C(x_shares[0]) + ... + C(x_shares[SHARES-1])
 ~~~
 
-(Note that, for this equality to hold, it may be necessary to scale any
-constants in the circuit by `SHARES`.) However this is not the case if `C` is
-not-affine (i.e., it contains at least one multiplication gate whose operands
-are non-constant). In the proof system of {{BBCGGI19}}, the proof is designed to
-allow the (distributed) verifier to compute the non-affine operations using only
-linear operations on (its share of) the measurement and proof.
+(Note that, for this equality to hold, it is necessary to scale any addition of
+a constant in the circuit by `1/SHARES`.) However, this is not the case if `C`
+contains multiplication gates with two non-constant inputs. Thus our goal is to
+transform these multiplication gates into computations on secret shared data
+that each Aggregator can perform locally.
 
-To make this work, the proof system is restricted to validity circuits that
-exhibit a special structure. Specifically, an arithmetic circuit with "G-gates"
-(see {{BBCGGI19}}, Definition 5.2) is composed of affine gates and any number of
-instances of a distinguished gate `G`, which may be non-affine. We will refer to
-this class of circuits as 'gadget circuits' and to `G` as the "gadget".
+The key idea is to have the prover construct a polynomial `p` such that `p(j)`
+is equal to the output of the `j`-th multiplication gate. Polynomial evaluation
+is fully linear, which means the coefficients of the polynomial can be secret
+shared in a way that allows each Aggregator to compute a share of `p(j)` for
+any `j`. These intermediate results can then be combined with the affine
+operations of the validity circuit to produce the final output.
 
-As an illustrative example, consider a validity circuit `C` that recognizes the
-set `L = set([0], [1])`. That is, `C` takes as input a length-1 vector `x` and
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -3153,107 +3162,106 @@ struct {
...
 
-To make this work, the proof system is restricted to validity circuits that
-exhibit a special structure. Specifically, an arithmetic circuit with "G-gates"
-(see {{BBCGGI19}}, Definition 5.2) is composed of affine gates and any number of
-instances of a distinguished gate `G`, which may be non-affine. We will refer to
-this class of circuits as 'gadget circuits' and to `G` as the "gadget".
+Validity is defined in terms of an arithmetic circuit evaluated over the
+measurement. The inputs to this circuit are elements of a finite field that
+comprise the encoded measurement; the gates of the circuit are multiplication,
+addition, and negation operations; and the output of the circuit is a single
```
divergentdave, line 3187: This formulation of a circuit seems unusual, I think either 'multiplication and addition' or 'multiplication, addition, and subtraction' would be more conventional.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3153,107 +3162,106 @@ struct {
...
+inputs.) Then each Aggregator can compute its share locally, since
 
 ~~~
-Mul(left, right) = left * right
+C(x_shares[0] + ... + x_shares[SHARES-1]) =
+    C(x_shares[0]) + ... + C(x_shares[SHARES-1])
 ~~~
 
-The validity circuit can then be rewritten in terms of `Mul` like so:
+(Note that, for this equality to hold, it is necessary to scale any constants
```
divergentdave, line 3216: Constants used in constant multiplications are fine as-is, but constants that get added need to be scaled, since this is one way of replacing them with additive shares of themselves.
```suggestion
(Note that, for this equality to hold, it is necessary to scale any addition of a constant
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3153,107 +3162,106 @@ struct {
...
+in the circuit by `1/SHARES`.) However, this is not the case if `C` is
```
divergentdave, line 3217: nit: typo
```suggestion
in the circuit by `1/SHARES`.) However, this is not the case if `C` has
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3153,107 +3162,106 @@ struct {
...
+multiplication gates with non-constant inputs. Thus or goal is to transform
```
divergentdave, line 3218: nit: typo
```suggestion
multiplication gates with non-constant inputs. Thus our goal is to transform
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3153,107 +3162,106 @@ struct {
...
+is fully linear, which means the coefficients of the polynomial can be secret
+shared in a way that allows each Aggregator to compute a share of `p(j)` for
+any `j`. These intermediate results can then be combined with the affine
+operations of the validity circuit to produce the final output.
+
+Applying this idea to the example circuit `C` above:
+
+1. The Client, given its measurement `x`, constructs the lowest degree
+   polynomial `p` for which `p(0) = x * (x - 1)`. It sends secret shares of `x`
+   and the coefficients of `p` to each of the Aggregators.
```
divergentdave, line 3233: This isn't quite right, because the wire polynomials evaluated at the first point produce blinding values. (and the gadget polynomial evaluated at the first point outputs some function of the blinding values) I think we should probably mention or allude to the blinding here, since it's necessary for zero knowledge, and it's not mentioned below.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3153,107 +3162,106 @@ struct {
...
-To make this work, the proof system is restricted to validity circuits that
-exhibit a special structure. Specifically, an arithmetic circuit with "G-gates"
-(see {{BBCGGI19}}, Definition 5.2) is composed of affine gates and any number of
-instances of a distinguished gate `G`, which may be non-affine. We will refer to
-this class of circuits as 'gadget circuits' and to `G` as the "gadget".
+Validity is defined in terms of an arithmetic circuit evaluated over the
+measurement. The inputs to this circuit are elements of a finite field that
+comprise the encoded measurement; the gates of the circuit are multiplication,
+addition, and negation operations; and the output of the circuit is a single
+field element. If the value is zero, then the measurement is deemed valid;
```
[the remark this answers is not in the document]
cjpatton: This is introduced in the extensions section, in the very first paragraph. Do you think we should either introduce it here or say more about it there? I figured we should keep the overview as conceptually close to the paper as possible.

draft-irtf-cfrg-vdaf.md
cjpatton: Hmm, I think of addition, multiplication, and negation as analogues of OR, AND, and NOT respectively for boolean circuits, which to me is more intuitive. What about stating this analogy here, and maybe me mentioning that subtraction is negation followed by addition?

draft-irtf-cfrg-vdaf.md
cjpatton: Good catch.

draft-irtf-cfrg-vdaf.md
cjpatton: Nice catch :) Please take another look: I used `s` to denote the blinding factor, which is short for `seed`, the word we use in the actual specification.

cjpatton:
Thanks for the careful review here! Note that I've rebased to resolve conflicts, while at it I decided to squash as well.

draft-irtf-cfrg-vdaf.md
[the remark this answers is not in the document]
divergentdave: I missed it on my first read (probably because it wasn't part of the diff), but this is addressed as the fourth extension in 7.3.1.1. LGTM.

draft-irtf-cfrg-vdaf.md
divergentdave: I don't think we need to draw direct parallels to boolean circuits. Besides, if we were to project a boolean circuit into an arithmetic circuit, logical negation would probably be implemented with (1 - x), so I think the analogy doesn't hold up.

Looking at the 2019 Prio paper, it says in section 7.1 that arithmetic circuits over a ring R "may contain addition, subtraction and multiplication gates, as well as constants from R." The Wikipedia page for [Arithmetic circuit complexity](https://en.wikipedia.org/wiki/Arithmetic_circuit_complexity) says that circuits are made up of input gates, sum gates, and product gates.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3157,107 +3157,108 @@ struct {
...
+
+Applying this idea to the example circuit `C` above:
+
+1. The Client, given its measurement `x`, constructs the lowest degree
+   polynomial `p` for which `p(0) = s` and `p(1) = x * (x - 1)`, where `s` is a
+   random blinding value generated by the Client. (The blinding value is to
+   protect the privacy of the measurement.) It sends secret shares of `x` and
+   the coefficients of `p` to each of the Aggregators.
+
+1. Each Aggregator locally computes and broadcasts its share of `p(0)`, which
```
divergentdave, line 3232: This needs to be updated to match the above changes.
```suggestion
1. Each Aggregator locally computes and broadcasts its share of `p(1)`, which
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3157,107 +3157,108 @@ struct {
...
+   is equal to its share of `C(x)`.
+
+In fact, our FLP is slightly more general than this. We can replace the
+multiplication gate with any non-affine sub-circuit and apply the same idea.
+For example, in {{prio3sum}}, the validity circuit uses the following
+sub-circuit multiple times:
 
 ~~~
-C(x[0]) = Mul(x[0], x[0]) - x[0]
+Range2(x) = x * (x -1) = x^2 - x
```
divergentdave, line 3241: nit: formatting
```suggestion
Range2(x) = x * (x - 1) = x^2 - x
```

draft-irtf-cfrg-vdaf.md
cjpatton: I'm not sure I aqgree, but I don't feel strongly either way. I replaced "negation" with "subtraction".

draft-irtf-cfrg-vdaf.md
```diff
@@ -3157,107 +3157,107 @@ struct {
...
+C(x) = x * (x-1)
 ~~~
 
-This polynomial recognizes `L` because `x[0]^2 = x[0]` is only true if `x[0] ==
-0` or `x[0] == 1`. Notice that the polynomial involves a non-affine operation,
-`x[0]^2`. In order to apply {{BBCGGI19}}, Theorem 4.3, the circuit needs to be
-rewritten in terms of a gadget that subsumes this non-affine operation. For
-example, the gadget might be multiplication:
+This circuit contains one subtraction gate (`x -1`) and one multiplication
+gate (`x * (x -1)`). Observe that `C(x) = 0` if and only if `x in range(2)`.
```
divergentdave, line 3194: nit: formatting
```suggestion
This circuit contains one subtraction gate (`x - 1`) and one multiplication
gate (`x * (x - 1)`). Observe that `C(x) = 0` if and only if `x in range(2)`.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3157,107 +3157,107 @@ struct {
...
+
+Our goal is to allow each Aggregator, who holds a secret share of `x`, to
+correctly compute a secret share of `C(x)`. This allows the Aggregators to
+determine validity by combining their shares of the output.
+
+Suppose for a moment that the validity circuit `C` is affine, meaning its only
+operations are addition, subtraction, and multiplication-by-constant. (The
+circuit above is non-affine because it contains a multiplication gate with
+non-constant inputs.) Then each Aggregator can compute its share locally, since
```
divergentdave, line 3203: Note that one constant input and one non-constant input is affine
```suggestion
two non-constant inputs.) Then each Aggregator can compute its share locally, since
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3157,107 +3157,107 @@ struct {
...
 ~~~
-Mul(left, right) = left * right
+C(x_shares[0] + ... + x_shares[SHARES-1]) =
+    C(x_shares[0]) + ... + C(x_shares[SHARES-1])
 ~~~
 
-The validity circuit can then be rewritten in terms of `Mul` like so:
+(Note that, for this equality to hold, it is necessary to scale any addition of
+a constant in the circuit by `1/SHARES`.) However, this is not the case if `C`
+contains multiplication gates with non-constant inputs. Thus our goal is to
```
divergentdave, line 3212: Same as above
```suggestion
contains multiplication gates with two non-constant inputs. Thus our goal is to
```

divergentdave (approved):

PR #400, opened 2024-08-13
Use lower case after colon

### LINKED ISSUE #396: Capitalization after colon
Sometimes we capitalize the first letter after a colon, other times we don't. Let's be consistent. (The latter is probably the correct one.)

cjpatton:
Closes #396.


draft-irtf-cfrg-vdaf.md
```diff
@@ -414,13 +414,13 @@ security considerations for DAFs and VDAFs.
 * Define Prio3MultihotCountVec, a variant of Prio3 for aggregating bit vectors
   with bounded weight.
 
-* FLP: Allow the output of the circuit to be a vector. This makes it possible
+* FLP: allow the output of the circuit to be a vector. This makes it possible
```
divergentdave, line 417: I lean towards leaving these occurrences in the changelog as-is, because I view the sentence as starting with "Allow", for example, and the "FLP:" that comes before it as a label, outside of the sentence structure.

divergentdave (approved):

draft-irtf-cfrg-vdaf.md

PR #401, opened 2024-08-13
Refer to an instance, not the class in msssage encoding sections

cjpatton:
To define message encoding, we need to be able to refer to attributes of the VDAF, e.g., the value of `SHARES`. Previously we referred to the class in this section, e.g., `Prio3`, but since getting rid of the class-factor pattern, it is now necessary to refer to an instance of the class.

To make this clear, replace upper case `Prio3` with lowercase `prio3`, to imply that we an instance of a sub-class of `Vdaf` rather than the sub-class itself.


divergentdave (approved):
Other candidates:

- `Prio3.xof.SEED_SIZE` on line 3027
- `Prio3.field.ENCODED_SIZE` on line 3028
- `Prio3.flp.JOINT_RAND_LEN > 0` on line 3044
- `Poplar1.xof.SEED_SIZE` on line 4699
- `Poplar1.idpf.field_inner` on line 4700
- `Poplar1.idpf.field_leaf` on line 4701

Actually, these last two `Fi`/`Fl` aliases should be for `poplar1.idpf.field_inner.ENCODED_SIZE` and `poplar1.idpf.field_leaf.ENCODED_SIZE`, now that I'm looking at it.

PR #405, opened 2024-08-20
idpf: Use XofTurboShake128 for the leaf nodes (*)

cjpatton:
Based on #404.
Closes #341.

This is a safer choice for ensuring the IDPF is extracatible.

PR #406, opened 2024-08-20
Improve circuit for `Prio3Sum` (*)

### LINKED ISSUE #306: Generalize range check in Prio3Sum

cjpatton:
Based on #405.
Closes #306.

Replace `Sum` with `SumOfRangeCheckedInputs`. The new circuit has two improvements:

1. No joint randomness is used, so it's safe to use a smaller field.

2. The largest measurement can be any positive integer, not just a power of two.

Update `Prio3Sum` accordingly by switching from `Field128` to `Field64`.


draft-irtf-cfrg-vdaf.md
```diff
@@ -3663,37 +3663,56 @@ way. The parameters for this circuit are summarized below.
...
+checked measurement. Let
+
+* `bits = max_measurement.bit_length()`, the number of bits needed to encode
+  the largest valid measurement
+
+* `offset = 2**bits - 1 - max_measurement`
+
+The range checked measurement is the bit-encoding of `offset` plus the
+measurement. Observe that only measurements between `0` and `max_measurement`
+inclusive can be encoded this way with `bits` bits.
```
divergentdave, line 3690: Same as below. My two main issues with the existing explanation are that it elides that the "reported measurement" is encoded as a bit vector as well, and the claim that the 'range checked measurement' ensures the measurement is "between `0` and `max_measurement`" is too strong, since that could represent a measurement of `-offset` as well.

```suggestion
The range check is accomplished by encoding the measurement as a bit vector,
encoding the measurement plus an offset as a bit vector, then checking that the
two encoded integers are consistent. Let

* `bits = max_measurement.bit_length()`, the number of bits needed to encode
  the largest valid measurement

* `offset = 2**bits - 1 - max_measurement`

The first bit-encoded integer is the measurement itself. Note that only
measurements between `0` and `2**bits - 1` can be encoded this way with
`bits` bits. The second bit-encoded integer is the sum of the measurement
and `offset`. Observe that only measurements between `-offset` and
`max_measurement` inclusive can be encoded this way with `bits` bits.
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3663,37 +3663,56 @@ way. The parameters for this circuit are summarized below.
...
 
-The validity circuit is denoted `Sum`. The measurement is encoded as a
-length-`bits` vector of field elements, where the `l`th element of the vector
-represents the `l`th bit of the summand:
+The circuit, denoted `Sum`, first checks that each entry of this bit vector is a
+one or a zero. It then decodes it and subtracts it from `offset` plus the
+reported value. Since the range checked measurement is in the correct range,
+equality implies that the reported measurement is as well.
```
divergentdave, line 3695: Continued
```suggestion
The circuit, denoted `Sum`, first checks that each entry of both bit vectors is a
one or a zero. It then decodes both the measurement and the offset
measurement, and subtracts `offset` from the latter. It then checks if these two
values are equal. Since both the measurement and the measurement plus `offset`
are in the same range of `0` to `2**bits - 1`, this means that the measurement
itself is between `0` and `max_measurement`.
```

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -1118,9 +1045,10 @@ def __init__(self, field: type[F], max_measurement: int):
...
-        The range checked measurement is the bit-encoding of `offset` plus the
-        measurement. Observe that only measurements in at most
-        `max_measurement` can be encoded with `bits` bits.
+        The range checked measurement is the bit-encoding of `offset`
+        plus the measurement. Observe that only measurements between `0`
+        and `max_measurement` incluisve can be encoded this way with
+        `bits` bits.
 
         To do the range check, the circuit first checks that each
         entry of this bit vector is a one or a zero. It then decodes
```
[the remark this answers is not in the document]
cjpatton: Done.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3663,37 +3663,60 @@ way. The parameters for this circuit are summarized below.
...
+* `bits = max_measurement.bit_length()`, the number of bits needed to encode
+  the largest valid measurement
+
+* `offset = 2**bits - 1 - max_measurement`
+
+The first bit-encoded integer is the measurement itself. Note that only
+measurements between `0` and `2**bits - 1` can be encoded this way with
+`bits` bits. The second bit-encoded integer is the sum of the measurement
+and `offset`. Observe that only measurements between `-offset` and
+`max_measurement` inclusive can be encoded this way with `bits` bits.
```
cjpatton, line 3692: @divergentdave if true, this is not the intent of the construction, and it contradicts the statement above: 

>  Each measurement is an integer in `range(max_measurement+1)`, where associated parameter. `max_measurement` is an associated parameter equal to the largest valid measurement.

cjpatton (commented):
I took all your suggestions and rebased. There's one thing I'm confused about: see my comment.

draft-irtf-cfrg-vdaf.md
divergentdave: This is referring specifically to the second encoded integer, but not the first encoded integer. We only get `range(max_measurement+1)` when combining the two, and checking they're consistent. Maybe this would be clearer?

> Observe that this sum can only be encoded this way if it is between `0` and `2**bits - 1`, which implies that the measurement is between `-offset` and `max_measurement`.

cjpatton: Ah ok, that makes more sense! Yes that's clearer.

draft-irtf-cfrg-vdaf.md
```diff
@@ -3663,37 +3663,61 @@ way. The parameters for this circuit are summarized below.
...
+* `bits = max_measurement.bit_length()`, the number of bits needed to encode
+  the largest valid measurement
+
+* `offset = 2**bits - 1 - max_measurement`
+
+The first bit-encoded integer is the measurement itself. Note that only
+measurements between `0` and `2**bits - 1` can be encoded this way with
+`bits` bits. The second bit-encoded integer is the sum of the measurement and
+`offset`. Observe that this sum can only be encoded this way if it is between 0
+and 2**bits - 1, which implies that the measurement is between `-offset` and
```
divergentdave, line 3692: nit: consistency
```suggestion
`offset`. Observe that this sum can only be encoded this way if it is between `0`
and `2**bits - 1`, which implies that the measurement is between `-offset` and
```

divergentdave (approved):

PR #407, opened 2024-08-20
Prio3: Double the length of the joint randomness seed (*)

### LINKED ISSUE #319: XofTurboShake128: Consider bumping SEED_SIZE from 16 to 32 bytes
When thinking about Prio3 robustness (ia.cr/2023/130, Theorem 1) we (err, I, really) have been ignoring the term that depends on the size of the seed: $(q_\text{RG} + {q_{\text{Prep}}}^2) / 2^{\kappa -1}$

* $q_\text{RG}$ - the number of random oracle queries (XofTurboShake128 is modeled as an RO)
* $q_\text{Prep}$ - the number of (potentially malicious) reports consumed in a given DAP task
* $\kappa$ - the seed size (16 bytes for XofTurboShake128)

As @albertpl pointed out in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/318#discussion_r1416669600, the bound becomes vacuous after $q_\text{Prep}=2^{63.5}$ reports (since $(2^{63.5})^2 = 2^{\kappa -1}$).

While it's doubtful that we'd ever consume so many reports in a single task, it's not a bad idea to have a more conservative safety margin. 32 bytes seems like a reasonable choice.

cjpatton:
Based on #406.
Closes #319.

A robustness attacker can try to find two distinct reports that derive the joint randomness seed by an offline birthday attack. By increasing the seed size from 16 to 32 bytes, we ensure this probability is negligible (in the random oracle model) even when the attacker can do a lot of computation (make many random oracle queries).


divergentdave:
Wouldn't we need to make joint randomness hints 32 bytes as well? I think a similar offline birthday attack could be mounted against either seed derivation, whether it be the input share => joint randomness hint RO call, or the joint randomness hints => joint randomness seed RO call.

cjpatton:
> Wouldn't we need to make joint randomness hints 32 bytes as well? I think a similar offline birthday attack could be mounted against either seed derivation, whether it be the input share => joint randomness hint RO call, or the joint randomness hints => joint randomness seed RO call.

@hannahdaviscrypto can correct me if I'm wrong, but I believe the reasoning here is that you have to find a collision on the entire vector of hints, of which there are at least 2 (`SHARES >= 2`). Since the seed sizes is `16`, that's at least `2*16` bytes. Does this make sense or am I missing something?

divergentdave:
You could still hold one input share constant, and then search for a collision in the other hint. That will randomize your measurement, though. I'm not sure whether it's feasible to mount an attack starting from such a joint randomness collision, with two different measurements. Intuitively, I think it would make sense to account for random oracle collision on both queries to be safe.

cjpatton:
> You could still hold one input share constant, and then search for a collision in the other hint. That will randomize your measurement, though. I'm not sure whether it's feasible to mount an attack starting from such a joint randomness collision, with two different measurements. Intuitively, I think it would make sense to account for random oracle collision on both queries to be safe.

Well, you still have to cause a collision on the first hint, right? You have to get lucky twice: you first find a collision on the first hint (i.e., part), then you fix that measurement share and blind, then you find the measurement share and blind for the other aggregator that causes a collision for the second hint.

divergentdave:
Hmm, good points, I had forgotten that the attacker can search by just changing the blind, not necessarily the input share. I'm also not sure at the moment how you would build an entire robustness attack from a birthday attack on the joint randomness seed. However, I still don't see why the same attack pattern couldn't build on a birthday attack on one joint randomness part. If you keep all inputs for one hint fixed, then the birthday attack on the other hint has a cost of $2^{128/2}$ queries. If you do two separate birthday attacks on both hints as you say, then that has a cost of $2^{1+128/2}$ queries, which is not much better.

hannahdaviscrypto:
> I think a similar offline birthday attack could be mounted against either seed derivation, whether it be the input share => joint randomness hint RO call, or the joint randomness hints => joint randomness seed RO call.

Making the joint randomness seed 32 bytes covers the second case here, and a collision in the first case only matters if all hints collide, as Chris said. So the advantage of an attack that searches for collisions one-at-a-time is going to be on the order of q^4/2^256 for two aggregators, and drop more as the number of aggregators increases. That said, the attack wouldn't have to randomize the measurement share; it could fix the nonces and measurement shares and just search among blinds to find collisions (as long as we have three seeds, not one as in the other proposed change).

cjpatton:
> > I think a similar offline birthday attack could be mounted against either seed derivation, whether it be the input share => joint randomness hint RO call, or the joint randomness hints => joint randomness seed RO call.
> 
> Making the joint randomness seed 32 bytes covers the second case here, and a collision in the first case only matters if all hints collide, as Chris said. So the advantage of an attack that searches for collisions one-at-a-time is going to be on the order of q^4/2^256 for two aggregators, and drop more as the number of aggregators increases. That said, the attack wouldn't have to randomize the measurement share; it could fix the nonces and measurement shares and just search among blinds to find collisions (as long as we have three seeds, not one as in the other proposed change).

Why $q^4$? This being the case, we definitely need to bump the joint randomness hints. At which point I wonder if we should just bump all seeds to 32 bytes, for consistency if nothing else.

cjpatton:
Answering my own question: $q^4$ because we need two collisions to happen.

Given this observation, I think we'd be better off just bumping the seed size across the board.

PR #409, opened 2024-08-27
Prio3: Merge the measurement and proof share seed (*)

### LINKED ISSUE #185: Prio3: Consider merging helper shares and blinds into a single seed
We currently use separate seeds for the helper's measurement share, proof share, and blind. It is likely safe to use the same PRG seed for all three.

The upshot is that this reduce bandwdith and simplify the Prio3 spec. However, it is not immediately clear how this change would impact known security analysis. See comments on https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/129 for discussion.

cjpatton:
Stacked on #408.
Closes #185.

Replace the measurement and proof share seeds in the helper's input share with a single seed from which both the measurement and proof share are derived. Usage of the same seed for both shares is safe as long as we have domain separation between them.


draft-irtf-cfrg-vdaf.md
```diff
@@ -2647,9 +2638,9 @@ The steps in this method are as follows:
 1. Encode each measurement and shares of each proof into an input share
 
 Notice that only one pair of measurement and proof(s) share (called the
-"leader" shares above) are vectors of field elements. The other shares
-(called the "helper" shares) are represented instead by XOF seeds, which
-are expanded into vectors of field elements.
+"leader" shares above) are vectors of field elements. The other shares (called
+the "helper" shares) are represented instead by an XOF, which is expanded into
```
divergentdave, line 2642: nit: missing a word
```suggestion
the "helper" shares) are represented instead by an XOF seed, which is expanded into
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -2981,15 +2967,16 @@ def expand_input_share(
             list[F],
             list[F],
             Optional[bytes]]:
-    (meas_share, proofs_share, k_blind) = input_share
     if agg_id > 0:
-        assert isinstance(meas_share, bytes)
```
divergentdave, line 2986: FYI, these assertions served the secondary purpose of narrowing the inferred types, so we didn't need to cast from a union type to one of its arms. We could do the same here with the following, to simplify a bit.

```python
if agg_id > 0:
    assert len(input_share) == 2
    # destructure without cast()
    ...
else:
    assert len(input_share) == 3
    # destructure without cast()
    ...
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3104,8 +3091,7 @@ as follows:
```
divergentdave, line 3091: The text above about helper shares should be updated to say "an XOF seed" instead of "XOF seeds".

poc/vdaf_poc/vdaf_prio3.py
```diff
@@ -438,15 +423,16 @@ def expand_input_share(
                 list[F],
                 list[F],
                 Optional[bytes]]:
-        (meas_share, proofs_share, k_blind) = input_share
         if agg_id > 0:
-            assert isinstance(meas_share, bytes)
```
divergentdave, line 443: See above

draft-irtf-cfrg-vdaf.md
cjpatton: Done.

draft-irtf-cfrg-vdaf.md
cjpatton: Done.

draft-irtf-cfrg-vdaf.md
cjpatton: Done!

cjpatton:
Oof, I'm going to rebase so this is hopefully easier to review.

divergentdave (approved):

cjpatton:
Squahed.

Issue #412, opened 2024-09-05
@armfazh: draft 11: In prose, write `[start, stop)` instead of `range(start, stop)`


cjpatton:

cjpatton:
I'm not sure I like this idea. I would like to have different notation for math and code, but the only thing we have for either is back ticks, e.g., `x`. It would be nice if everything in backticks was code.

PR #413, opened 2024-09-05
Some editorial suggestions from @armfazh

cjpatton:
* Say "linear" instead of "affine"
* Be consistent about range notation. Pick `range()`, but see #412.


divergentdave:
I lean towards keeping "affine" when talking about circuits and gates because it's the language our references use, and it seems to be fairly common in ZK writ large. For example, [here's a ZK library with an affine circuit abstraction](https://hackage.haskell.org/package/arithmetic-circuits-0.2.0/docs/Circuit-Affine.html). I have found one reference to a "linear arithmetic circuit", and it was clear that all gates were of the form "ax + by". Note that our classic Range2 gadget, "x*(x - 1)", is affine, but not linear, by these definitions, since it includes a subtraction of a constant.

Maybe we could say "affine arithmetic circuit" in places where we currently say "affine circuit", to get the gist across more easily, while retaining precision?

cjpatton:
> I lean towards keeping "affine" when talking about circuits and gates because it's the language our references use, and it seems to be fairly common in ZK writ large. For example, [here's a ZK library with an affine circuit abstraction](https://hackage.haskell.org/package/arithmetic-circuits-0.2.0/docs/Circuit-Affine.html). I have found one reference to a "linear arithmetic circuit", and it was clear that all gates were of the form "ax + by". Note that our classic Range2 gadget, "x*(x - 1)", is built with one multiplication gate and subcircuits that are affine, but not linear, by these definitions, since one includes a subtraction of a constant.
> 
> Maybe we could say "affine arithmetic circuit" in places where we currently say "affine circuit", to get the gist across more easily, while retaining precision?

Works for me!

draft-irtf-cfrg-vdaf.md
```diff
@@ -3302,7 +3303,7 @@ high probability.
 The second extension implemented by our FLP allows the validity circuit to
 contain multiple gadget types. (This generalization was suggested in
 {{BBCGGI19}}, Remark 4.5.) This provides additional flexibility for designing
-circuits by allowing multiple, non-affine sub-circuits. For example, the
+circuits by allowing multiple, non-afine sub-circuits. For example, the
```
divergentdave, line 3306: nit: typo
```suggestion
circuits by allowing multiple, non-affine sub-circuits. For example, the
```

divergentdave (approved):

PR #421, opened 2024-09-06
Add defenses-in-depth section

cjpatton:
Partially addresses #418.

Note defenses against cross protocol attacks resulting from weak entropy or weak parameters. Also, acknowledge the possibility of cross protocol attacks within a circuit family (no binding of the circuit parameters) and the blast radius of weak parameters (all deployed tasks).

PR #422, opened 2024-09-09
poc: Add application context string (*)

### LINKED ISSUE #418: Bind application context to VDAF computations
Consider the risk that, for a specific variant of Prio3, an attacker does enough offline computation that it can efficiently find  invalid reports that get accepted by the aggregators. We know this attack is feasible for a sufficiently small finite field and/or number of proofs: when the circuit uses joint randomness, the attacker can search for input shares that derive joint randomness that forces the circuit to accept an invalid measurement. (See #311 for details.)

While we believe we've mitigated this risk for standardized variants, like Prio3SumVec, there is still a possibility of non-standardized variants getting deployed that haven't. The draft should, at a minimum, acknowledge this risk. We could also modify the spec in order to reduce the blast radius of this attack as much as possible.

One idea that's been suggested (by @junyechen1996 and @kunal-talwar) is to add the [DAP task ID](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.4) to the transcript hash for the joint randomness derivation. This way all of the offline computation carried out by the attacker is only useful for a specific DAP task. This is similar to [salting a password hash](https://en.wikipedia.org/wiki/Salt_(cryptography)) in order to ensure that whatever computation the attacker does in advance, like building [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table), is specific to that user the attacker wants to impersonate.

More generally, we could modify the VDAF syntax in order to allow binding the computation to its application context. There are a couple ways we can do this:

1. Add a context parameter to (V)DAF sharding and preparation.
2. Allow the nonce to encode the context. At the moment, the nonce size is determined by the (V)DAF, and the client is supposed to choose it at random; we would need modify the syntax to support any nonce size and relax the requirements for choosing it.

Both of these options will require about the same amount of work for the DAP and VDAF specs. The downside of (1.) is that it requires API breaking changes. The downside of (2.) is that it makes the requirements more technical: the nonce now needs to have high min-entropy rather than being completely random. We could probably address this by providing some examples of secure ways to choose the nonce.

Do folks support this change?

Some reasons to do this:
1. Defense-in-depth is always good (and this change certainly can't hurt).
2. This aligns with a design philosophy of more conventional ZKP applications, which is to add as much context as possible to the transcript whenever you do Fiat-Shamir.

Some reasons not to do this:
1. It's a fairly big change, and there's a chance we get it wrong. Though we at least have a pretty good idea of what to do for Prio3.
3. It could create a false sense of security, leading to overly optimistic parameter choices getting deployed.

cjpatton:
Based on #421 (merge that first).
Partially addresses #418 (still need to propagate to draft).

Add an application context string, denoted `ctx`, to sharding and preparation APIs for DAFs and VDAFs. The intent is to use this for defense-in-depth: by binding execution of a VDAF to its application (concretely, in DAP we would set `ctx` to the task ID), we hope that an attack on that application does not translate directly to an attack on another application using the same parameters.

This change is motivated by the possibility of an offline attack against the robustness of some Prio3 variants. The attack is as follows: fix an invalid measurement, then search for measurement shares that derive joint randomness that causes the circuit to accept. Once these shares are found, they can be used to corrupt any batch that uses the same Prio3 parameters.

To reduce the blast radius of this attack, we can bind `ctx` to the joint randomness derivation, thereby ensuring the offline work done by the attacker is only useful to break a single application.

This change is slightly more general: wherever we compute a domain separation tag in either Prio3 or Poplar1, we append `ctx`. This is a more invasive than necessary, but it's also more conservative.

Finally, we modify XofTurboShake128 and XofFixedKeyAes128 by changing the length prefix for the domain separation tag from a single byte to two bytes, thereby accommodating larger application context strings.


cjpatton:
Scoped the PR to just changing the poc and converted to "ready for review". (We'll propagate the changes to the draft in the next PR.) Also rebased.

cjpatton:
Rebased.

divergentdave (approved):

PR #423, opened 2024-09-10
Add an RFC EDITOR note for TurboSHAKE

cjpatton:
We currently point to the draft; we'll want to replace this pointer with a reference to the RFC once it's published.


cjpatton:
CI should be fixed after merging #424.

divergentdave (approved):

PR #425, opened 2024-09-10
Strengthen recommendation about field size

cjpatton:
The security consideration section has the following recommendation for choosing parameters for circuits with joint randomness: "Field128 is RECOMMENDED, but Field64 MAY be used with at least three proofs".

Clarify where this requirement comes from (the base proof system needs around 128 bits of security) and strengthen this language somewhat: "Either Field128 or Field64 with three proofs MUST be used".

This change was suggested in an offline conversation with @henrycg.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -5521,7 +5521,7 @@ also stress that even if the IDPF is not extractable, Poplar1 guarantees that
 every client can contribute to at most one prefix among the ones being
 evaluated by the helpers.
 
-## Choosing the Field Size {#security-multiproof}
+## Choosing FLP Parameters {#security-multiproof}
 
 Prio3 and other systems built from FLPs ({{flp-bbcggi19}} in particular) may
 benefit from choosing a field size that is as small as possible. Generally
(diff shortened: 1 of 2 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -5526,11 +5526,15 @@ generate and verify multiple, independent proofs. (See {{multiproofs}}.) For
...
-broken for such a small field.
+For circuits involving joint randomness, we aim for a soundness error of the
+FLP close to `2^-128` in order to mitigate offline attacks. (The soundness
+error is the probability of a random joint randomness accepting an invalid
+measurement plus the base soundness error of the proof system, which is given
+by Theorem 4.3 of {{BBCGGI19}}.) Such circuits MUST use Field128 with at least
+one proof or Field64 with at least three proofs.
+
+In the case of Field64, breaking robustness for two proofs is feasible, if
+impractical; but one proof is completely broken for such a small field.
```
bifurcation, line 5537: This paragraph is just hand-waving as written.  If you want to make an argument for the MUST above, you should provide some explanation for why 3 is the right break point.  Presumably there's something that scales as `2^{k * PROOFS}`, so that `k * PROOFS` is big enough with `PROOFS == 3` but not below?

bifurcation (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: That's close to the truth. I've added the formula that is relevant here. What matters is that, depending on the circuit, Field64/2 might have significantly lower soundness than Field128/1.

PR #428, opened 2024-09-17
tls-presentation: Fix limits for length-prefixed values

cjpatton:
Assume the length of the encoded prep share is at least 1 byte for every VDAF; otherwise there would be no mechanism to convey validity. The prep message, however, may be empty, as it is for some Prio3 variants.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1582,17 +1582,18 @@ struct {
   MessageType type;
   select (Message.type) {
     case initialize:
-      opaque prep_share<0..2^32-1>;
+      opaque prep_share<1..2^32-1>;
     case continue:
       opaque prep_msg<0..2^32-1>;
-      opaque prep_share<0..2^32-1>;
+      opaque prep_share<1..2^32-1>;
     case finish:
       opaque prep_msg<0..2^32-1>;
   };
 } Message;
 ~~~
 
-These messages are used to transition between the states described in
+Note that the shortest encodable ping-pong message is `5` bytes long. These
+messages are used to transition between the states described in
 {{vdaf-prep-comm}}. The Leader's initial transition is computed with the
 following method, implemented on `Vdaf`:
 


draft-irtf-cfrg-vdaf.md
```diff
@@ -1582,10 +1582,10 @@ struct {
...
     case initialize:
-      opaque prep_share<0..2^32-1>;
+      opaque prep_share<1..2^32-1>;
     case continue:
       opaque prep_msg<0..2^32-1>;
-      opaque prep_share<0..2^32-1>;
+      opaque prep_share<1..2^32-1>;
     case finish:
       opaque prep_msg<0..2^32-1>;
   };
```
cjpatton, line 1591: The smallest message is 10 bytes:
* 1 byte for the type discriminator
* 9 bytes for the longest variant, which is continue:
  * 0 bytes for the prep message + 4 bytes for the length prefix
  * 1 byte for the prep share + 4 bytes for the length prefix

branlwyd: Following this logic, I think the smallest message is 5 bytes:

* 1 byte for the type discriminator
* 4 bytes for the shortest variant, which is finish:
  * 0 bytes for the prep message + 4 bytes for the length prefix

cjpatton: Oofta, you're right. Some times my brain just swaps min/max, $\leq$/$\greq$, $x$/$-x$, etc. Will fix here and in the DAP PR.

cjpatton: Noted here that the shortest pingpong message is 5 bytes.

branlwyd (approved):

cjpatton:
Let's hold off on merging this until we settle https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/110.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1582,17 +1582,18 @@ struct {
...
 } Message;
 ~~~
 
-These messages are used to transition between the states described in
+Note that the shortest encodable ping-pong message is `5` bytes long. These
```
bifurcation, line 1595:
```suggestion
Note that the shortest valid ping-pong message is `5` bytes long. These
```

Why does the reader need to know how long the shortest valid message is?  If it matters, you could get it down to 2 bytes using the [variable-length headers](https://www.rfc-editor.org/rfc/rfc9420.html#name-variable-size-vector-length) introduced in MLS.

bifurcation (commented):
I generally think that encoding these sorts of things in the TLS syntax is not that useful.  It gives you marginally faster failure, but makes it more complicated to set up the TLS parsing.

cjpatton:
It looks like the most likely outcome for DAP is that we'll just set all lower limits to 0: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/587. I'm therefore closing this.

PR #433, opened 2024-09-26
Change handling of IDPF indices

### LINKED ISSUE #416: IDPF: Change APIs to take alpha as a bit string, not an integer
As discussed in the 2024-09-06 call, we should change the IDPF gen() and eval() methods to take in bit strings (lists of booleans) instead of bigints. Iterating over the bits of the bit string would be clearer than shifting and masking an integer. We'll likewise need to change the recommendation for handling byte string, which currently suggests `from_be_bytes()`, and rewrite the prefix packing done in the aggregation parameter encoding/decoding.

divergentdave:
Closes #416. This changes the IDPF and Poplar1 APIs to explicitly take in a bit string, rather than an arbitrary precision integer. The encoding of Poplar1 aggregation parameters is changed so that the first prefix is encoded first, each prefix is padded out to the next byte boundary with zero bits, instead of packing them, and the last byte of each prefix may be partially used, instead of the first one.


draft-irtf-cfrg-vdaf.md
```diff
@@ -4967,31 +4967,39 @@ The aggregation parameter is encoded as follows:
...
-for (i, prefix) in enumerate(prefixes):
-    packed |= prefix << ((level + 1) * i)
-packed_len = ((level + 1) * len(prefixes) + 7) // 8
-packed_prefixes = to_be_bytes(packed, packed_len)
+prefixes_len = ((level + 1) + 7) // 8 * len(prefixes)
+encoded_prefixes = bytearray()
+for prefix in prefixes:
+    for chunk in itertools.batched(prefix, 8):
+        byte_out = 0
+        for bit_position, bit in enumerate(chunk):
```
cjpatton, line 4985: nit: When unpacking a tuple, elsewhere in the code we prefer to the keep the parentheses.
```suggestion
        for (bit_position, bit) in enumerate(chunk):
```

poc/tests/test_idpf_bbcggi21.py
```diff
@@ -84,46 +87,160 @@ def run_idpf_exhaustive_test(self, idpf: Idpf, alpha: int) -> None:
     def test(self) -> None:
         self.run_idpf_test(
             IdpfBBCGGI21(2, 16),
-            0b1111000011110000,
+            (
+                True, True, True, True, False, False, False, False,
+                True, True, True, True, False, False, False, False,
+            ),
```
cjpatton, line 93: Suggestion: `0b1111000011110000` is a nicer visual representation than the tuple of bools. For testing purposes, it might be nice to add a method for decoding a `tuple[bool, ...]` from an `int`

```suggestion
            prefix_from_int(0b1111000011110000),
```

poc/tests/test_idpf_bbcggi21.py
```diff
@@ -181,3 +331,18 @@ def test_public_share_roundtrip(self) -> None:
             idpf.decode_public_share(idpf.encode_public_share(public_share)),
             public_share,
         )
+
+
+def bytes_to_bit_string(s: bytes) -> tuple[bool, ...]:
```
cjpatton, line 336: Wouldn't this be the same procedure used by the agg param decoder?

poc/tests/test_idpf_bbcggi21.py
cjpatton: Bonus: it would make it easier to make sure these tests haven't changed.

poc/tests/test_idpf_bbcggi21.py
divergentdave: This is similar, but simpler since it only produces one bit string, and is restricted to only handling multiples of eight bits.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4249,10 +4241,10 @@ In the remainder we write `Output` as shorthand for the type
 denotes either a vector of inner node field elements or leaf node field
 elements.) The scheme is comprised of the following algorithms:
 
-* `idpf.gen(alpha: int, beta_inner: list[list[FieldInner]], beta_leaf:
+* `idpf.gen(alpha: tuple[bool, ...], beta_inner: list[list[FieldInner]], beta_leaf:
```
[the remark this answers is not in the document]
divergentdave: Sure, I'll add a comment in the Python code, as I don't think this is relevant at the spec level.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4326,19 +4319,27 @@ state across evaluations. See {{idpf-bbcggi21}} for details.
...
 
 How data are represented as IDPF indices is up to the application. When the
-inputs are fixed-length byte strings, the most natural choice of encoder is
-`from_be_bytes()`. This ensures that, when a string is a prefix of another, so
-too is its index. (Index prefixes are defined in {{idpf}}). For example,
+inputs are fixed-length byte strings, the most natural choice of representation
+is as a bit string formed from all the bits of the byte string, first ordered by
+byte position, then ordered from most significant bit to least significant bit
+within each byte. This ensures that, when a byte string is a prefix of another,
+so too is its corresponding index. (Index prefixes are defined in {{idpf}}). For
```
cjpatton, line 4326: This parenthesis is never closed
```suggestion
so too is its corresponding index. Index prefixes are defined in {{idpf}}). For
```

cjpatton (approved):
🦆

draft-irtf-cfrg-vdaf.md
divergentdave: It is closed, though it's clustered with curly braces here. This is carried over from existing text.

cjpatton: Oh, I think I missed it because the period comes after.

Normally if I have a paranthetical that is an entire sentence, I put the period with in the paranthetical.

cjpatton, line 4326:
```suggestion
so too is its corresponding index. (Index prefixes are defined in {{idpf}}.) For
```

PR #434, opened 2024-09-27
Add application context string

### LINKED ISSUE #418: Bind application context to VDAF computations

cjpatton:
Closes #418.

Propagate changes to the reference code from a recent change that add an application context string to sharding and preparation. Explain the intended role of context: disagreement should imply failure to aggregate a report. However, the affect is different for DAFs and VDAFs, since for DAFs there is no explicit signal of rejection; disagreement would manifest as a garbled aggregate result. Point this out.

Also, note in security considerations that this provides a cushion for weak parameters.

Finally, update the order of the context string in the `Daf` input parameters to align with the `Vdaf`. The context string comes ahead of every input except for the VDAF verification key.


draft-irtf-cfrg-vdaf.md
```diff
@@ -1168,18 +1177,19 @@ of the `Vdaf` interface specify a method of encoding the `PublicShare`,
 
 Each VDAF is identified by a unique, 32-bit integer `ID`. Identifiers for each
 (V)DAF specified in this document are defined in {{codepoints}}. The following
-method is defined for every VDAF:
+method is defined for each VDAF specified in this document:
```
cjpatton, line 1180: Forgot to note this in the commit: worth clarifying that this is for Prio3 and Poplar1. Other VDAFs don't need to version the same way, and perhaps they shouldn't. They'll need to version their own drafts anyway.

draft-irtf-cfrg-vdaf.md
```diff
@@ -902,6 +902,14 @@ this purpose.
 
   * The number of input shares MUST equal `SHARES`.
 
+Sharding is bound to a specific "application context". The application context
+is a string intended to uniquely identify the higher level protocol that uses
+the DAF. This is intended to ensure that aggregation succeeds only if the
+Clients and Aggregators agree on the application context. (Preparation binds
+the application context, too; see {{sec-daf-prepare}}.) Note that, unlike VDAFs
+({{vdaf}}), there is no explicit signal of disagreement; it may only manifest
+as a garbled aggregate result.
```
cjpatton, line 911: @divergentdave : identify an "instance" of the higher level protocl, not the higher level protocol itself

divergentdave, line 911: Here's a suggestion for the above comment.
```suggestion
Sharding is bound to a specific "application context". The application context
is a string intended to uniquely identify an instance of the higher level
protocol that uses the DAF. This is intended to ensure that aggregation succeeds
only if the Clients and Aggregators agree on the application context.
(Preparation binds the application context, too; see {{sec-daf-prepare}}.) Note
that, unlike VDAFs ({{vdaf}}), there is no explicit signal of disagreement; it
may only manifest as a garbled aggregate result.
```

divergentdave (approved):

PR #436, opened 2024-09-30
Prio3: Improve soundness of reductions (*)

### LINKED ISSUE #427: Prio3: Soundness improvement for reductions
Prio3Sum, Prio3SumVec, Prio3Histogram, and Prio3MultihotCountVec all involve checking that each of $n$ "sub-checks" outputs $0$. This works by interpreting the outputs as coefficients of a polynomial $p$, then evaluating $p$ at a random point $r\in F$. If $p(r)=0$, then we deem the input valid. Some of us have been calling this the circuit's "reduction".

Our current analysis of the circuit's soundness goes as follows. When the input is invalid, i.e., at least one sub-check output is non-zero, the probability that $p(r) = 0$ is equal to the probability that $r$ is a root of $p$, of which there are at most $\deg{p} / |F|$. The degree of $p$ is at most $n$, which gives us a soundness error of $n / |F|$.

This is also given by the [Schwartz-Zippel lemma](https://en.wikipedia.org/wiki/Schwartz%E2%80%93Zippel_lemma). More generally, let $p$ is a polynomial of $n$ variables with at least one non-zero coefficient. When we evaluate $p$ at a random point $(r_1, \ldots, r_n) \in F^n$, the probability that $p(r_1, \ldots, r_n) = 0$ is at most $\deg p / |F|$, where $\deg p$ denotes in particular the [total degree](https://en.wikipedia.org/wiki/Degree_of_a_polynomial#Extension_to_polynomials_with_two_or_more_variables) of $p$.

This lemma suggests an alternative reduction with better soundness. Let $c_1, \ldots, c_n$ denote the sub-check outputs. $p(x_1, \ldots, x_n) = c_1 \cdot x_1 + \cdots + c_n \cdot x_n$ be the polynomial with $n$ variables that has $c_1 \ldots, c_n$ as its coefficients. Choose a random vector $\vec{r} = (r_1, \ldots, r_N) \in F^n$ and evaluate $p$ at $\vec{r}$. If $p(\vec{r}) = 0$, then deem the input valid.

Soundness: since $\deg p = 1$, by the Schwartz-Zippel lemma it follows that the soundness error for this circuit is $1 / |F|$.

This is a pretty significant improvement, especially for larger inputs with lots of sub-checks. The base soundness of the current circuit is $\frac{n}{|F|} + \frac{2m}{|F|-m}$, where the second term is from {{BBCGGI19}}, Theorem 4.3; the soundness of the new circuit is $\frac{1}{|F|} + \frac{2m}{|F|-m}$. Concretely, for Prio3SumVec with `bits == 1`, `length = 100_000` and `chunk_length = 316` , this drops from $~ 2^{-111}$ to $2^{-118}$:

```sage
def sound(n, m, f): return -int((log(n/f + (2*m)/(f-m))/log(2)).n())
```

There is a catch: we end up with a much longer joint randomness, which means more times spent on XOF evaluation. On the other hand, we spend less time on finite field arithmetic, since we don't need to compute powers of $r$. Let's benchmark this, and if we don't see a performance degradation, then I think we should adopt this new reduction in each of the circuits, as well as for the [built-in reduction for circuits with multiple outputs](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-11#section-1.1-6.2.1).

Credit to @rozbb for pointing out this use of Shwartz-Zippel and thanks to @bwesterb and @bhalleycf for confirming the math.

cjpatton:
Closes #427.

In a validity circuit we may have several values we expect to be 0. To check that they're all zero, we interpret them as coefficients of a polynomial and evaluate the polynomial at random point. We do this in several places:

* Whenever `EVAL_OUTPUT_LEN > 0`
* In the `SumVec` circuit
* In the `MultihotCountVec` circuit
* In the `Histogram` circuit

This induces a soundness error of `n / |F|`, where `n` is the number of values and $F$ is the finite field. This is significant especially for larger inputs.

We can reduce this soundness error by reducing the degree of the polynomial. By the Schwartz-Zippel lemma, when we evaluate a non-zero, multivariate polynomial with degree `d` at a random point (each variable gets a random value), the probability that the output is zero is at most `d / |F|`. For instance, if each value is a different variable, then the maximum degree is `1`, resulting in a `1 / |F|` soundness error.

However, the more variables the polynomial, the more randomness we need to produce. This can become a CPU bottleneck for longer inputs.

This commit makes the following changes to Prio3 and its variants:

1. When `EVAL_OUTPUT_LEN > 0`, we include in the query randomness `EVAL_OUTPUT_LEN` field elements and compute the reduced output as the dot product of this random vector and the outputs. This improves the soundness error from `EVAL_OUTPUT_LEN / |F|` to `1 / |F|`.

2. We modify the `SumVec`, `MultihotCountVec` and `Histogram` circuits by including in the joint randomness a vector of length `GADGET_CALLS[0]`. For the `i`-th gadget call, we generate powers of `joint_rand[i]`. This results in a soundness error of `sqrt(length) / |F|` when the chunk length is chosen optimally.


cjpatton:
Rebased.

divergentdave (approved):
One other change: section 7.4.3 introduces the ParallelSum gadget and SumVec circuit, and there we should change "multiplied by a power of the joint randomness value" with "multiplied by a power of a joint randomness value". The same applies to the description of the Prio3Histogram circuit.

PR #437, opened 2024-09-30
Histogram, MultihotCountVec: Move final reduction (*)

cjpatton:
Based on #436 (merge that first)

Simplify the `Histogram` and `MultihotCountVec` circuits by moving the final reduction out of the circuit.


cjpatton:
Rebased.

divergentdave (approved):

divergentdave:
The sentence "All the individual checks are combined together in a random linear combination" in the Prio3Histogram section is technically still correct, but the random linear combination now happens both inside and outside the circuit, which could be a bit confusing.

cjpatton:
> The sentence "All the individual checks are combined together in a random linear combination" in the Prio3Histogram section is technically still correct, but the random linear combination now happens both inside and outside the circuit, which could be a bit confusing.

Changed to "The individual checks constitute the output of the circuit."

Issue #431, opened 2024-09-23
Consider removing sage as a dependency


cjpatton:
[It was suggested on the CFRG mailing list](https://mailarchive.ietf.org/arch/msg/cfrg/ht4YzBMYWoQvKxlCxx-CAPJFtC0/) that VDAF implementers could integrate the reference code directly into their tests, similar to the [SUPERCOP](https://bench.cr.yp.to/supercop.html) project. However such tests would need sage as a dependency, which is not ideal. If we removed the sage dependency, then the only dependency would by python3, which is just about everywhere these days. Of course, they would also need to install pycryptodomex.

I believe we only use sage for the finite field arithmetic and polynomial interpolation, neither of which would be terribly complicated to implement ourselves.

Note that this conflicts with #282.

cjpatton:
Closed by #445.

PR #445, opened 2024-10-04
Remove use of Sage, reimplement field and polynomial operations

divergentdave:
This is a first attempt at #431. It works, and it's slower, which is to be expected, but currently it's a lot slower. Unit tests take 22 seconds instead of 2 seconds. I think I may take a second pass at polynomial interpolation, using the method of divided differences, and see if that improves the speed somewhat.


cjpatton:
> This is a first attempt at #431. It works, and it's slower, which is to be expected, but currently it's a lot slower. Unit tests take 22 seconds instead of 2 seconds. I think I may take a second pass at polynomial interpolation, using the method of divided differences, and see if that improves the speed somewhat.

22 is definitely too slow :) Another option is to just implement NTT? We only use it in FLP anyway.

divergentdave:
Divided differences brings the time down to 10s. These are the remaining ideas I have for algorithmic speedups:

- Conditionally use NTT if the interpolation points match
- Use the extended Euclidean algorithm for multiplicative inverses, instead of Euler's theorem. Sage currently uses the extended Euclidean algorithm via GMP. (addition-chain exponentiation could speed up using Euler's theorem, but we don't care about constant timeness, and extended GCD is likely faster and simpler)
- I recall discovering that Python's modular exponentiation algorithm for 3-argument `pow()` is slower than what Sage has, but that might have been only relevant for much larger inputs.

I'll do the extended Euclidean algorithm next, as that is easy to do, and we'll see how far that gets us.

> By the time you're done, you will have written an ad hoc, informally-specified, bug-ridden, slow implementation of one percent of SageMath.

-- https://cryptopals.com/sets/8

divergentdave:
Test suite runtime is now 6.7s in CI and 4.1s locally for me. I think this is about as good as we can get it without a lot more complexity.

cjpatton:
> Test suite runtime is now 6.7s in CI and 4.1s locally for me. I think this is about as good as we can get it without a lot more complexity.

4 seconds is doable. Have you tried generating the test vectors? Hopefully that's quick, too.

My main concern about landing this is the runtime of PINE's (https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/) unit tests, which at the moment take about 10 seconds. Let's prototype this in that repo before land this.

divergentdave:
Test vector generation just takes 0.6s for me.

FWIW, here are the slowest tests. We could probably also speed things up by tweaking the dimensions they use, but I think that would be best handled separately.

```
Slowest test durations
----------------------------------------------------------------------
2.053s     test (tests.test_vdaf_prio3.TestPrio3SumVecWithMultiproof.test)
0.905s     test (tests.test_idpf_bbcggi21.TestIdpfBBCGGI21.test)
0.478s     test (tests.test_vdaf_prio3.TestPrio3SumVec.test)
0.219s     test_poplar1 (tests.test_vdaf_poplar1.TestPoplar1.test_poplar1)
```

cjpatton:
> My main concern about landing this is the runtime of PINE's (https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/) unit tests, which at the moment take about 10 seconds. Let's prototype this in that repo before land this.

I hacked this together here: https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/commit/bfe25cd9bb08008c608f6455bf71193990337580

As expected, the unit tests take longer, about twice as long (now 12 seconds). Test vector generation also takes about twice as long, now 120 seconds. Note that we generate test vectors for higher dimensional data for PINE than we do for Prio3.

I still think we should take this change, given how much more portable it makes the reference code, but let's check with @junyechen1996 to make sure this isn't a deal breaker.

divergentdave:
For PINE's sake, there's one other algorithmic improvement we may want to consider other than those listed above: multiplying polynomials using NTT instead of the straightforward quadratic algorithm. This only matters when multiplying large polynomials.

cjpatton:
> For PINE's sake, there's one other algorithmic improvement we may want to consider other than those listed above: multiplying polynomials using NTT instead of the straightforward quadratic algorithm. This only matters when multiplying large polynomials.

Right, as we do in libprio already. How much would we have to change the spec? I wouldn't want to have to change too much to accommodate this. In particular, I wouldn't want to make the spec less clear.

cjpatton:
@junyechen1996 is cool with this change. I'll review tomorrow.

divergentdave:
> Right, as we do in libprio already. How much would we have to change the spec? I wouldn't want to have to change too much to accommodate this. In particular, I wouldn't want to make the spec less clear.

Neither `poly_mul()` nor `poly_interp()` are excerpted in the spec, so this is all implementation details of the proof of concept.

cjpatton:
> > Right, as we do in libprio already. How much would we have to change the spec? I wouldn't want to have to change too much to accommodate this. In particular, I wouldn't want to make the spec less clear.
> 
> Neither `poly_mul()` nor `poly_interp()` are excerpted in the spec, so this is all implementation details of the proof of concept.

Got it. Let's keep it that way: if anyone tells us we need to list this, then we can do so.

As far as optimizing `poly_mul()`, I'm fine with holding off for now.

.github/workflows/test.yml
```diff
@@ -16,20 +16,23 @@ jobs:
...
 
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+        cache: pip
+
     - name: Install dependencies
-      run: sage --pip install pycryptodomex mypy
+      run: python -m pip install pycryptodomex mypy matplotlib
```
cjpatton, line 30: Why is matplotlib needed?
If this is strictly for `plot_prio3_multiproof_robustness.py`, then IMO we're better reomving this script rather than adding the dependency.

.gitignore
```diff
@@ -14,8 +14,6 @@ draft-patton-cfrg-vdaf.xml
 lib
 report.xml
 venv/
-sagelib
-*.sage.py
```
cjpatton, line 18: Man, remember when we we're just had a bunch of `.sage` files? That was truly terrible. Thank you so much for your hard work on the software engineering for this spec.

poc/README.md
```diff
@@ -8,16 +8,16 @@ generate test vectors.
 
 ## Installation
 
-This code requires [SageMath](https://www.sagemath.org/) version >= 9.6 to run. To install:
+This code requires Python 3.12 or later to run. To install:
 
 ```
-sage -pip install git+https://github.com/cfrg/draft-irtf-cfrg-vdaf@draft-irtf-cfrg-vdaf-11#subdirectory=poc
+pip install git+https://github.com/cfrg/draft-irtf-cfrg-vdaf@draft-irtf-cfrg-vdaf-11#subdirectory=poc
```
cjpatton, line 14: We suggest elsewhere to invoke pip this way. Unless, is there a different affect you intend here?
```suggestion
python -m pip install git+https://github.com/cfrg/draft-irtf-cfrg-vdaf@draft-irtf-cfrg-vdaf-11#subdirectory=poc
```

poc/gen_test_vec.py
cjpatton, line 1: Suggestion: run this from CI so that we don't accidentally merge a broken script.

poc/gen_test_vec.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
```
cjpatton, line 1: Don't we also need to change the file permissions to make this executable?

poc/vdaf_poc/field.py
```diff
@@ -249,13 +231,23 @@ def poly_strip(field: type[F], p: list[F]) -> list[F]:
...
     return poly_strip(field, r)
 
 
+def poly_add(field: type[F], p: list[F], q: list[F]) -> list[F]:
+    """Add two polynomials."""
+    r = field.zeros(max(len(p), len(q)))
+    for i, pi in enumerate(p):
+        r[i] = pi
+    for i, qi in enumerate(q):
+        r[i] += qi
```
cjpatton, line 247:
```suggestion
    for (i, p_i) in enumerate(p):
        r[i] = pi
    for (i, q_i) in enumerate(q):
        r[i] += qi
```

poc/vdaf_poc/field.py
```diff
@@ -272,6 +264,103 @@ def poly_eval(field: type[F], p: list[F], eval_at: F) -> F:
 
 def poly_interp(field: type[F], xs: list[F], ys: list[F]) -> list[F]:
     """Compute the Lagrange interpolation polynomial for the given points."""
-    R = PolynomialRing(field.gf, 'x')
-    p = R.lagrange_polynomial([(x.val, y.val) for (x, y) in zip(xs, ys)])
-    return poly_strip(field, list(map(field, p.coefficients())))
+
+    # This uses Newton's divided difference interpolation formula.
```
cjpatton, line 268: nit: This would look a little neater if included in the `""" ... """` comment block above.
Also, consider adding an appropriate reference. Wikipedia is perfectly fine for our purposes here.

Here's perhaps a nice reference: https://pythonnumericalmethods.studentorg.berkeley.edu/notebooks/chapter17.05-Newtons-Polynomial-Interpolation.html

In fact, you might mention here that this is a re-implementation of what numpy does, and that we don't want to add numpy as a dependency. In fact, it may not work for us here directly.

poc/vdaf_poc/field.py
```diff
@@ -272,6 +264,103 @@ def poly_eval(field: type[F], p: list[F], eval_at: F) -> F:
...
+
+    assert len(xs) == len(ys)
+    n = len(xs)
+    one = field(1)
+
+    # We interleave three computations in each iteration of the outermost loop.
+    # First, we compute the ith Newton basis polynomial. Second, we compute the
```
cjpatton, line 275: nit: I find adding backticks around math improves readability. Just a suggestion, feel free to ignore. (Here and below.)
```suggestion
    # First, we compute the `i`th Newton basis polynomial. Second, we compute the
```

poc/vdaf_poc/field.py
```diff
@@ -272,6 +264,103 @@ def poly_eval(field: type[F], p: list[F], eval_at: F) -> F:
...
+    last_remainder, remainder = a, b
+    a, last_a, b, last_b = 0, 1, 1, 0
+    while remainder:
+        last_remainder, (quotient, remainder) = (
+            remainder,
+            divmod(last_remainder, remainder),
+        )
+        a, last_a = last_a - quotient * a, a
+        b, last_b = last_b - quotient * b, b
+    return last_remainder, last_a, last_b
```
cjpatton, line 353: Suggestion: I generally prefer adding parentheses when constructing/destructing tuples. In any case I think we should try to be consistent. At the moment there's a mixture of both.
```suggestion
    return (last_remainder, last_a, last_b)
```

poc/vdaf_poc/field.py
```diff
@@ -272,6 +264,103 @@ def poly_eval(field: type[F], p: list[F], eval_at: F) -> F:
...
+
+
+def invmod(x: int, p: int) -> int:
+    """
+    Modular multiplicative inverse.
```
cjpatton, line 358:
```suggestion
    Return x_inv such that (x_inv * x % p) == 0.
```

cjpatton (approved):

poc/README.md
divergentdave: Good point, `python -m pip` just tends to be a bit more foolproof on systems with multiple Python installations, and inconsistent scripts in each.

poc/vdaf_poc/field.py
```diff
@@ -272,6 +264,103 @@ def poly_eval(field: type[F], p: list[F], eval_at: F) -> F:
 
 def poly_interp(field: type[F], xs: list[F], ys: list[F]) -> list[F]:
     """Compute the Lagrange interpolation polynomial for the given points."""
-    R = PolynomialRing(field.gf, 'x')
-    p = R.lagrange_polynomial([(x.val, y.val) for (x, y) in zip(xs, ys)])
-    return poly_strip(field, list(map(field, p.coefficients())))
```
cjpatton, line 277: I would add a historical note here that we previously invoked the reference code with sage, which gave us a faster implementation of this algorithm. The note should give justification for this (dropping sage makes the code more portable).

poc/gen_test_vec.py
divergentdave: I did so, this is indicated as "100644 → 100755" next to the filename in GitHub's UI.

PR #447, opened 2024-10-07
Remove RFC EDITOR note about DAP draft

### LINKED ISSUE #439: Make sure all "RFC EDITOR" notes are actually tasks for the RFC editor
@bifurcation points out here that we've been a bit fast and lose with this: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/430#discussion_r1775648553

cjpatton:
Closes #439.

RFC 9180 (HPKE) and has an informative reference to draft-ietf-tls-esni (a protocol that uses HPKE). It's therefore reasonable to believe that VDAF may have an informative reference to draft-ietf-ppm-dap.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -373,11 +373,10 @@ literature. These protocols vary in their operational and security
 requirements, sometimes in subtle but consequential ways. This document
 therefore has two important goals:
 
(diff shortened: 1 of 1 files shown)


divergentdave (approved):

cjpatton:
Done: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/454

PR #450, opened 2024-10-07
Fix long line

divergentdave:
This fixes another long line, introduced in a recent PR. It also adds a check to CI to catch such warnings going forward.

I tested that this works in CI, see https://github.com/cfrg/draft-irtf-cfrg-vdaf/actions/runs/11220047961/job/31187354251.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,9 +8,14 @@ jobs:
   build:
     name: "Lint document"
     runs-on: ubuntu-latest
+    container:
+      image: docker://ghcr.io/martinthomson/i-d-template-action:latest
     steps:
     - name: "Checkout"
       uses: actions/checkout@v2
 
     - name: "Check for typos"
       run: "bash -c '! grep -r --file=rejected_dictionary draft-irtf-cfrg-vdaf.md poc'"
+
+    - name: "Check for warnings emitted by xml2rfc"
+      run: "bash -c 'make |& (! grep Warning)'"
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -8,9 +8,14 @@ jobs:
...
+    container:
+      image: docker://ghcr.io/martinthomson/i-d-template-action:latest
     steps:
     - name: "Checkout"
       uses: actions/checkout@v2
 
     - name: "Check for typos"
       run: "bash -c '! grep -r --file=rejected_dictionary draft-irtf-cfrg-vdaf.md poc'"
+
+    - name: "Check for warnings"
```
cjpatton, line 20:
```suggestion
    - name: "Check for warnings emitted by the I-D tool"
```

.github/workflows/lint.yml
```diff
@@ -8,9 +8,14 @@ jobs:
...
+      run: "bash -c 'make |& (! grep Warning)'"
```
cjpatton, line 21: nit: run `make clean` first just in case we already have run `make` in this container. 

(Or tell me that this can't happen because of how containers work!)

cjpatton (approved):

.github/workflows/lint.yml
divergentdave: This is just reusing the same container as the reusable GitHub action we use elsewhere, so that various utilities are already present. The Markdown file is created by the checkout above, and this is the only `make` invocation in this job, so the .txt and .html output files won't exist yet.

PR #451, opened 2024-10-07
Prio3: Improve circuit specifications

### LINKED ISSUE #411: @armfazh: draft 11: Replace circuit and gadget descriptions with their classes
We currently [describe a circuit](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-11#section-7.4.1) this way:

```
The Count validity circuit is defined as

def eval(
        self,
        meas: list[F],
        joint_rand: list[F],
        _num_shares: int) -> list[F]:
    self.check_valid_eval(meas, joint_rand)
    squared = self.GADGETS[0].eval(self.field, [meas[0], meas[0]])
    return [squared - meas[0]]
```

Here "Count" refers to a sub-class of "Valid" and "eval()" refers to an abstract method on Valid. This is not at all clear from the text. In particular, "eval" here doesn't appear to have anything to do with "Count".

@armfazh suggests we fix this by listing the "Count" class here instead. This also has the benefit of making this section of the draft more copy-pastable. 

We should do the same for gadgets as well. IN fact, rather than define them "on first use" as we're doing now, we should pull them into their own section. I believe we just have three: `Mul`, `PolyEval`, and `ParallelSum`.

cjpatton:
Stacked on #449 (to reduce rebase conflicts later on).
Closes #411.

Rework the spec of each Prio3 variant by listing the entire class of each circuit. Also, list the specification of each gadget in 
the appendix.

* Prio3Sum: Use `PolyEval` instead of `Range2`, as this is a much more useful gadget to list in the doc.
* Say "variant" instead of "instance".
* Document missing operations on polynomials
* Modify the constructor for `Field` elements by interpreting negative integers as additive inverses. This is necessary in  order to allow expressing polynomial coefficients in the `PolyEval` circuit independent of the field modulus.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -830,10 +830,21 @@ Some common functionalities:
   overriding the start of the range and overriding the step between successive
   output values.
 
+* `poly_eval(field: type[F], p: list[F], x: F) -> F` returns the result of
+  evaluating the polynomial, `p(x)`. The coefficients of polynomials are stored
+  in lists in ascending order of degree, starting with the constant coefficient.
+
 * `poly_interp(field: type[F], inputs: list[F], outputs: list[F]) -> list[F]`
   returns the coefficients of the lowest degree polynomial `p` for which
   `p(input[k]) == output[k]` for all `k`. Normally this will be computed using
   the Number Theoretic Transform (NTT) {{SML24}}.
+
+* `poly_mul(field: type[F], p: list[F], q: list[F]) -> list[F]` returns
+  the product of two polynomials.
+
+* `poly_strip(field: type[F], p: list[F]) -> list[F]` strips the zeros from the
+  end of the input polynomial. That is, it returns `p[:i]` where `i` is the
+  index of the last non-zero coefficient of `p`.
 
 # Overview
 
@@ -3799,12 +3810,13 @@ def decide(self, verifier: list[F]) -> bool:
     return True
 ~~~
 
-## Instantiations {#prio3-instantiations}
+## Variants {#prio3-instantiations}
 
-This section specifies instantiations of Prio3 for various measurement types.
-Each is determined by a field ({{field}}), a validity circuit
-({{flp-bbcggi19-valid}}), an XOF ({{xof}}). and the number of proofs to
+This section specifies instantiations of Prio3 for various aggregation tasks.
+Each variant is determined by a field ({{field}}), a validity circuit
+({{flp-bbcggi19-valid}}), an XOF ({{xof}}), and the number of proofs to
 generate and verify. Test vectors for each can be found in {{test-vectors}}.
+All gadgets are listed in {{gadgets}}.
 
 ### Prio3Count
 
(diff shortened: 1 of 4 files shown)


cjpatton:
Reviewer note: the diff is quite big because we are coalescing the circuit specification, encoding scheme, and parameter table into a single listing. My recommendation would be not look at the diff, but instead pull the branch and read through {{prio3-instantiations}}.

draft-irtf-cfrg-vdaf.md
```diff
@@ -5906,7 +5925,126 @@ collisions.  Applications SHOULD prefer the use of registered code points.
...
+                  inp_poly: list[list[F]]) -> list[F]:
+        p = [field(coeff) for coeff in self.p]
+        out = [field(0)] * (self.DEGREE * len(inp_poly[0]))
+        out[0] = p[0]
+        x = inp_poly[0]
+        for i in range(1, len(p)):
+            for j in range(len(x)):
+                out[j] += p[i] * x[j]
+            x = poly_mul(field, x, inp_poly[0])
+        return poly_strip(field, out)
```
cjpatton, line 5993: This needs to be defined.

cjpatton:
Rebased.

cjpatton:
Rebased. Initially the unit tests failed because `Field` expects a non-negative integer and we were expressing the `Range2()` as `PolyEval([0, -1, 1])`. We need to express this gadget this way because we don't know the field modulus at the time the gadget is constructed.

cjpatton:
Rebased.

draft-irtf-cfrg-vdaf.md
```diff
@@ -4123,105 +4114,113 @@ proofs up to 50% larger than the optimal proof size.
...
+
+    def truncate(self, meas: list[F]) -> list[F]:
+        return meas
+
+    def decode(
+            self,
+            output: list[F],
+            _num_measurements: int) -> list[int]:
+        return [bucket_count.as_unsigned() \
+            for bucket_count in output]
```
divergentdave, line 4222: nit: style (and this matches the Python file too)
```suggestion
        return [bucket_count.as_unsigned()
                for bucket_count in output]
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4259,94 +4257,125 @@ be encoded with `bits_for_weight` bits.
...
+
+        self.length = length
+        self.max_weight = max_weight
+        self.chunk_length = chunk_length
+        self.GADGETS: list[Gadget[F]] = [
+            ParallelSum(Mul(), chunk_length),
+        ]
+        self.GADGET_CALLS = [
+            (length + self.bits_for_weight + chunk_length - 1) \
+                // chunk_length
```
divergentdave, line 4305: nit: style and matching the Python file
```suggestion
            (length + self.bits_for_weight + chunk_length - 1)
            // chunk_length
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4259,94 +4257,125 @@ be encoded with `bits_for_weight` bits.
...
-| `AggResult`       | `list[int]`                                     |
-{: title="Parameters of validity circuit MultihotCountVec."}
+    def truncate(self, meas: list[F]) -> list[F]:
+        return meas[:self.length]
+
+    def decode(
+            self,
+            output: list[F],
+            _num_measurements: int) -> list[int]:
+        return [bucket_count.as_unsigned() for \
```
divergentdave, line 4376: nit: style and matching the Python file
```suggestion
        return [bucket_count.as_unsigned() for
```

divergentdave (approved):

PR #452, opened 2024-10-07
Replace `aggregate()` with streaming API

### LINKED ISSUE #432: Replace `aggregate()` with a streaming API
The current aggregation API takes a sequence of output shares and returns an aggregate share. This envisions VDAFs for which the aggregate result depends on the order in which output shares are aggregated, but so far this isn't true of anything we've come up with. In fact, DAP (implicitly?) requires a VDAF for which output shares can be aggregated in any order. (The Leader and Helper may not aggregate output shares in the same order.)

One problem with the current API is that it doesn't align well with practice. Normally what we expect Aggregators to do is aggregate in a "streaming" fashion, meaning output shares are not stored individually, but accumulated into an existing aggregate share as soon as they're ready to be aggregated.

In fact, we'd like to make this "streaming aggregation" normative in DAP: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/581.

However in order to do this, we'll need to modify the VDAF syntax. There's a few things we could do, but probably the simplest is to split `vdaf.aggregate()` into the following functions:

* `vdaf.agg_init() -> AggShare`: returns an empty aggregate share.
* `vdaf.agg_update(agg_share: AggShare, out_share) -> AggShare`: updates an aggregate share with the incoming output share.

cjpatton:
Closes #432.

Modify both `Daf` and `Vdaf` by replacing `aggregate()` with two
methods:

* `agg_init()` returns an empty aggregate share
* `agg_update()` updates an aggregate share with an output share

Also, add a method `merge()` for merging two aggregate shares, as will be required for specifying aggregation in DAP.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -875,9 +875,9 @@ measurement process is as follows:
       this process succeeds, then each aggregator holds one output share.
     * In VDAFs, Aggregators will need to exchange information among themselves
       as part of the validation process.
-* Each Aggregator combines the output shares in the batch to compute the
-  "aggregate share" for that batch, i.e., its share of the desired aggregate
-  result.
+* As each Aggregator recovers output shares, it accumulates them into an
+  "aggregate share" for the batch, i.e., its
+  share of the desired aggregate result.
 * The Aggregators submit their aggregate shares to the Collector, who combines
   them to obtain the aggregate result over the batch.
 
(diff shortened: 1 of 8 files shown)


cjpatton:
Reviewer note: So far I've just modified the reference code. If folks are happy with this approach, then I'll propagate to the doc.

draft-irtf-cfrg-vdaf.md
```diff
@@ -858,9 +858,9 @@ measurement process is as follows:
       this process succeeds, then each aggregator holds one output share.
     * In VDAFs, Aggregators will need to exchange information among themselves
       as part of the validation process.
-* Each Aggregator combines the output shares in the batch to compute the
-  "aggregate share" for that batch, i.e., its share of the desired aggregate
-  result.
+* As each Aggregator recovers output shares, it accumulates them into an
+  "aggregate share" for the batch, "aggregate share" for that batch, i.e., its
```
branlwyd, line 862:
```suggestion
  "aggregate share" for the batch, i.e., its
```

(duplicated text)

draft-irtf-cfrg-vdaf.md
```diff
@@ -1040,46 +1040,62 @@ DAFs MUST implement the following function:
...
+~~~~
+{: #aggregate-flow title="Local aggregation of output shares by Aggregator j.
+The number of measurements in the batch is denoted by MEAS_COUNT."}
+
+In many situations it is desirable to split an aggregate share across multiple
+storage elements, then merge the aggregate shares together just before
+releasing the completed aggregate share to the Collector. DAFs facilitate this
+with the following method:
+
+* `daf.merge(agg_param: AggParam, agg_shares: list[AggShare]) -> AggShare`
```
branlwyd, line 1085: nitpicky Q. `agg_update` is specified as a binary operation on an `AggShare` and an `OutShare`; `merge` is specified as acting on a list of `AggShare`s (both operations parameterized on an `AggParam`).

Is there a reason for this difference, i.e. do we think we might be able to implement more efficient merging than if we specified `merge(agg_param, agg_share_left, agg_share_right)`?

draft-irtf-cfrg-vdaf.md
```diff
@@ -1040,46 +1040,62 @@ DAFs MUST implement the following function:
...
+### Aggregation Order {#agg-order}
 
-Implementation note: for most natural DAFs (and VDAFs) it is not necessary for
-an Aggregator to store all output shares individually before aggregating.
-Typically it is possible to merge output shares into aggregate shares as they
-arrive, merge these into other aggregate shares, and so on. In particular, this
-is the case when the output shares are vectors over some finite field and
-aggregating them involves merely adding up the vectors element-wise. Such is the
-case for Prio3 {{prio3}} and Poplar1 {{poplar1}}.
+For most DAFs and VDAFs, the outcome of aggregation is not sensitive to the
```
branlwyd, line 1090: I forgot about this wrinkle; I don't think we mention in DAP that we can only work with VDAFs that allow aggregating output shares in an arbitrary order.

Do you think it is important to maintain that VDAFs might require a specific aggregation ordering? If not, let's drop this section (and maybe replace it with a note that aggregation can occur in any order). If so, I'll update DAP's VDAF requirements to note the arbitrary-ordering requirement.

draft-irtf-cfrg-vdaf.md
cjpatton: Motivation: I think this will be helpful for DAP, where we will have a sequence of aggregate shares, one for each "bucket", where a collection request specifies a sequence of buckets (e.g., a a time interval specifies a sequence of non-overlapping time windows).

Another benefit is that we can use `merge()` in the implementation of `unshard()` for both Prio3 and Poplar1.

Is there any benefit to making`merge()` a binary operator besides symmetry?

draft-irtf-cfrg-vdaf.md
cjpatton: Yup, this is indeed one of our requirements for DAP.

It seems unlikely, but the possibility has been brought up by enough people that I don't think we should rule it out completely. Better I think to document the requirement as part off https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/581.

draft-irtf-cfrg-vdaf.md
branlwyd: Nope, symmetry was my only concern, and making sure I wasn't missing anything. I don't think any of those uses would be hindered by making this a binary operator (I suppose it's a matter of where the `for` loop goes), but I also don't think this is important enough to hold this PR up over.

branlwyd (approved):

draft-irtf-cfrg-vdaf.md
branlwyd: Closing the loop: I sent https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/603 for this.

cjpatton:
Rebased and squashed @branlwyd's feedback.

divergentdave (approved):

cjpatton:
Rebased and squashed.

PR #478, opened 2024-10-10
Pass of intro, conventions, and overview

cjpatton:


draft-irtf-cfrg-vdaf.md
```diff
@@ -354,16 +355,16 @@ In fact, our framework admits DAFs with slightly more
 functionality, computing aggregation functions of the form
 
 ~~~
-f(agg_param, meas_1, ..., meas_N) =
-    g(agg_param, meas_1) + ... + g(agg_param, meas_N)
+F(agg_param, meas_1, ..., meas_BATCH_SIZE) =
+    G(agg_param, meas_1) + ... + G(agg_param, meas_BATCH_SIZE)
```
cjpatton, line 359: * capital letters for functions looks a little nicer in the *.txt file, which is what I'm reviewing.
* `N` is used in the overview section to denote the number of aggregators.

draft-irtf-cfrg-vdaf.md
```diff
@@ -397,13 +398,14 @@ therefore has two important goals:
     1. General patterns of communications among the various actors involved in
        the system (clients, aggregation servers, and the collector of the
        aggregate result);
-    1. Capabilities of a malicious coalition of servers attempting to divulge
+    1. Capabilities of a malicious coalition of parties attempting to divulge
```
cjpatton, line 401: collector may be under the attacker's control and is not necessarily a server

draft-irtf-cfrg-vdaf.md
```diff
@@ -781,44 +785,43 @@ a particular encoding of that quantity as a byte string.
 
 Some common functionalities:
```
cjpatton, line 786: In general I've tried to harmonize variable names across all functions:
* `x` denotes the input that's "acted upon"
* `len` denotes the length when required

draft-irtf-cfrg-vdaf.md
```diff
@@ -781,44 +785,43 @@ a particular encoding of that quantity as a byte string.
...
 
-* `zeros(len: int) -> bytes` returns an array of zero bytes. The length of
-  `output` MUST be `len`.
```
cjpatton, line 785: Don't use MUST to express bounds on outputs

draft-irtf-cfrg-vdaf.md
```diff
@@ -781,44 +785,43 @@ a particular encoding of that quantity as a byte string.
...
+  greater than or equal to `x` that is also a power of two.
 
-* `additive_secret_share(vec: list[F], num_shares: int, field: type[F])
-  -> list[list[F]]` takes a vector of field elements and returns multiple
-  vectors of the same length, such that they all add up to the input vector,
-  and each proper subset of the vectors are indistinguishable from random.
+* `additive_secret_share(x: list[F], num_shares: int, field: type[F]) ->
+  list[list[F]]` takes a vector `x` of field elements and returns multiple
+  vectors of the same length, such that they all add up to the input vector.
+  Note that this function is not used normatively in this document.
```
cjpatton, line 822: this method is only used for `run_flp()` and we don't care if it's an actually secure secret sharing.

draft-irtf-cfrg-vdaf.md
```diff
@@ -781,44 +785,43 @@ a particular encoding of that quantity as a byte string.
...
-  cryptographically secure pseudorandom number generator (CSPRNG). The length
-  of `output` MUST be `len`.
+* `gen_rand(len: int) -> bytes` returns a byte array of the requested length
+  (`len`) generated by a cryptographically secure pseudorandom number generator
+  (CSPRNG).
 
-* `byte(int: int) -> bytes` returns the representation of `int` as a byte
-  string. The value of `int` MUST be in `[0,256)`.
+* `byte(x: int) -> bytes` returns the representation of `x` in range
+  `[0, 256)`.
```
divergentdave, line 796: I think we should still mention that this is mapping to single-byte byte strings.
```suggestion
* `byte(x: int) -> bytes` returns the representation of the integer `x` in range
  `[0, 256)` as a single-byte byte string.
```

divergentdave (approved):

PR #479, opened 2024-10-11
Pass of DAP/VDAF up to communication patterns

cjpatton:
Stacked on #478.

The gist is that we lean more heavily on the DAF section for the normative VDAF definition. We replace the sharding, aggregation, and unsharding sections with references to the same topics for DAFs, plus pointers to subsections or bounds on behavior wherever relevant.

* Harmonize language around report rejection. That is, invalid measurements manifest as failures to complete preparation rather than "errors computed by each Aggregator".

* Harmonize "BATCH_SIZE" in intro with "MEAS_COUNT" in DAF aggregation section. What we want is actually `num_measurements`, as this is the parameter name for unsharding, but this is a little too long to put in figures. Go with `M`, as this is short and seems the least fussy.

* Move illustrations to the top of sections.

* Define "public share" before its first use.

* Remove the paragraph in the DAF overview about the ID. It doesn't really help clarify anything, and we don't use it at all in this section.

* Expand the CSPRNG acronym where it's first used.

* Add a paragraph to DAF sharding that explains the role of the nonce.

* Add a paragraph to DAF preparation saying the Aggregators MUST agree on the agg param.

* Add a forward reference to `format_dst()`.

* In VDAF preparation, define `prep_shares_to_prep()` prior to `prep_next()` since this is the order in which they're executed.

* RECOMMEND specifying an encoding for `AggParam`.

* Use consistent language about the nonce.


draft-irtf-cfrg-vdaf.md
```diff
@@ -1390,28 +1355,44 @@ before they can begin preparing the measurement shares for aggregation.
...
+At the start of each round, each Aggregator broadcasts a message called a
+"preparation share", or "prep share" for short. The prep shares are then
+combined into a single message called the "preparation message", or "prep
+message". The prep message may be computed by any one of the Aggregators.
+
+The prep message is disseminated to each of the Aggregators to begin the next
+round, or compute the output shares in case of the last round. An Aggregator
+begins the first round with its input share and it begins each subsequent round
+with the previous prep message. Its output in the last round is its output
+share and its output in each of the preceding rounds is a prep share.
```
divergentdave, line 1372: I think these paragraphs would be a good place to introduce "prep state". Its first use is currently in the definition of `vdaf.prep_init()` below. How about this?

> Aggregators retain some local state between successive rounds of preparation. This is referred to as `preparation state` or `prep state` for short.

draft-irtf-cfrg-vdaf.md
```diff
@@ -1424,89 +1405,55 @@ methods:
...
-protocol). Another way to draw this boundary would be to have the Aggregators
-recover output shares first, then verify that they are valid. However, this
-would allow the possibility of misusing the API by, say, aggregating an invalid
-output share. Moreover, in protocols like Prio+ {{AGJOP21}} based on oblivious
-transfer, it is necessary for the Aggregators to interact in order to recover
-aggregatable output shares at all.
+An exception may be raised by one of these algorithms, in which case the report
+MUST be deemed invalid and not processed any further.
+
+Implementation note: The preparation-state update accomplishes two tasks:
```
divergentdave, line 1424: This phrasing doesn't make sense any more without the context of the preceding "linear state machine" paragraph
```suggestion
Implementation note: The preparation algorithm accomplishes two tasks:
```

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: Changed to "process" since it's actually a collection of agorithms.

draft-irtf-cfrg-vdaf.md
cjpatton: Done.

PR #501, opened 2024-10-15
Take some of @armfazh's suggestions for common conventions

### LINKED ISSUE #488: comments on Section 2
These are some observations after my reading of Section 2. Some suggested changes are in the form of diff comments.

- Some suggestions

```diff
- Sequence[T] is either a list or tuple
+ Sequence[T] is either a list or tuple of elements of type `T`.

- zeros(len: int) -> bytes returns an array of zero bytes. The length of output MUST be len
+ zeros(len: int) -> bytes returns an array of `len` bytes set to zero.

- gen_rand(len: int) -> bytes returns an array of random bytes generated by a cryptographically secure pseudorandom number generator (CSPRNG). The length of output MUST be len.
+ gen_rand(len: int) -> bytes returns an array of `len` bytes generated by a cryptographically secure pseudorandom number generator (CSPRNG).

- byte(int: int) -> bytes returns the representation of int as a byte string. The value of int MUST be in [0,256).
+ byte(n: int) -> bytes returns the representation of an integer `n` in the range [0,256) as a byte string.

- front(length: int, vec: list[Any]) -> (list[Any], list[Any]) splits vec into two vectors, where the first vector is made up of the first length elements of the input.
+ front(length: int, vec: list[Any]) -> (list[Any], list[Any]) splits vec into two vectors, where the first vector is made up of the first length elements of the input, and the second vector contains the remainder elements. This function is equivalent to return `(vec[:length], vec[length:])` in Python.
```

- The `to_be_bytes` and `from_be_bytes` functions correspond to I2OSC and OS2IP defined in [RFC8017]. This covers the big-endian case, but I couldn't find a reference for the little-endian functions. [RFC7748] offers Python snippet only:

```python
def decodeLittleEndian(b, bits):
    return sum([b[i] << 8*i for i in range((bits+7)/8)])```
```

- The definition of `to_be_bytes` and `to_le_bytes` does not specify how many bytes are returned.

- The `from_be_bytes` and `from_le_bytes` does not specify what are the valid encodings. For example:

```python
 5 = from_be_bytes([0,5])
   = from_be_bytes([0,0,5])
   = from_be_bytes([0,0,0,5])
```

all these are valid encodings of five. Also, does an empty `encoding` maps to zero?

- The `num_shares` and `field` parameters don't play role in the definition of `additive_secret_share` function.

- more sugestions

```diff
- cast(typ: type, val: object) -> object
+ cast(typ: type, val: object) -> typ

- range(stop) or range(start, stop[, step])
+ range(stop: int) or range(start: int, stop: int[, step: int])

- poly_eval(field: type[F], p: list[F], x: F) -> F returns the result of evaluating the polynomial,
+ poly_eval(field: type[F], p: list[F], x: F) -> F returns `p(x)`, the result of evaluating the polynomial `p` on `x`.

- p(input[k]) == output[k] for all k.
+ p(inputs[k]) == outputs[k] for all k.

- poly_strip(field: type[F], p: list[F]) -> list[F] strips the zeros from the end of the input polynomial.
+ poly_strip(field: type[F], p: list[F]) -> list[F] continuously reduces the degree of the input polynomial until its leading coefficient is non-zero, and returns the resultant polynomial.
```

- In `poly_eval`, `poly_interp`, `poly_mul`, and `poly_strip` functions, it is not described the purpose of the `field` parameter.

cjpatton:
Closes #488.
cc @armfazh 

These changes are mostly for clarity, but there are a few things we had underspecified:

* The length of the output of the little-endian and big-endian encoders was not specified.

* How the length of the little-endian and big-endian decoders was interpreted was not specified.

* the length of the output of `additive_secret_shares()` was ambiguous.


draft-irtf-cfrg-vdaf.md
```diff
@@ -823,14 +826,16 @@ Some common functionalities:
...
 
 * `to_le_bytes(x: int, len: int) -> bytes` converts an integer `x` whose value
-  is in the range `[0, 2^(8*len))` to little-endian bytes.
+  is in the range `[0, 2^(8*len))` to a little-endian, `len`-byte string.
 
 * `poly_eval(field: type[F], p: list[F], x: F) -> F` returns the result of
   evaluating the polynomial, `p(x)`. The coefficients of polynomials are stored
-  in lists in ascending order of degree, starting with the constant coefficient.
+  in lists in ascending order of degree, starting with the constant
+  coefficient. The `field` parameter is the class object for `F` and is used by
```
armfazh, line 837: I consider the parameter `field` is just an implementation detail, so it may not be necessary to be included in the signature of the function.

draft-irtf-cfrg-vdaf.md
```diff
@@ -799,15 +799,18 @@ Some common functionalities:
 * `concat(parts: list[bytes]) -> bytes` returns the concatenation of the input
   byte strings, i.e., `parts[0] + ... + parts[len(parts)-1]`.
 
-* `from_be_bytes(encoded: bytes) -> int` computes the inverse of
-  `to_be_bytes()`.
+* `from_be_bytes(encoded: bytes) -> int` decodes a big-endian byte string,
+  i.e., returns the integer `x` for which `to_be_bytes(x, len(encoded)) ==
```
armfazh, line 803:
```suggestion
  i.e., returns the integer `x` for which `to_be_bytes(x, len(encoded)) =
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -784,8 +784,8 @@ a particular encoding of that quantity as a byte string.
 Some common functionalities:
 
 * `additive_secret_share(x: list[F], num_shares: int, field: type[F]) ->
```
armfazh, line 786: `field: type[F]`
I consider this parameter is not needed, as it's an implementation detail.
This comment also applies to any other function with this parameter.

draft-irtf-cfrg-vdaf.md
cjpatton: We include it for completeness: when we use `poly_eval()` in the reference code, we must also provide the class object.

draft-irtf-cfrg-vdaf.md
```diff
@@ -799,15 +799,18 @@ Some common functionalities:
 * `concat(parts: list[bytes]) -> bytes` returns the concatenation of the input
   byte strings, i.e., `parts[0] + ... + parts[len(parts)-1]`.
 
-* `from_be_bytes(encoded: bytes) -> int` computes the inverse of
-  `to_be_bytes()`.
+* `from_be_bytes(encoded: bytes) -> int` decodes a big-endian byte string,
```
armfazh, line 802: nit: `to_be_bytes` should be defined before `from_be_bytes`

draft-irtf-cfrg-vdaf.md
cjpatton: Actually, I intended to use `==`, i.e., Python's notation for equality testing. I changed `from_le_bytes()` to be consistent with this.

draft-irtf-cfrg-vdaf.md
cjpatton: This is is alphabetized so that it can be referenced more easily.

draft-irtf-cfrg-vdaf.md
armfazh: I see two problems with this: `type[F]` 
1) notation of this expression is not defined.
2) A function with this signature `poly_eval(field: type[F], p: list[F], x: F)`
implies that the function can be invoked either:

```
poly_eval( F64, [ F64(1), F64(2)] , F64(3) )
```
or 
```
poly_eval( F32, [ F64(1), F64(2)] , F64(3) )
```

One possible solution is to mention at the beginning that all functions are parametrized with the field, or that the field is fixed for the entire scheme.

cjpatton: > I see two problems with this: `type[F]`
> 
>     1. notation of this expression is not defined.

I've tried to define this, PTAL at the PR once more.

>     2. A function with this signature `poly_eval(field: type[F], p: list[F], x: F)`
>        implies that the function can be invoked either:
> 
> 
> ```
> poly_eval( F64, [ F64(1), F64(2)] , F64(3) )
> ```
> 
> or
> 
> ```
> poly_eval( F32, [ F64(1), F64(2)] , F64(3) )
> ```
> 
> One possible solution is to mention at the beginning that all functions are parametrized with the field, or that the field is fixed for the entire scheme.

Python's type checker prevents the second case: `F` is a generic parameter, so the type checker checks that `field` is the class object for `F` and that each element of the input list is an instance of `F`.

armfazh: it's ok, I needed this reference
https://docs.python.org/3/library/typing.html#typing.Type

armfazh (approved):

divergentdave (approved):

PR #505, opened 2024-10-16
Add a common base class to Daf and Vdaf

### LINKED ISSUE #490: Create a base class for `Daf` and `Vdaf`
Sharding, aggregation, and unsharding are meant to be identical, but we don't currently enforce this in the API.

Create a base class for sharding, aggregation, and unsharding of which`Daf` and `Vdaf` are a subclass that define different APIs for preparation.

divergentdave:
Closes #490.

### DIFF
--- a/poc/vdaf_poc/daf.py
+++ b/poc/vdaf_poc/daf.py
@@ -1,7 +1,7 @@
 """Definition of DAFs."""
 
 from abc import ABCMeta, abstractmethod
-from typing import Generic, TypeVar
+from typing import Generic, TypeVar, override
 
 from vdaf_poc.common import gen_rand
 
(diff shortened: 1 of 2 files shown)


poc/vdaf_poc/daf.py
```diff
@@ -51,6 +104,7 @@ class Daf(
     # Number of random bytes consumed by `shard()`.
     RAND_SIZE: int
 
+    @override
```
cjpatton, line 107: What's the purpose of this decorator?

poc/vdaf_poc/daf.py
```diff
@@ -51,6 +104,7 @@ class Daf(
...
     @abstractmethod
     def shard(
```
cjpatton, line 109: Do we need to redefine these methods? I wonder if we can now remove `shard()` et al. since they're required by `Daf''s base class.

poc/vdaf_poc/daf.py
divergentdave: Static checkers will warn you if you have a method decorated with `@override` that is not actually overriding a superclass method. I added this as a bit of extra insurance against typos and inconsistent changes.

poc/vdaf_poc/daf.py
divergentdave: We could, but I think I'd prefer to have `Daf` and `Vdaf` be readable as a self-contained interface. Plus, they have decent docstrings, and I didn't bother writing any in the new ABC.

cjpatton (approved):

PR #508, opened 2024-10-16
Clean up references

### LINKED ISSUE #497: Weird formatting in references
There is a space after the "seriesinfo" field, e.g., "CRYPTO 2019 ,"

image

I'm not sure if this is intentional.

### LINKED ISSUE #503: Use full names for citations
I.e., `"name": "Robert Paulson"` rather than `"ins": "R. Paulson"`.
cc @martinthomson

cjpatton:
Closes #497.
Closes #503.

* Use "refcontent" instead of "seriesinfo". The value is a string rather than a bunch of key/value pairs. In effect, this deletes the " " before the ",".

* Put quotes around each "refcontent" value

* Replace "ins" with "name" and use the author's full name as it appears in the reference. `xml2rfc` still wants to abbreviate the first name, but we at least preserve more information.

* Add a "refcontent" for {{SML24}}.

* Reorder {{MRH04}} fields to match the other references

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -37,13 +37,14 @@ normative:
   AES:
     title: "Advanced Encryption Standard (AES)"
     author:
-      - ins: M. Dworkin
-      - ins: E. Barker
-      - ins: J. Nechvatal
-      - ins: J. Foti,
-      - ins: L. Bassham
-      - ins: E. Roback
-      - ins: J. Dray Jr.
+      - name: Morris J. Dworkin
+      - name: Elaine Barker
+      - name: James R. Nechvatal
+      - name: James Foti
+      - name: Lawrence E. Bassham
+      - name: E. Roback
+      - name: James F. Dray Jr.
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -37,13 +37,13 @@ normative:
...
-      - ins: L. Bassham
-      - ins: E. Roback
-      - ins: J. Dray Jr.
+      - name: Morris J. Dworkin
+      - name: Elaine Barker
+      - name: James R. Nechvatal
+      - name: James Foti
+      - name: Lawrence E. Bassham
+      - name: E. Roback
+      - name: James F. Dray Jr.
```
divergentdave, line 46: Maybe we should add the `ins` field here, since the heuristic gets it wrong
```suggestion
      - name: James F. Dray Jr.
        ins: J. Dray Jr.
```

divergentdave (approved):

draft-irtf-cfrg-vdaf.md
cjpatton: Done, and added a comment.

PR #513, opened 2024-10-17
Poplar1: Move prefix order check to is_valid()

### LINKED ISSUE #494: Poplar1: Move enforcement of prefix uniqueness and order to `is_valid()`

divergentdave:
Closes #494.

### DIFF
--- a/poc/tests/test_vdaf_poplar1.py
+++ b/poc/tests/test_vdaf_poplar1.py
@@ -220,9 +220,24 @@ class TestPoplar1(TestVdaf):
         ]
         self.assertFalse(cls.is_valid(agg_params[1], list(agg_params[:1])))
 
-        # Test `is_valid` rejects unsorted prefixes.
+        # Test `is_valid` rejects unsorted and duplicate prefixes.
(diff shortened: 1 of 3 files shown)


poc/tests/test_vdaf_poplar1.py
```diff
@@ -220,6 +220,12 @@ def test_is_valid(self) -> None:
         ]
         self.assertFalse(cls.is_valid(agg_params[1], list(agg_params[:1])))
 
+        # Test `is_valid` rejects unsorted prefixes.
+        agg_params = [
+            (0, (int_to_bit_string(0b1, 1), int_to_bit_string(0b0, 1))),
+        ]
+        self.assertFalse(cls.is_valid(agg_params[0], list(agg_params)))
```
cjpatton, line 227: Somewhat random thought here: A quirk in Python's implementation of `>=` on tuples might mean these are not actually in lexicographic order. It might be worth expanding this to longer vectors, in particular checking that ordering starts with the front of the tuple rather than the end.

PR #521, opened 2024-11-01
Editorial pass

divergentdave:
This rolls up a bunch of editorial changes, clarifications, and rephrasings from my pass over the document. I split it out into several topical commits, plus a couple omnibus commits of more minor changes.


cjpatton (approved):
Excellent. I think the only thing I slightly disagree with is MAY versus SHOULD for the the none-affineness of the gadget, given the semantics of SHOULD in RFCs. But I agree that it would be silly for this to be non-affine.

divergentdave:
Yeah, looking at RFC 2119, maybe this ought to be a lower-case "should". There's no valid reason for an affine gadget, but also an affine gadget wouldn't break anything.

PR #522, opened 2024-11-01
Rewrite mention of replay attacks in nonce section

divergentdave:
The current text here seems too permissive about processing a report more than once, given that two sections below we say that "DAFs and VDAFs may impose restrictions on the re-use of reports". This PR strengthens the text leading into the recommendation to make use of the nonce in replay protections.


draft-irtf-cfrg-vdaf.md
```diff
@@ -5692,11 +5692,9 @@ required in order to leverage security analysis for the privacy definition of
...
 is controlled by the attacker.
 
-Other security considerations may require the nonce to be non-repeating in a
-given context. For example, to achieve differential privacy it is necessary to
-avoid "over exposing" a report by including it too many times in a single batch
-or across multiple batches. It is RECOMMENDED that the nonce generated by the
-Client be used by the Aggregators for replay protection.
+Applications will need to protect against replay attacks to prevent disallowed
+re-use of reports (see {{agg-param-security}}). It is RECOMMENDED that the
+nonce generated by the Client be used by the Aggregators for replay protection.
```
cjpatton, line 5697: I'm fine with the stronger language, but I think we want to keep the mention DP consideration around.

cjpatton (approved):

PR #523, opened 2024-11-01
Mention that decode_from_bit_vec() is linear

divergentdave:
This mentions in a couple places that decoding a bit representation is a linear operation, and thus we can use it in validity circuits, transparently processing secret shares of bit representations.


cjpatton (approved):
Nice.

PR #532, opened 2025-01-07
Make the pictures pretty

### LINKED ISSUE #477: Pretty pictures
Time permitting, we should take advantage of the work done to render nice looking diagrams and illustrations in RFCs.

According to @martinthomson, all we have to do is change `~~~` to `~~~ aasvg` and add an npm package:
```bash
npm i --save aasvg; git add package.json
```

cjpatton:
Closes #477.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,5 +18,4 @@ jobs:
       run: "bash -c '! grep -r --file=rejected_dictionary draft-irtf-cfrg-vdaf.md poc'"
 
     - name: "Check for warnings emitted by xml2rfc"
-      run: |
-        bash -c -o pipefail 'make |& (! grep -E "Warning|Error")'
+      run: .github/workflows/make-with-lints.sh
--- a/.github/workflows/make-with-lints.sh
+++ b/.github/workflows/make-with-lints.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+make |& \
+	grep -v 'Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.' | \
+	(! grep -E "Warning|Error")
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -895,9 +895,9 @@ Some common functionalities:
 
 # Overview
 
-~~~
+~~~ aasvg
                  +--------------+
-           +---->| Aggregator 0 |----+
+           +---->| Aggregator 0 +----+
      input |     +--------------+    | aggregate
     shares |             ^           | shares
            |             |           |
@@ -906,7 +906,7 @@ Some common functionalities:
            | +-->| Aggregator 1 |--+ |
            | |   +--------------+  | |
 +--------+-+ |           ^         | +->+-----------+
-| Client |---+           |         +--->| Collector |--> aggregate
+| Client +---+           |         +--->| Collector +--> aggregate
 +--------+-+                         +->+-----------+
            |            ...          |
            |                         |
@@ -1029,25 +1029,27 @@ for the `PublicShare`, `InputShare`, `AggParam`, and `AggShare` types.
 
 ## Sharding {#sec-daf-shard}
 
-~~~~
+~~~~ aasvg
     Client
     ======
 
     measurement
       |
+      |
       V
     +----------------------------------------------+
     | shard                                        |
-    +----------------------------------------------+
+    +-+--------------+--------------+-----+--------+
       |              |              |     |
       |              |         ...  |    public_share
       |              |              |     |
       |    +---------|-----+--------|-----+
       |    |         |     |        |     |
-      V    |         V     |        V     |
+      v    |         v     |        v     |
      input_share_0  input_share_1  input_share_[SHARES-1]
+      |    |         |     |        |     |
       |    |         |     |   ...  |     |
-      V    V         V     V        V     V
+      v    v         v     v        v     v
     Aggregator 0   Aggregator 1    Aggregator SHARES-1
 ~~~~
 {: #shard-flow title="Illustration of the sharding algorithm."}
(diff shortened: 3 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -19,4 +19,8 @@ jobs:
 
     - name: "Check for warnings emitted by xml2rfc"
       run: |
-        bash -c -o pipefail 'make |& (! grep -E "Warning|Error")'
+        bash -c -o pipefail <<EOF
+make |& \
+  (! grep -E "Warning|Error)") | \
+  (! grep -v 'Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.')
```
divergentdave, line 25: We need to filter this message out first, since we're using the error code of the sub-shell inside the parentheses
```suggestion
          grep -v 'Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.' | \
          (! grep -E "Warning|Error)")
```

cjpatton: Hmm... I took your suggestion but I'm not sure it had the intended effect. Here's what I've gotten working on the command line:
```bash
make |& (! grep -E "Warning|Error") | (! grep -v 'Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.')
```

cjpatton: I'm trying to figure out how to plumb this into GH actions. I'll have to resume tomorrow however.

.github/workflows/make-with-lints.sh
```diff
@@ -0,0 +1,3 @@
+make |& \
```
divergentdave, line 1: We could move the option in here, and add the interpreter. We can also set the executable bit on the file in the git repository to avoid having to chmod the script before running it.
```suggestion
#!/bin/bash
set -o pipefail
make |& \
```

.github/workflows/make-with-lints.sh
```diff
@@ -0,0 +1,3 @@
...
+	(! grep -v 'Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.') | \
```
divergentdave, line 2: I don't think we should have this `(! ... )` around this grep invocation, because it negates the exit status, and causes a failure. See this abbreviated example:

```
$ make clean && bash -c -o pipefail "make |& (! grep -v 'SVG with width') | (! grep -E 'Warning|Error')" && echo ok || echo fail
rm -f .tags .targets.mk issues.json \
    draft-irtf-cfrg-vdaf.{txt,html,pdf} index.html \
    draft-irtf-cfrg-vdaf-[0-9][0-9].{xml,md,org,txt,raw.txt,html,pdf} \
    draft-irtf-cfrg-vdaf.xml \
     diff-draft-irtf-cfrg-vdaf.html
fail
```

.github/workflows/make-with-lints.sh
divergentdave: Since we are now invoking a separate script, I think we need to set the `-e` option as well in order to propagate failures back out to the parent shell. I think this is masking the unconditional failure introduced by `(! grep -v 'Warning: Found SVG ...')`.

draft-irtf-cfrg-vdaf.md
```diff
@@ -895,7 +895,7 @@ Some common functionalities:
 
 # Overview
 
-~~~
+~~~ aasvg
                  +--------------+
            +---->| Aggregator 0 |----+
```
divergentdave, line 900: We're missing a few more connections, here and on the other boxes in the middle
```suggestion
           +---->| Aggregator 0 +----+
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1371,31 +1375,34 @@ cross protocol attacks; see {{deep}}.
...
       |       |      |       |         ...  |       |     |
-      V       V      V       V              V       V     | x ROUNDS
+      v       v      v       v              v       v     | x ROUNDS
     +-----------+  +-----------+          +-----------+   |
     | prep_next |  | prep_next |          | prep_next |   |
-    +-----------+  +-----------+          +-----------+   |
+    +-+-------+-+  +-+-------+-+          +-+-------+-+   |
       |       |      |       |         ...  |       |     |
-      V       V      V       V              V       V     /
+      v       v      v       v              v       v     /
```
divergentdave, line 1399: We could make this a curved line as follows:

```
    +---------|--------------|----------------------|--+  -.
    |         |              |  prep_shares_to_prep |  |    |
    +-+-------|------+-------|--------------+-------|--+    |
      |       |      |       |         ...  |       |       |
      v       v      v       v              v       v       | x ROUNDS
    +-----------+  +-----------+          +-----------+     |
    | prep_next |  | prep_next |          | prep_next |     |
    +-+-------+-+  +-+-------+-+          +-+-------+-+     |
      |       |      |       |         ...  |       |       |
      v       v      v       v              v       v     -'
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3588,19 +3595,21 @@ def verifier_len(self) -> int:
...
+| prove()            |
+|  +-------------+   |
+|  | Valid        |<----- meas
+|  |  +--------+  |<----- joint rand
+|  |  | Gadget |  |  |<-- prove rand
+|  |  +-------+   |  |
+|  |              |  |
+|  +--------------+  |
+|                    |
++-+-----------------+
```
divergentdave, line 3609: These aren't quite aligned, and turned into rounded corners
```suggestion
|  +--------------+  |
|  | Valid        |<----- meas
|  |  +--------+  |<----- joint rand
|  |  | Gadget |  |  |<-- prove rand
|  |  +--------+  |  |
|  |              |  |
|  +--------------+  |
|                    |
+-+------------------+
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -3666,19 +3675,22 @@ def prove(self,
...
+| query()  |         |
+|  +-------|-----+   |
+|  | Valid |      |<----- meas (share)
+|  |  +----v---+  |<----- joint rand
+|  |  | Gadget |  |  |<-- query rand
+|  |  +-------+   |  |
+|  |              |  |
+|  +--------------+  |
+|                    |
++-+-----------------+
```
divergentdave, line 3691: Same here
```suggestion
|  +-------|------+  |
|  | Valid |      |<----- meas (share)
|  |  +----v---+  |<----- joint rand
|  |  | Gadget |  |  |<-- query rand
|  |  +--------+  |  |
|  |              |  |
|  +--------------+  |
|                    |
+-+------------------+
```

.github/workflows/make-with-lints.sh
cjpatton: Added `set -e`. However I tested this by modifying the match string, and it caused the action to fail as expected.

draft-irtf-cfrg-vdaf.md
cjpatton: I was being artistic, lol. I'm fine making these square.

draft-irtf-cfrg-vdaf.md
cjpatton: nit: I changed the number of columns between the curved line and the rest of the figure from 2 to 1, in order to accommodate the artwork width requirements.

divergentdave (approved):

PR #533, opened 2025-01-07
Update Change Log for draft 14

cjpatton:


draft-irtf-cfrg-vdaf.md
```diff
@@ -432,6 +432,11 @@ particular.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+14:
+
+* Poplar1: Require trailing bits of each prefix of the aggregation parameter to
+  be cleared.
```
divergentdave, line 438: We should make it clear this is about the serialization format, and not the candidate prefixes themselves.
```suggestion
* Poplar1: When decoding an aggregation parameter, require the padding bits
  after each prefix to be cleared.
```

divergentdave (approved):

PR #537, opened 2025-01-24
Silence SVG warning via front matter

divergentdave:
This adds some configuration to the front matter to silence the SVG warning we're getting, and cleans up the shell script we were using to filter these out previously. See https://mailarchive.ietf.org/arch/msg/rfc-markdown/reUEfraLI_-XH7fO0wKe2ErFxno/.

This works locally, but I think we'll need to wait for a new Docker image to be built that includes kramdown-rfc 1.7.22 before it'll work in CI.

### DIFF
--- a/.github/workflows/make-with-lints.sh
+++ b/.github/workflows/make-with-lints.sh
@@ -2,6 +2,4 @@
 
 set -e
 
-make |& \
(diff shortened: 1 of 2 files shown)


divergentdave:
The Docker image has been updated, and this works now.

cjpatton (approved):

PR #538, opened 2025-04-22
Negative test vectors

### LINKED ISSUE #536: Add test vectors for preparation failures
[list feedback](https://mailarchive.ietf.org/arch/msg/cfrg/YY8LvTfVwZUHiypnWKV467ITtaw/):

It should be pretty easy to modify `gen_test_vec.py` to output tweaked test vectors that trigger preparation to fail. For example:
* Tweak the nonce
* Tweak the first byte of the public share
* Tweak the first byte of a random input share
* Tweak the first byte of a random prep share
* Tweak the first byte of the prep message

divergentdave:
This extends the test vector file format to support negative test vectors. Each file now includes a list of which operations should be run using its messages, and whether they should succeed or fail. Thus, we can represent a report that could never be produced by the sharding algorithm, skip comparing against the shard implementation, and check what happens in each step of the preparation process with that input. I include eight new negative test vectors to start with. Existing test vectors are modified to add this list of operations. Note that test vector consumers do not necessarily need to reimplement `run_vdaf()` anymore to make use of the all-success test vectors, since the operations list now captures that high-level coordination.

Up until now, we have not had strong conventions for "round numbers" to associate with every message or operation. I'm open to suggestions on whether we should adjust the round numbers used for "prep_shares_to_prep" or "prep_next" operations.

I'm working on consuming these negative test vectors in libprio-rs, and thus far I'm getting partial mismatches of Prio3 prepare shares when either the helper's joint randomness blind or the public share is corrupted (but not the leader's joint randomness blind, interestingly). It's possible we have some error in the implementation that cancels out when there's agreement on the joint randomness seed.

Closes #536.

### DIFF
--- a/poc/gen_test_vec.py
+++ b/poc/gen_test_vec.py
@@ -2,11 +2,17 @@
 
 import json
 import os
+from typing import Any, cast
 
 from vdaf_poc.common import print_wrapped_line
 from vdaf_poc.field import Field64, Field128
-from vdaf_poc.idpf import Idpf
-from vdaf_poc.test_utils import gen_test_vec_for_vdaf, test_vec_gen_rand
+from vdaf_poc.idpf import FieldVec, Idpf
+from vdaf_poc.test_utils import (VdafTestVectorDict, gen_test_vec_for_vdaf,
+                                 pretty_print_vdaf_test_vec, test_vec_gen_rand,
+                                 write_test_vec)
+from vdaf_poc.vdaf_poplar1 import Poplar1, Poplar1InputShare
+from vdaf_poc.vdaf_prio3 import (Prio3, Prio3Count, Prio3Histogram,
+                                 Prio3InputShare)
 from vdaf_poc.xof import Xof
 
 # The path where test vectors are generated.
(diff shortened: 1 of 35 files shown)


divergentdave:
The mismatch was due to an issue in the proof of concept implementation: `Prio3.prep_init()` was modifying the list passed in as the public share when it 'corrected' joint randomness parts. After fixing that and regenerating test vectors, everything matches with the Rust implementation.

cjpatton (commented):
The semantics of operations is unclear to me:

> Each file now includes a list of which operations should be run using its messages, and whether they should succeed or fail.

What does it mean "to run [the test vector] using its messages"? Here's my interpretation, taking the public share as an example:
1. For a positive test, we'd shard and check if the public share matches the test vector: if not, the test fails.
2. For a negative test, we'd shard, overwrite the public share with the value from the test vector, and continue. (This expresses a non-compliant sharding algorithm.) If preparation succeeds, then the test fails.

Is this interpretation correct? If so:

1. Do we really need a sequence of operations, or is it sufficient to just express 
2. What about non-compliance that wouldn't affect preparation, like aggregation and unsharding? Are these expressible? Should they be?


> Up until now, we have not had strong conventions for "round numbers" to associate with every message or operation. I'm open to suggestions on whether we should adjust the round numbers used for "prep_shares_to_prep" or "prep_next" operations.

Do you think this is underspecified in the draft? IIUC the PR currently matches the notion of rounds in the draft, which I think is what we should do.

divergentdave:
> Here's my interpretation, taking the public share as an example:
> 
> 1. For a positive test, we'd shard and check if the public share matches the test vector: if not, the test fails.
> 
> 2. For a negative test, we'd shard, overwrite the public share with the value from the test vector, and continue. (This expresses a non-compliant sharding algorithm.) If preparation succeeds, then the test fails.

The negative test vector here that tampers with the public share does not include "shard" in its list of operations, thus the intent is that the tests consuming the test vector would skip sharding entirely. If the test still sharded a report, we would need to either supply both the public share that sharding should output and the public share that preparation should use, or indicate that the output of sharding should not be checked against any value from the test vector.

> 2. What about non-compliance that wouldn't affect preparation, like aggregation and unsharding? Are these expressible? Should they be?

I think so, you could provide a tampered output share in the `out_shares` list, and only include aggregation and unsharding in the list of operations, for example. This might be less interesting, since aggregation only exercises some form of decoding and field vector addition for our VDAFs, and unsharding is straightforward preprocessing. But you could check what happens if an output share or aggregate share includes an integer larger than the field modulus, or if the message has the wrong length.

> Do you think this is underspecified in the draft? IIUC the PR currently matches the notion of rounds in the draft, which I think is what we should do.

I settled on `prep_init()` being in round 0, and `prep_next()` round numbers starting at 1. I wasn't sure whether `prep_shares_to_prep()` should be thought of as the last step in one round, or the first step in the next round.

poc/gen_test_vec.py
```diff
@@ -96,6 +102,421 @@ def gen_test_vec_for_xof(cls: type[Xof]) -> None:
...
+    write_test_vec(
+        test_vec_path,
+        test_vec,
+        vdaf.test_vec_name,
+        filename_suffix,
+    )
+
+
+def gen_poplar1_negative_test_vec(test_vec_path: str) -> None:
+    "Generates various negative test vectors for Poplar1"
```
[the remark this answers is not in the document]
divergentdave: I can change it for consistency, but either form is a valid docstring.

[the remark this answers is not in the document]
cjpatton: No need, thanks for clarifying.

poc/vdaf_poc/test_utils.py
```diff
@@ -69,18 +71,88 @@ def run_vdaf_test(
...
 
+class VdafTestVectorOperationDict(TypedDict):
+    """
+    A description of one of the operations that should be performed with this
+    test vector.
+
+    Attributes:
+        operation: The type of operation to be performed. This is one of
+            "shard", "prep_init", "prep_shares_to_prep", "prep_next",
+            "aggregate", or "unshard".
```
cjpatton, line 82: "aggregate" corresponds to `agg_init()` followed by `agg_update()`, correct? We should spell this out.

poc/vdaf_poc/test_utils.py
```diff
@@ -69,18 +71,88 @@ def run_vdaf_test(
...
+            determines which prepare share, prepare state, and/or prepare
+            message to use.
+        aggregator_id: The aggregator ID to use when performing this
+            operation. This determines which messages and which prepare state
+            to use, in addition to the aggregator ID argument itself.
+        report_index: The index of the report on which to perform this
+            operation. This is an index into the `prep` array.
+        success: If this is true, the operation should succeed, and its output
+            should match the corresponding values in the test vector. If this
+            is false, the operation should fail.
```
cjpatton, line 93: What about cases where the operation succeeds but the output is not the same? I'm thinking of "aggregate" for instance where a bad implementation of `agg_update()` could result in an incorrect aggregate share

poc/vdaf_poc/test_utils.py
```diff
@@ -69,18 +71,88 @@ def run_vdaf_test(
...
 
 class VdafTestVectorDict(Generic[Measurement, AggParam, AggResult], TypedDict):
+    """
+    A test vector for evaluation of a VDAF on one or more reports.
+
+    All VDAF messages are encoded to byte strings, and then hex-encoded.
+
+    Attributes:
+        operations: A list of operations that should be executed using
+            messages from this test vector.
```
cjpatton, line 144: A couple things we should clarify:

1. The order of operations: run the commands from `operations[0]` to `operations[-1]`.
2. When an input is needed from a previous operation. For example, to run `prep_next()`, you need to know the aggregator's prep state, which is not included in the test vector. We might resolve this by combining all of the "prep_*" commands into a single command "prep" like we do for "aggregate".
3. Whether there are "illegal" programs. For instance, is it valid to have an operation following one that fails to succeed?

order of operations is unclear
some "programs" are illegal (no operations allowed after failure)

cjpatton (commented):
> The negative test vector here that tampers with the public share does not include "shard" in its list of operations, thus the intent is that the tests consuming the test vector would skip sharding entirely. If the test still sharded a report, we would need to either supply both the public share that sharding should output and the public share that preparation should use, or indicate that the output of sharding should not be checked against any value from the test vector.

I see! So the operations list is meant to be a recipe for how to consume the test vector. That makes more sense. I think the algorithm is a little unclear, but I'll add some concrete suggestions inline. It might be worth actually implementing the algorithm that consumes test vectors. As a bonus, we could have unit tests that make sure our test vectors actually match the reference code!
 
My primary bit of feedback is that it would be nice for this change to be backwards compatible so that we don't have to specify this algorithm in the draft. WDYT about making "operations" optional? If absent, the behavior would be to treat the vector as a positive test.

1. We should describe how 

> > 2. What about non-compliance that wouldn't affect preparation, like aggregation and unsharding? Are these expressible? Should they be?
> 
> I think so, you could provide a tampered output share in the `out_shares` list, and only include aggregation and unsharding in the list of operations, for example. This might be less interesting, since aggregation only exercises some form of decoding and field vector addition for our VDAFs, and unsharding is straightforward preprocessing. But you could check what happens if an output share or aggregate share includes an integer larger than the field modulus, or if the message has the wrong length.

It would be nice to be able to express this kind of thing, but not essential for this PR.

> I settled on `prep_init()` being in round 0, and `prep_next()` round numbers starting at 1. I wasn't sure whether `prep_shares_to_prep()` should be thought of as the last step in one round, or the first step in the next round.

I see. I could go either way, though the latter seems to correspond to how we illustrate this in the draft: https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-5.2-1.1.1

poc/vdaf_poc/test_utils.py
divergentdave: I think this is covered, by "and its output should match..."

divergentdave (commented):
> It might be worth actually implementing the algorithm that consumes test vectors. As a bonus, we could have unit tests that make sure our test vectors actually match the reference code!

I'd like to punt this to a follow-up, because this will require writing decoding routines to mirror our test vector encoding routines.

> My primary bit of feedback is that it would be nice for this change to be backwards compatible so that we don't have to specify this algorithm in the draft. WDYT about making "operations" optional? If absent, the behavior would be to treat the vector as a positive test.

The test vectors are backwards compatible in that a consumer that ignores the `operations` list of a positive test vector can still use it as-is with their `run_vdaf()`-style orchestration. You would just need to filter out the negative test vectors by hand.

FWIW, I spent some time looking through a code coverage report for other interesting conditions to cover, and the only one I found was the "joint randomness mismatch" error in Prio3's prep_next(). I spent some time thinking about how to test this a while back, and couldn't come up with a way to craft a report to hit it, short of defining a degenerate circuit that takes in a joint randomness value and never passes it into a gadget. Instead, we could define a separate test vector that only runs the prep_init and prep_next operations on one aggregator, with a crafted prepare message.

cjpatton (approved):
> > It might be worth actually implementing the algorithm that consumes test vectors. As a bonus, we could have unit tests that make sure our test vectors actually match the reference code!
> 
> I'd like to punt this to a follow-up, because this will require writing decoding routines to mirror our test vector encoding routines.

SGTM. I'm also fine if you don't want to bother doing this.

> > My primary bit of feedback is that it would be nice for this change to be backwards compatible so that we don't have to specify this algorithm in the draft. WDYT about making "operations" optional? If absent, the behavior would be to treat the vector as a positive test.
> 
> The test vectors are backwards compatible in that a consumer that ignores the `operations` list of a positive test vector can still use it as-is with their `run_vdaf()`-style orchestration. You would just need to filter out the negative test vectors by hand.

Fair enough, but I think we'll then need to explain how to handle `"operations"` in the draft. We'll need to bump the test vector reference in a follow-up PR in any case.

PR #543, opened 2025-05-05
topo: Merge outbound into state

### LINKED ISSUE #542: Type safety for return values
I see that, for instance `ping_pong_leader_continued` returns:

```
           tuple[State, Optional[bytes]]:
```

But this allows, for instance, this function to return `(Rejected(), [stuff])` (although the body of the function doesn't seem to do so) and assuming I understand the situation, is actually not a valid situation. Moreover, it makes the rules for how to process the response are confusing. For instance, here is some text from DAP  S 4.6.2.1 on how to handle this function

```
       If outbound != None, then the Leader stores state and outbound
       and proceeds to Section 4.6.3.1.  If outbound == None, then the
       preparation process is complete: either state == Rejected(), in
```

So if `outbound != None` I'm not even supposed to look at `state` at this point. Is that really right?

I would suggest that instead we should adopt some sort of type safety mechanism. My Python type hinting is a little rusty but Gemini seems to think what  we want is something like:

```
class ContinuingData(NamedTuple):
    state: bytes
    output: bytes

class Finished(NamedTuple):
    output: bytes,
   

State = Literal["Rejected"] | Finished | ContinuingData
```

This will make it impossible to have invalid states.

cjpatton:
Closes #542.

Merge the outbound message into the Aggregator's ping pong state:

1. Remove outbound message from each API call

2. Add outbound message to `Continued`

3. Define a new state, `FinishedWithOutbound` that is like `Finished` except it include an outbound message

This reduces the number of cases the caller has to handle to only cases that are valid. In particular, the following cases are invalid:

* `state == Rejected() and outbound != None`
* `state == Continued(...) and outbound == None`

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -1614,22 +1614,24 @@ channel.
 The state machine of each Aggregator is shown below.
 
 ~~~ aasvg
-                  +----------------+
-                  |                |
-                  v                |
-Start ----> Continued(prep_state, prep_round) --> Finished(out_share)
- |                |
- |                |
- +--> Rejected <--+
+ +--> Rejected <--+   +----------------+   Finished(out_share)
+ |                |   |                |           ^
+ |                |   |                v           |
+Start ----> Continued(prep_state, prep_round, outbound)
+ |                                                 |
+ |                                                 v
+ +-----------------> FinishedWithOutbound(out_share, outbound)
 ~~~
 {: #vdaf-prep-state-machine title="State machine of VDAF preparation."}
 
 State transitions are made when the state is acted upon by the Aggregator's
 local inputs and/or messages sent by its co-Aggregators. The initial state is
 `Start`. The terminal states are: `Rejected`, indicating that the report cannot
-be processed any further; and `Finished(out_share)`, indicating that the
-Aggregator has recovered an output share `out_share`. For completeness, we
-define these states in {{topo-states}}.
+be processed any further; `Finished(out_share)`, indicating that the
+Aggregator has recovered an output share `out_share`; and
+`FinishedWithOutbound(out_share, outbound)`, indicating that the Aggregator has
+recovered an output share, and has one more outbound message to send. For
+completeness, we define these states in {{topo-states}}.
 
 The methods described in this section are defined in terms of opaque byte
 strings. A compatible `Vdaf` MUST specify methods for encoding public shares,
@@ -1709,7 +1711,7 @@ def ping_pong_leader_init(
         agg_param: bytes,
         nonce: bytes,
         public_share: bytes,
-        input_share: bytes) -> tuple[State, Optional[bytes]]:
+        input_share: bytes) -> Continued | Rejected:
     """Called by the Leader to initialize ping-ponging."""
     try:
         (prep_state, prep_share) = self.prep_init(
@@ -1721,20 +1723,22 @@ def ping_pong_leader_init(
             self.decode_public_share(public_share),
             self.decode_input_share(0, input_share),
         )
+
         encoded_prep_share = self.encode_prep_share(prep_share)
-        return (
-            Continued(prep_state, 0),
+        return Continued(
+            prep_state, 0,
             encode(0, encoded_prep_share),  # initialize
         )
     except:
-        return (Rejected(), None)
+        return Rejected()
 ~~~
 
-The output is the `State` to which the Leader has transitioned and an encoded
-`Message`. If the Leader's state is `Rejected`, then processing halts.
-Otherwise, if the state is `Continued`, then processing continues. The function
-`encode`  is used to encode the outbound message, which has the message type of
-`initialize` (identified by the number `0`).
+The output is the `State` to which the Leader has transitioned. If the Leader's
+state is `Rejected`, then processing halts. Otherwise, if the state is
+`Continued`, then processing continues. In this case, the state also includes
+the Leader's outbound message. The function `encode` is used to encode the
+outbound message, which has the message type of `initialize` (identified by the
+number `0`).
 
 To continue processing the report, the Leader sends the outbound message to the
 Helper. The Helper's initial transition is computed using the following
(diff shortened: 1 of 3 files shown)


cjpatton:
cc @ekr

draft-irtf-cfrg-vdaf.md
```diff
@@ -6279,6 +6284,16 @@ class Finished(State, Generic[OutShare]):
         return isinstance(other, Finished) and \
             self.out_share == other.out_share
 
+class FinishedWithOutbound(State, Generic[OutShare]):
```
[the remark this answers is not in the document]
cjpatton: Done.

tgeoghegan:
As promised, I've worked up a couple different alternatives here, and have some Python and Rust implementations to help support discussion.

First, we have the spec change in this PR (#543). This same PR implements it in Python, and I've implemented it in Rust [here](https://github.com/divviup/libprio-rs/pull/1256). To recap my earlier comments, my major gripe here is that various illegal states are still representable under this API, and as a result, both the ping pong implementation and its users have to do more error checking and structure unpacking. For example, `leader_initialized` can only ever return `PingPongState::Initialize`, but because it returns `PingPongState`, users have to check that the struct is of the desired variant and pull out the `A::PrepareState`.

Additionally, if we look at `helper_initialize`, it takes a `PingPongMessage`, but fails if it's anything but `PingPongMessage::Initialize`. So it could just take `A::PrepareState` and save everyone some error handling. Similarly, `helper_continue` takes `PingPongState`, but barfs if the state is anything other than `PingPongState::Continue`. So we might as well just pass in `(A::PrepareState, PingPongMessage)` (which is what's inside a `PingPongState::Continue`).

I made a PR (#545) which I think is just a bit better. Take a look at that change for some detailed discussion, but basically it's a minor tweak to narrow what `PingPongState` variants are returned or taken by various functions. I've implemented it in Rust in https://github.com/divviup/libprio-rs/pull/1259. Several functions now either take or return a `Self::PrepareState` instead of `PingPongState`, which means the compiler can enforce that values of the right type are getting passed around.

This try retains the `Initialize` variant of the ping pong message, because @cjpatton observed that ideally, we'd do this in a way that doesn't break the wire format. But I think we can do better still if we're willing to delete that message variant. I've implemented that in Rust in https://github.com/divviup/libprio-rs/pull/1258. I don't have a Python implementation for it, because it's a little more awkward to do (because the Python implementation needs an explicit round counter to decide if preparation is finished, and if you're not passing a `Continued` around, there's no obvious place to stick that integer). But the Rust side shows that we get to delete even more error cases, because for instance it's no longer possible to pass `PingPongMessage::Continued` to `helper_initialized`.

To summarize:

| | Wire message break? | Python implementation | Rust implementation |
|-|-|-|-|
| put ping pong message inside ping pong state | No | #543 | https://github.com/divviup/libprio-rs/pull/1256 |
| put ping pong message inside ping pong state, making illegal states impossible to represent | No | #545 | https://github.com/divviup/libprio-rs/pull/1259 |
| put ping pong message inside ping pong state, making illegal states impossible to represent, AND removing `PingPongMessage::Initialize` | Yes | none yet | https://github.com/divviup/libprio-rs/pull/1258 |

My vote is to go with the last option, which admittedly is more disruptive to implementations, but I think all the known implementations have work left to do in order to implement the latest DAP anyway.

poc/tests/test_vdaf_ping_pong.py
```diff
@@ -185,35 +186,35 @@ def test_one_round(self) -> None:
...
-        (helper_state, msg) = vdaf.ping_pong_helper_init(
+        helper_state = vdaf.ping_pong_helper_init(
             verify_key,
             ctx,
             vdaf.encode_agg_param(agg_param),
             nonce,
             vdaf.test_vec_encode_public_share(public_share),
             vdaf.test_vec_encode_input_share(input_shares[1]),
-            cast(bytes, msg),
+            cast(bytes, leader_state.outbound),  # XX cast needed?
```
divergentdave, line 207: This indeed doesn't need a cast anymore, since we narrow `leader_state` to `Continued` above, and it is apparent from the constructor that the `outbound` attribute is of type `bytes`.

divergentdave (commented):
Some editorial comments:

poc/tests/test_vdaf_ping_pong.py
```diff
@@ -236,43 +237,58 @@ def test_multi_round(self) -> None:
                 rand,
             )
 
-            (leader_state, msg) = vdaf.ping_pong_leader_init(
+            leader_init_state = vdaf.ping_pong_leader_init(
```
divergentdave, line 240: By declaring this variable to be of a broader type, we can skip using a separate variable name, and get rid of the cast below.
```suggestion
            leader_state: State = vdaf.ping_pong_leader_init(
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -1833,7 +1837,7 @@ def ping_pong_leader_continued(
     agg_param: bytes,
     state: State,
     inbound: bytes,
-) -> tuple[State, Optional[bytes]]:
+) -> State:
```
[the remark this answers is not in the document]
tgeoghegan: Yeah, I still need to reflect the changes in the poc code in the main specification, and make any necessary text changes. I can tack on a commit to that effect to this PR.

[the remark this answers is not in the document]
tgeoghegan: `ping_pong_leader` continued can return any of `Continued | Finished | FinishedWithOutbound | Rejected`. That does exclude `Start`, so enumerating the other four would be more specific. However, the `Start` state is never instantiated anywhere (both leader and helper are in state `Continued` after their respective `init` calls) so we could also just remove that class altogether.

poc/tests/test_vdaf_ping_pong.py
tgeoghegan: Excellent tip, thank you!

cjpatton: Done.

tgeoghegan (approved):
Looks cromulent to me (not that anyone asked, I am not a VDAF editor!)

tgeoghegan:
Chris and I chatted offline and agreed that the second row of my table is the way forward: it better expresses what is and isn't legal without breaking the wire format for messages, which the VDAF editors are reluctant to do at this stage. The further safety and expressiveness changes I sketched out can still be achieved in an implementation like libprio-rs while still conforming to the specification, so that's good enough for me.

poc/tests/test_vdaf_ping_pong.py
```diff
@@ -236,43 +237,52 @@ def test_multi_round(self) -> None:
...
                 if isinstance(leader_state, Continued):
-                    (leader_state, msg) = vdaf.ping_pong_leader_continued(
+                    assert isinstance(helper_state, Continued) or \
+                        isinstance(helper_state, FinishedWithOutbound)
+                    leader_state = vdaf.ping_pong_leader_continued(
                         vdaf.encode_agg_param(agg_param),
                         ctx,
-                        leader_state,
-                        cast(bytes, msg),
+                        cast(Continued, leader_state),
```
divergentdave, line 279: This cast can be removed, because the enclosing `if` narrows the inferred type here.
```suggestion
                        leader_state,
```

cjpatton:
Squashed. I'll hold off on merging until we're done reviewing https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/705.

PR #545, opened 2025-05-15
Ping pong topo

tgeoghegan:
This adds one more commit on top of the changes in #543 that I think make the API easier to use, in particular by making illegal states impossible to represent, which I believe is the overall goal here. I've left remarks throughout the diff to highlight important changes and the motivation.

I also haven't yet reflected this proposed change in the actual protocol text. It's only in the Python `poc` for now, so we can discuss whether we like it. Based on the consensus we reach, I can fix up the protocol text accordingly.


cjpatton (approved):
This is strictly better, so happy to take this!

PR #558, opened 2025-08-15
Rename robustness to verifiability

### LINKED ISSUE #551: Crypto panel: robustness => verifiability
> - Section 1 explains the V=verifiability property of VDAFs and then says
> that it is referred to as "robustness". Why not call it verifiability,
> i.e., why having two different names for the same thing?

The term "robustness" was borrowed from the original Prio paper and it just kind of stuck. Having one name instead of two is a good idea.

https://mailarchive.ietf.org/arch/msg/cfrg/Omdhr4rO1pla_nlju2l7OJEGWPM/

cjpatton:
Closes #551.

This avoids using two terms for the same security property.


tgeoghegan (approved):
Looks sound to me! I filed https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/724 to make a corresponding change in DAP.

PR #561, opened 2025-08-21
Clarify what we mean by "secure" VDAF execution

### LINKED ISSUE #553: Crypto panel: Section 5.6: "insecure network" versus "secure channel"
> - I am not familiar with the terminology of "simulating a procedure over
> an insecure network", but if RFC readers usually are, this is fine. For
> me, it was confusing to read this instruction in Section 5.6 while also
> demanding the application to instantiate secure channels (end of 5.6).

The point of `run_vdaf()` is to describe the sequence of computations that are carried out by the parties. What we're trying to say here is that, in practice, this sequence of computations is executed in the presence of an adversary as defined in Section 9. 

https://mailarchive.ietf.org/arch/msg/cfrg/Omdhr4rO1pla_nlju2l7OJEGWPM/

cjpatton:
"Simulate [some procedure] over an insecure network" is a confusing concept that we don't unpack very well. All we mean is that `run_vdaf()` shows ideally how the computation is carried out, but in reality messages are exchanged over a network the attacker has control of. It's therefore up to the application to do the stuff described in security considerations.

While at it, make `run_vdaf()` more readable.

Closes #553.


draft-irtf-cfrg-vdaf.md
```diff
@@ -1598,25 +1606,31 @@ def run_vdaf(
...
     agg_result = vdaf.unshard(agg_param, agg_shares,
                               num_measurements)
     return agg_result
 ~~~
 
-The inputs to this algorithm are the verification key, application context,
-aggregation parameter, and a list of measurements. As explained in
-{{daf-execution}}, the secure execution of a VDAF requires the application to
-instantiate secure channels between each of the protocol participants.
+Depending on the VDAF, prpearation and aggregation may be carried out multiple
```
divergentdave, line 1626: nit: typo
```suggestion
Depending on the VDAF, preparation and aggregation may be carried out multiple
```

PR #568, opened 2025-09-09
Fix incorrect reference to gadget polynomials

### LINKED ISSUE #566: Say wire polynomials are constructed interpolating at roots of unity rather than gadgets
Unrelated: I think this whole paragraph should probably be talking about wire polynomials instead of gadget polynomials, since it is the wire polynomials that are directly constructed by polynomial interpolation.

_Originally posted by @divergentdave in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/564#discussion_r2330888634_

divergentdave:
This closes #566. Wire polynomials have not yet been introduced at this point (that comes two sections later), so I just changed this paragraph to talk about polynomial interpolation more generally.


cjpatton (approved):
Perfect, exactly the way I would have fixed this.

PR #576, opened 2025-10-09
Apply improved range proof to Sum and SumVec circuits

### LINKED ISSUE #549: prio: Improve range proof in Sum
Sum gadget requires a range proof that checks that `measurement<MAX_MEAS`, in the [current draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-15#name-prio3sum), this is done using two range proofs to the power-two bit decomposition.

Note that this can be done more efficiently as Ian described here: https://mailarchive.ietf.org/arch/msg/privacy-pass/xK8bi2p08w5k3JjXzoQLUn6YXIo/

divergentdave:
This applies the idea from #549 to Sum and SumVec. In the case of Sum, it simplifies the description and halves the encoded measurement size. In the case of SumVec, there's no performance change, and we get subjectively better utility through more flexibility in parameter choices. Since this is a breaking change to both affine-aggregatable encodings, and the Sum validity circuit, this results in test vector changes.

Closes #549.

### DIFF
--- a/poc/vdaf_poc/flp_bbcggi19.py
+++ b/poc/vdaf_poc/flp_bbcggi19.py
@@ -1067,12 +1067,13 @@ class Sum(Valid[int, int, F]):
         `[0, max_measurement]`. This is accomplished by encoding the
         measurement as a vector of zeroes and ones, such that a
         weighted sum of the "bits" can only be in this range. All but
-        the last of the weights are successive powers of two, as in a
-        binary bit decomposition, and the last weight is chosen such
-        that the sum of all weights is equal to `max_measurement`.
-        With these weights, valid measurements have either one or two
-        possible representations as a vector of 0/1 field elements,
-        and invalid measurements cannot be represented.
+        the last of the weights are successive powers of two, as in
+        the binary bit decomposition, and the last weight is chosen
+        such that the sum of all weights is equal to
+        `max_measurement`. With these weights, valid measurements
+        have either one or two possible representations as a vector
+        of field elements with value zero or one, and invalid
+        measurements cannot be represented.
 
         The validity circuit checks that each entry of the bit vector
         has a value of zero or one.


draft-irtf-cfrg-vdaf.md
```diff
@@ -4004,39 +4004,24 @@ The next variant of Prio3 supports summing of integers in a pre-determined
...
-subtracts the offset from the latter. It then checks if these two values are
-equal. Since both the measurement and the measurement plus `offset` are in the
-same range of `[0, 2**bits)`, this means that the measurement itself is between
-`0` and `max_measurement`.
-
-The circuit uses the polynomial-evaluation gadget `PolyEval` specified in
+The range check is accomplished by encoding the measurement as a vector of field
+elements with value zero or one, such that a weighted sum of the "bits" can only
+be in this range. All but one of the weights are successive powers of two, as in
+a binary bit decomposition, and the last weight is chosen such that the sum of
```
armfazh:
```suggestion
the binary bit decomposition, and the last weight is chosen such that the sum of
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4004,39 +4004,24 @@ The next variant of Prio3 supports summing of integers in a pre-determined
...
-The circuit first checks that each entry of both bit vectors is a one or a
-zero. It then decodes both the measurement and the offset measurement, and
-subtracts the offset from the latter. It then checks if these two values are
-equal. Since both the measurement and the measurement plus `offset` are in the
-same range of `[0, 2**bits)`, this means that the measurement itself is between
-`0` and `max_measurement`.
-
-The circuit uses the polynomial-evaluation gadget `PolyEval` specified in
+The range check is accomplished by encoding the measurement as a vector of field
+elements with value zero or one, such that a weighted sum of the "bits" can only
```
armfazh: suggestion: this is one way to avoid quoting the _bits_ word.

```suggestion
elements with value zero or one, such that a weighted sum of these values can only
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4004,39 +4004,24 @@ The next variant of Prio3 supports summing of integers in a pre-determined
...
+be in this range. All but one of the weights are successive powers of two, as in
+a binary bit decomposition, and the last weight is chosen such that the sum of
+all weights is equal to `max_measurement`. With these weights, valid
+measurements have either one or two possible representations as a vector of 0/1
+field elements, and invalid measurements cannot be represented.
```
armfazh: "a vector of 0/1 field elements" can be consufed with vectors of length zero or one.

```suggestion
measurements have either one or two possible representations as vectors of 
field elements with zero or one values, and invalid measurements cannot be represented.
```

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -902,32 +902,31 @@ def test_vec_set_type_param(self, test_vec: dict[str, Any]) -> list[str]:
...
         """
         # REMOVE ME
-        if 2 ** bits >= field.MODULUS:
-            raise ValueError('bit size exceeds field modulus')
-        if bits <= 0:
-            raise ValueError('invalid bits')
         if length <= 0:
             raise ValueError('invalid length')
         if chunk_length <= 0:
             raise ValueError('invalid chunk_length')
```
armfazh:
```suggestion
            raise ValueError('invalid chunk_length')
        if max_measurement <= 0:
            raise ValueError('invalid max_measurement')            
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -4047,69 +4032,76 @@ class Sum(Valid[int, int, F]):
...
-of positive integers with length equal to the `length` parameter. Each element
-of the measurement is an integer in the range `[0, 2**bits)`. It is RECOMMENDED
-to set `chunk_length` to an integer near the square root of `length * bits`
-(see {{parallel-sum-chunk-length}}).
+parameters: `length`, `max_measurement`, and `chunk_length`. Each measurement is
+a vector of positive integers with length equal to the `length` parameter. Each
+element of the measurement is an integer in the range `[0, max_measurement]`. It
+is RECOMMENDED to set `chunk_length` to an integer near the square root of
+`length * bits`, where `bits` is the number of bits needed to represent
+`max_measurement` (see {{parallel-sum-chunk-length}}).
```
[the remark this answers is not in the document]
divergentdave: I don't want to add this formula, as the linked subsection notes that it could produce proofs up to 50% larger than the optimal size found through a parameter search. I'll add a formula for deriving `bits` here, which should help clarify the next paragraph as well.

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -902,32 +902,33 @@ def test_vec_set_type_param(self, test_vec: dict[str, Any]) -> list[str]:
...
+        if max_measurement <= 0:
+            raise ValueError('invalid max_measurement')
 
         self.field = field
         self.length = length
+        bits = max_measurement.bit_length()
         self.bits = bits
+        self.max_measurement = max_measurement
+        self.last_weight = max_measurement - (2 ** (bits - 1) - 1)
```
cjpatton: nit: It might be nice to give this a name since we reuse it. "mask" may not be right.

```suggestion
        self.last_mask = (2 ** (bits - 1) - 1)
        self.last_weight = max_measurement - self.last_mask
```

draft-irtf-cfrg-vdaf.md
```diff
@@ -6581,6 +6598,9 @@ time to developing definitions and security proofs.
 Thanks to Julia Hesse who provided feedback on behalf of the Crypto Review
 Panel.
 
+Thanks to Ian Goldberg for the bit decomposition technique used in the Sum
+circuit.
```
cjpatton: Not just the sum circuit!
```suggestion
Thanks to Ian Goldberg for the bit decomposition technique used in the range proofs
of the Prio3 variants.
```

divergentdave:
I rebased this on main.

PR #580, opened 2025-10-21
Add Prio3 test with a degree 3 gadget

divergentdave:
I noticed that while we currently have one gadget test that exercises a gadget of degree greater than two in isolation, we don't have any tests using such gadgets in the FLP. This defines a new validity circuit consisting of a degree three gadget, and adds an FLP test using it.


poc/tests/test_vdaf_prio3.py
cjpatton: Does testing this circuit in Prio3 increase test coverage at all? If not, I would probably not add this code.

cjpatton (approved):

Issue #574, opened 2025-09-23
prio: Mantain polynomial always in the Lagrange basis.


armfazh:
It has been observed in the rhizomes paper, Prio executes faster if the polynomials are kept in the Lagrange basis. For example, the FLP Prover sending the gadget polynomial to the verifier in such a representation. 

The changes needed are minor:
 - In the spec: indicate that all polynomial operations must be performed in Lagrange basis. 
 - In the code: implement Lagrange polynomial arithmetic and recalculate test vectors.

A verbose example of the changes needed in librpio-rs is here: https://github.com/armfazh/rhizomes/pull/2

Reference: 
[link](https://eprint.iacr.org/2025/1727) Rhizomes and the Roots of Efficiency, Latincrypt 2025.

armfazh:
The PR for libprio-rs (https://github.com/armfazh/rhizomes/pull/2) is meant to give ideas about the changes needed.

It would be useful if someone (other than me) replicated the results in the Python implementation. Please comment if you prefer I work on the changes for the Python code. I'm ok with either option.

cjpatton:
I would say if anyone is enthusiastic about seeing this through, then go ahead and write a PR.

divergentdave:
I was initially worried about whether this would impact the zero knowledge security of the FLP, since the random wire seed values blind various values later in the protocol, and the Lagrange-to-monomial conversion mixes the wire seed into all polynomial coordinates, but I've read through the paper and reviewed BBCGGI '19, and I'm satisfied this wouldn't affect the security analysis. The prover's message is $\pi=(w, f_1(\omega_0), ..., f_L(\omega_0), c_p)$ and this would change the coordinates of the proof polynomial or gadget polynomial, $c_p$, from one basis to another. The HVZK simulator only needs to simulate the verifier's randomness and the response to the verifier's linear query, which evaluates $f'\_i(r)$, $p'(r)$, and $p'(\omega\_{M-1})$. The linear query that the verifier makes would naturally change due to the change in polynomial basis, but the values produced by the query would be the same. Since the polynomial coordinates $c_p$ are not in the verifier's view at all, the joint distributions are unchanged. Since we exclude cases where r is a root of unity, $f'\_i(r)$ will still be uniformly distributed due to the influence of $f\_i(0)$, chosen by the honest prover.

I think this change is worth considering, the speedups look promising.

cjpatton:
[...]

PR #585, opened 2025-11-06
prio: Perform polynomial arithmetic in the Lagrange basis.

armfazh:
The main change is in the FLP, where the (input/output) wire polynomials must remain in the Lagrange basis. This eliminates conversions to the monomial basis that are performed by NTT invocations.

The Prover must send the gadget_poly in the Lagrange basis, so the Verifier can evaluate it directly in this basis. Algorithms are described in [Faz25](https://ia.cr/2025/1727).

Updates test vectors. As can be seen, the test vectors only change the last part of each share, because the `gadget_poly` is in a different basis.

Partially addresses #574

### DIFF
--- a/poc/vdaf_poc/flp_bbcggi19.py
+++ b/poc/vdaf_poc/flp_bbcggi19.py
@@ -276,6 +276,7 @@ class QueryGadget(Gadget[F]):
         # Recover all the values of the gadget_poly.
         lag = Lagrange(field)
         n = next_power_of_2(len(gadget_poly))
+        gadget_poly = list(gadget_poly)
         lag.extend_values_to_power_of_2(gadget_poly, n)
 
         # Calculate 'size' evaluations of the gadget_poly.


poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -275,11 +273,33 @@ def __init__(
...
+        assert l <= n
+        lag = Lagrange(field)
+        nodes = field.nth_root_powers(n)
+        for _ in range(l, n):
+            # Extending the dimension (one by one) to the closest power of two.
+            # This keeps the degree of polynomial unchanged.
+            next_value = lag.extend_dimension_one(nodes, gadget_poly)
+            gadget_poly += [next_value]
+        # Calculate 'size' evaluations of gadget_poly at the roots of unity.
+        size = next_power_of_2(g.DEGREE*p)
```
divergentdave: I think this is calculating more points than needed in edge cases. Note that the wire polynomials are all polynomials of degree at most (p-1), so the gadget polynomial has a degree of at most g.DEGREE * (p - 1), and therefore the following number of coordinates.
```suggestion
        size = next_power_of_2(g.DEGREE * (p - 1) + 1)
```

divergentdave (approved):
These changes look good, I just have some small comments. We'll need to reflect these changes in the draft next, and update the description of the FLP. I can help out with that next week I think.

poc/tests/test_field.py
```diff
@@ -49,6 +49,43 @@ def run_ntt_field_test(self, cls: type[NttField]) -> None:
         # Test generator.
         self.assertTrue(cls.gen()**cls.GEN_ORDER == cls(1))
 
+        for logN in range(1, 8):
```
cjpatton: Let's follow python naming conventions. Here and below:
```suggestion
        for log_n in range(1, 8):
```

poc/tests/test_field.py
```diff
@@ -49,6 +49,43 @@ def run_ntt_field_test(self, cls: type[NttField]) -> None:
...
+            self.run_nth_roots_test(cls, logN)
+
+        for logN in range(8):
+            self.run_ntt_test(cls, logN)
+            self.run_ntt_star_test(cls, logN)
+
+    def run_nth_roots_test(self, cls: type[NttField], logN: int) -> None:
```
cjpatton: Would you mind adding a comment to explain what's going on in this test? Here and below.

poc/vdaf_poc/field.py
```diff
@@ -166,6 +168,77 @@ class NttField(Field):
...
 
+    @classmethod
+    def nth_root(cls, n: int) -> Self:
+        """Returns the $2^n$-th root of unity."""
+        return cls.gen() ** (cls.GEN_ORDER >> n)
+
+    @classmethod
+    def nth_root_powers(cls, n: int) -> list[Self]:
+        """Returns the powers of the $n$-th root of unity."""
+        logN = assert_power_of_2(n)
```
cjpatton:
```suggestion
        log_n = assert_power_of_2(n)
```

poc/vdaf_poc/flp_bbcggi19.py
divergentdave: That still works with the following change. Historically, there were a few places where we had overly conservative formulas for polynomial degrees like this, so I think we've just missed this one.

```patch
diff --git a/src/flp.rs b/src/flp.rs
index 7c85614..1153f76 100644
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -717,7 +717,7 @@ impl<F: NttFriendlyFieldElement> QueryShimGadget<F> {
         let gadget_degree = inner.degree();
         let gadget_arity = inner.arity();
         let m = (1 + inner.calls()).next_power_of_two();
-        let p = m * gadget_degree;
+        let p = (m - 1) * gadget_degree + 1;
 
         // Each call to this gadget records the values at which intermediate proof polynomials were
         // interpolated. The first point was a random value chosen by the prover and transmitted in
```

In this Python code, `p = next_power_of_2(1 + g_calls)` is the number of coordinates that we use to define the wire polynomials, so the wire polynomials are of degree p-1 or less, and composition means the gadget polynomial will have degree `g.DEGREE * (p - 1)`. Thus, we need at least `g.DEGREE * (p - 1) + 1` coordinates for the gadget polynomial. Note that this also matches the calculation of `gadget_poly_len` in `prove()`. Here we round up to the next power of two, and later rely on that when getting roots of unity, but I'm not sure if that rounding up is strictly necessary.

divergentdave: I did some further testing and experimentation of this in draft PR #589. One thing I discuss there is that the calculation of `step` will compensate for the extra basis extension in cases where `p` is larger than necessary. So, that explains why verification still works either way.

armfazh: The calculation is correct. As also noted in the [the paper](https://eprint.iacr.org/2019/188.pdf) the degree is at most `M*deg(G)`, where `M=p-1`. So the length of the coordinates vector is `g.degree*(p-1) + 1`.

image

poc/tests/test_field.py
```diff
@@ -49,6 +49,43 @@ def run_ntt_field_test(self, cls: type[NttField]) -> None:
...
+        got = cls.ntt(p_mon, N)
+
+        root = cls.nth_root(logN)
+        want = [poly_eval(cls, p_mon, root**i) for i in range(N)]
+        self.assertEqual(
+            got, want,
+            f"logN: {logN} p_mon: {p_mon}"
+        )
+
+    def run_ntt_star_test(self, cls: type[NttField], logN: int) -> None:
```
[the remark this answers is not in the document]
armfazh: done

poc/vdaf_poc/field.py
armfazh: done

poc/vdaf_poc/common.py
```diff
@@ -23,12 +24,28 @@ def __neg__(self) -> Self:
...
     assert n > 0
     return 1 << (int(n - 1).bit_length())
 
 
+def bitrev(k: int, n: int) -> int:
+    """Return the integer obtained by reversing the k-bit representation of n."""
+    maxnum = 1 << k
+    assert 0 <= n < maxnum, f"{n} out of range [0, {maxnum})."
+    rev = bin(n)[2:][::-1]
+    padded = rev + '0' * (k-len(rev))
```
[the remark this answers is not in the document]
armfazh: either way works. Here reversing first, then padding.

poc/vdaf_poc/__init__.py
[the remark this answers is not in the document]
armfazh: it declares that `vdaf_poc` is a module.
Similar to `go.mod` in Go.

poc/tests/test_field.py
```diff
@@ -49,6 +49,46 @@ def run_ntt_field_test(self, cls: type[NttField]) -> None:
...
+        got = cls.ntt(p_mon, n)
+
+        root = cls.nth_root(log_n)
+        want = [poly_eval(cls, p_mon, root**i) for i in range(n)]
+        self.assertEqual(got, want, f"log_n: {log_n} p_mon: {p_mon}")
+
+    def run_ntt_star_test(self, cls: type[NttField], log_n: int) -> None:
+        # Tests that NTT*(P) is the same as evaluating the polynomial P
+        # on the powers of an n-th root of unity times an 2n-th root of
+        # unity.
```
cjpatton: Thanks for the explanation. Let's try to find a more meaningful name for this algorithm than `ntt_star`. How would you explain the purpose of this function?

poc/vdaf_poc/field.py
```diff
@@ -166,6 +168,79 @@ class NttField(Field):
...
+        return [root**i for i in range(n)]
+
+    @classmethod
+    def ntt(cls, p: list[Self], n: int) -> list[Self]:
+        r"""
+        Number Theoretic Transform (NTT) over a field.
+
+        Given a polynomial P in monomial basis, returns
+        $\\{ P({w_n}^i) : 0 \\leq i < n \\}$
+        where $w_n$ is the $n$-th root of unity.
```
cjpatton: We're going to end up copy-pasting this into the draft. Let's try to make this readable without the latex formatting.

poc/vdaf_poc/field.py
```diff
@@ -363,32 +416,82 @@ def poly_interp(field: type[F], xs: list[F], ys: list[F]) -> list[F]:
...
+    def poly_eval(self, nodes: list[F], values: list[F], x: F) -> F:
+        """Evaluate a polynomial in the Lagrange basis at x."""
+        return self.poly_eval_batched(nodes, [values], x).pop()
+
+    def poly_eval_batched(
+            self, nodes: list[F], polynomials: list[list[F]], x: F,
+    ) -> list[F]:
+        """Evaluate polynomials in the Lagrange basis at x.
+
+        See Alg. 7 of [Faz25](https://ia.cr/2025/1727).
```
cjpatton: Define "nodes"

poc/vdaf_poc/field.py
```diff
@@ -363,32 +416,82 @@ def poly_interp(field: type[F], xs: list[F], ys: list[F]) -> list[F]:
...
+                if i < len(p):
+                    u[j] += t*p[i]
+
+        sgn = field(-1)**(n-1)
+        inv_n = field(n).inv()
+        return [sgn*inv_n*u_j for u_j in u]
+
+    def extend_dimension_one(self, nodes: list[F], values: list[F]) -> F:
+        """
+        Extends the dimension of a polynomial in the Lagrange basis from k to k+1.
```
cjpatton: Define "dimension"

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -275,11 +273,33 @@ def __init__(
...
+            # This keeps the degree of polynomial unchanged.
+            next_value = lag.extend_dimension_one(nodes, gadget_poly)
+            gadget_poly += [next_value]
+        # Calculate 'size' evaluations of gadget_poly at the roots of unity.
+        size = next_power_of_2(g.DEGREE * (p - 1) + 1)
+        while len(gadget_poly) < size:
+            gadget_poly = lag.extend_dimension_double(gadget_poly)
+        self.poly = gadget_poly
+
+        # Calculate the step.
```
cjpatton: What do you mean by "step"

cjpatton (commented):
Looking good. I'm still looking for some intuition of what `ntt_star` is doing. Plus, we should think of a name for this that reflects its role in the FLP.

poc/vdaf_poc/field.py
armfazh: refactored.

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -525,15 +550,16 @@ def eval_poly(self,
...
-        for i in range(1, len(p)):
-            for j in range(len(x)):
-                out[j] += p[i] * x[j]
-            x = poly_mul(field, x, inp_poly[0])
-        return poly_strip(field, out)
+        l = len(inp_poly[0])
+        assert_power_of_2(l)
+
+        # Convert the input polynomial I from Lagrange to monomial.
+        I_mon_l = field.inv_ntt(inp_poly[0], l)
```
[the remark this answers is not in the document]
armfazh: fixed.

poc/vdaf_poc/field.py
```diff
@@ -363,32 +399,103 @@ def poly_interp(field: type[F], xs: list[F], ys: list[F]) -> list[F]:
...
+        than or equal to the initial length.
+
+        See Eq. (3.2.1) of [Faz25](https://ia.cr/2025/1727).
+        """
+        n = next_power_of_2(len(p))
+        assert len(p) <= n
+        field = cast(type[F], self.field)
+        x = cast(list[F], self.field.nth_root_powers(n))
+
+        def w_inv(m: int, i: int) -> F:
```
cjpatton: What is `w` here refer to?

poc/vdaf_poc/flp_bbcggi19.py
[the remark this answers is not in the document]
armfazh: I'd rather defer to @divergentdave to add comments on this.

divergentdave (approved):

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -275,11 +273,27 @@ def __init__(
             wire[0] = s  # set the wire seed
             self.wires.append(wire)
 
+        # Recover all the values of the gadget_poly.
+        lag = Lagrange(field)
+        n = next_power_of_2(len(gadget_poly))
+        lag.extend_values_to_power_of_2(gadget_poly, n)
```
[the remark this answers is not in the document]
armfazh: resolved.

divergentdave (approved):

PR #589, opened 2025-11-14
Experiment: Combine higher degree gadgets with gadget polynomials in the Lagrange basis

divergentdave:
This rebases #588 on top of #585, and tries out some other changes. The main outcome I found is that everything still works with a wide variety of gadget degrees and number of gadget calls.

I also tried changing the calculation of `size` in the `QueryGadget` constructor as described in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/585#discussion_r2500947483, and verification still worked. If the value of `size` is calculated to be larger than it needs to be, then it will be twice as large as necessary. Later in the constructor, we compute `step` to be the ratio of `size`, the gadget polynomial length, over `p`, the wire polynomial length. If `size` is double what it needs to be, then `step` would correspondingly be doubled. When we extend the gadget polynomial with `extend_dimension_double()`, all the elements of the input become the even-indexed elements of the output. Thus, `QueryGadget.eval()` produces the same values, because the stride of `self.poly[self.k * self.step]` correspondingly doubles, and the same values are returned whether `size` is higher than necessary or not. The later computation of g(r) during the query algorithm trivially produces the same value, since we are dealing with the same polynomial, just basis-extended further.

I also noticed that verification slowed down significantly when processing gadgets with very high degrees. Most of the time was in `extend_dimension_one()`. This is not of a practical concern, because we only care about gadgets of degree two, in which case we'd only need to call `extend_dimension_one()` once. If, in the future, we want to make higher degree gadgets more performant, I think we could change some implementation choices without making any changes to the protocol. Note that we only extend the basis of the gadget polynomial until its length is the next power of two because our later computation of g(r) uses `poly_eval_batched()`. This requires the basis length to be a power of two. We could probably change either how we extend the basis or how we evaluate the polynomial to use algorithms that are more concretely efficient for such parameter choices. I recently learned about "bidirectional FFT" algorithms, which take as input a prefix of the evaluations of a polynomial and a suffix of the monomial coefficients. Future work may be able to apply such algorithms to do basis extension in one O(n log n) step. But again, we only care about degree two gadgets in practice, so our implementation is fine, and the protocol still allows for efficient implementations with any parameter choice.

### DIFF
--- a/poc/tests/test_test_vectors.py
+++ b/poc/tests/test_test_vectors.py
@@ -3,6 +3,7 @@ import os.path
 import unittest
 from typing import Generic, Optional, TypeVar, cast
 
+from tests.test_vdaf_prio3 import Prio3HigherDegree
 from vdaf_poc.field import Field64, NttField
 from vdaf_poc.test_utils import VdafTestVectorDict
 from vdaf_poc.vdaf import Vdaf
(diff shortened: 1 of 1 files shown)


divergentdave:
This came from https://eprint.iacr.org/2024/2010, drawing on a line of work on "truncated FFT". In that setting it's applied to an additive FFT, so it would take some additional work to apply it to fields of characteristic greater than two.

PR #599, opened 2025-12-12
Refactor encoding/decoding of range checked ints

### LINKED ISSUE #587: Replace encode_into_bit_vec() and decode_from_bit_vec() with helpers for range checked integers
Following #576 and #581, `encode_into_bit_vec()` and `decode_from_bit_vec()` will only be used with range checked integers, on all but the last field elements for each encoded integer. We could replace these with new helper functions that encapsulate the whole range checked integer encoding/decoding process, simplifying circuit definitions.

divergentdave:
This closes #587.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -4042,8 +4042,9 @@ def encode_range_checked_int(
         value: int,
         max_measurement: int) -> list[F]:
     """
-    Encode an integer into multiple field elements in a way that
-    allows for efficient range proofs.
+    Encode an integer into multiple field elements, representing a
+    weighted sum of zeros and ones, to allow for efficient range
+    proofs.
 
     Pre-conditions:
 
@@ -4060,9 +4061,9 @@ def encode_range_checked_int(
     rest_all_ones_value = 2 ** (bits - 1) - 1
     last_weight = max_measurement - rest_all_ones_value
 
-    # Implementation note: this conditional should be replaced
-    # with constant time operations in practice in order to
-    # reduce leakage via timing side channels.
+    # Implementation note: this conditional should be replaced with
+    # constant time operations in order to reduce leakage via timing
+    # side channels.
     if value <= rest_all_ones_value:
         rest = value
         last_elem = field(0)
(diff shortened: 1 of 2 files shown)


poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -1147,3 +1109,65 @@ def decode(self, output: list[F], _num_measurements: int) -> int:
...
         return ['max_measurement']
+
+
+def encode_range_checked_int(
+        field: type[F],
+        value: int,
+        max_measurement: int) -> list[F]:
+    """
+    Encode an integer into multiple field elements in a way that
+    allows for efficient range proofs.
```
cjpatton: I like the way you expressed this elsewhere:
```suggestion
    Encode an integer as a weighted sum of zeros and ones to allow
    for efficient range proofs.
```

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -1147,3 +1109,65 @@ def decode(self, output: list[F], _num_measurements: int) -> int:
...
+
+    Pre-conditions:
+
+        - `value >= 0`
+        - `max_measurement > 0`
```
cjpatton: nit: is `max_measurement == 0` a valid value? It's obviously useless, but I just wanted to check my understanding.

poc/vdaf_poc/flp_bbcggi19.py
```diff
@@ -1147,3 +1109,65 @@ def decode(self, output: list[F], _num_measurements: int) -> int:
...
+    if value > max_measurement:
+        raise ValueError("measurement is too large")
+
+    bits = max_measurement.bit_length()
+    rest_all_ones_value = 2 ** (bits - 1) - 1
+    last_weight = max_measurement - rest_all_ones_value
+
+    # Implementation note: this conditional should be replaced
+    # with constant time operations in practice in order to
+    # reduce leakage via timing side channels.
```
cjpatton:
```suggestion
    # with constant time operations in order to
    # reduce leakage via timing side channels.
```

draft-irtf-cfrg-vdaf.md
cjpatton: We should add a backwards reference to the section that defines `encode_range_checked_int()` and `decode_range_checked_int()`. Here and below.

cjpatton (commented):
Nice.

poc/vdaf_poc/flp_bbcggi19.py
divergentdave: The SumVec circuit constructor rejects a `max_measurement` of zero, for example. The construction doesn't make much sense with a maximum measurement of zero, because that can be represented with zero bits, but we want to reserve one of those zero bits for the last weight.

draft-irtf-cfrg-vdaf.md
divergentdave: I'll add references a couple paragraphs up, where we're talking about the encoding procedure.

cjpatton (approved):

PR #611, opened 2026-03-19
Clarify description of FLP consistency checks

### LINKED ISSUE #610: `Flp.query` starts but never finishes
In [7.3.4](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-18#section-7.3.4) we have the specification of `Flp.query`:

```python
def query(self,
          meas: list[F],
          proof: list[F],
          query_rand: list[F],
          joint_rand: list[F],
          num_shares: int) -> list[F]:
    # Evaluate the validity circuit, recording the value of each
    # input wire for each evaluation of each gadget. Use the gadget
    # polynomials encoded by `proof` to compute the gadget outputs.
    valid = QueryGadget.wrap(self.valid, proof)
    out = valid.eval(meas, joint_rand, num_shares)

    # Reduce the output.
    if self.valid.EVAL_OUTPUT_LEN > 1:
        (rand, query_rand) = front(
            self.valid.EVAL_OUTPUT_LEN,
            query_rand,
        )
        v = self.field(0)
        for (r, out_elem) in zip(rand, out):
            v += r * out_elem
    else:
        [v] = out

    # Construct the verifier message, which consists of the reduced
    # circuit output and each gadget test.
    lag = Lagrange(self.field)
    verifier = [v]
    for (g, t) in zip(cast(list[QueryGadget[F]], valid.GADGETS),
                      query_rand):
        p = len(g.wires[0])

        # Abort if `t` is one of the inputs used to compute the wire
        # polynomials so that the verifier message doesn't leak the
        # gadget output. It suffices to check if `t` is a root of
        # unity, which implies it is a power of `alpha`.
        if t ** p == self.field(1):
            raise ValueError('test point is a root of unity')

        # To test the gadget, we re-compute the wire polynomials and
        # check for consistency with the gadget polynomial provided
        # by the prover. To start, evaluate the gadget polynomial          <-- start without a finish?
        # and each of the wire polynomials at the random point `t`.
        wire_checks = lag.poly_eval_batched(g.wires[:g.ARITY], t)
        gadget_check = lag.poly_eval(g.poly, t)

        verifier += wire_checks
        verifier.append(gadget_check)

    return verifier
```

The second to last line in the comment that I marked with `<--` discusses starting a consistency check, but the reader is left hanging. I guess this is finished in [7.3.5](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-18#section-7.3.5) but the connection could be more clear:

> To finish each gadget test, evaluate the gadget on the wire checks: [...]

divergentdave:
This revises the description of how the `query()` and `decide()` parts of the FLP work together, both in the spec text and in reference implementation comments. This gets rid of the terms "wire check" and "gadget check", which were only used here and not properly defined.

Closes #610.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -4059,7 +4059,7 @@ To finish each gadget test, evaluate the gadget on the evaluations of the wire
 polynomials parsed from the verifier message: if the
 encoded measurement and joint randomness used to generate the proof are the
 same as the measurement (share) and joint randomness used to verify the proof,
-then the output of the gadget will be equal to the evaluation of the wire
+then the output of the gadget will be equal to the evaluation of the gadget
 polynomial in the verifier message; otherwise, the
 output will not equal the gadget polynomial evaluation with high probability.
 
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -4022,8 +4022,13 @@ def query(self,
 
         # To test the gadget, we re-compute the wire polynomials and
         # check for consistency with the gadget polynomial provided
-        # by the prover. To start, evaluate the gadget polynomial
-        # and each of the wire polynomials at the random point `t`.
+        # by the prover. Here, we evaluate secret shares of the
+        # gadget polynomial and secret shares of each of the wire
+        # polynomials at the random point `t`. These secret shares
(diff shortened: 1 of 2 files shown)


cjpatton (approved):
Nice!

divergentdave (commented):
I spotted one more correction to make

PR #613, opened 2026-06-24
Clarify how we ensure useful query randomness

cjpatton:
This addresses Julia Hesse's feedback here in
<https://mailarchive.ietf.org/arch/msg/cfrg/vEOC8Dmlzp-vMaO-Nql5T_rAVQ4/>.


draft-irtf-cfrg-vdaf.md
```diff
@@ -6061,11 +6061,11 @@ scheme would not allow key rotation over the lifetime of a task.
...
-measurement, the Client MUST generate this nonce using a CSPRNG. This is
-required in order to leverage security analysis for the privacy definition of
-{{DPRS23}}, which assumes the nonce is chosen at random prior to generating the
-report. Uniqueness of the nonce is not sufficient because the verification key
-is controlled by the attacker.
+measurement, the Client MUST generate this nonce using a CSPRNG. Assuming the
+verification key is fixed ({the-verification-key}), this ensures any malicious
+Aggregators cannot control the randomness derived for the verification process.
+Uniqueness of the nonce is not sufficient because the verification key is
+controlled by the attacker {{DPRS23}}.
```
divergentdave: We should clarify which attacker we're talking about here. The existing text mentioned this is required for privacy, but we no longer make explicit mention of that.

divergentdave (approved):

PR #616, opened 2026-07-29
Add text required for IRTF stream documents

divergentdave:
This adds a couple required sentences by request.


cjpatton (approved):
I'm pretty sure these would be added eventually by the RFC editor

PR #617, opened 2026-07-29
Fix idnits warnings

divergentdave:
This fixes relevant warnings from https://author-tools.ietf.org/api/idnits?url=https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-20.txt. I reproduced these checks locally by checking out the v2 branch of https://github.com/ietf-tools/idnits/, running `TEXT_PAGINATION=true make versioned/draft-irtf-cfrg-vdaf-21.txt` in this repository, and running either `/path/to/idnits /path/to/versioned/draft-irtf-cfrg-vdaf-21.txt` or `/path/to/idnits --verbose /path/to/versioned/draft-irtf-cfrg-vdaf-21.txt`.

* Non-ASCII characters: These arise in the author name and the title of referenced documents. We shouldn't make any changes.
* Long lines: Fixed in the middle two commits. I think our Makefile stopped warning on long lines after an update, so some of these crept back in with recent changes.
* Non-RFC2606-compliant FQDNs: These are false positives from code expressions like `self.flp.decode`.
* Lowercase 'not' together with uppercase 'MUST': Fixed.
* Document date in the past: This will get fixed when we publish the next version.
* Code comments: Added '&lt;CODE BEGINS&gt;' and '&lt;CODE ENDS&gt;' lines.
* Missing references: These are false positives due to the use of square brackets in code expressions or TLS presentation language types.
* Outdated references: The draft-ietf-ppm-dap informative reference will be automatically updated when we publish the next version. The mention of draft-irtf-cfrg-cryptography-specification-02 is in the change log, and will be deleted before publishing. The mention of draft-irtf-cfrg-vdaf-15 is in the title of a mailing list message, and should not be changed.

P.S. I just realized that the line numbers `idnits` references only make sense when used with the `--showtext` output, since they don't count header and footer lines in the input.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -4645,7 +4645,8 @@ class MultihotCountVec(Valid[list[bool], list[int], F]):
 
         # Make sure `length` and `max_weight` don't overflow the
         # field modulus. Otherwise we may not correctly compute the
-        # sum of measurement vector entries during circuit evaluation.
+        # sum of measurement vector entries during circuit
+        # evaluation.
         if self.field.MODULUS <= length:
             raise ValueError('length is too large for the '
                              'current field size')
@@ -6432,7 +6433,8 @@ class PolyEval(Gadget[F]):
         inp_poly_len = len(inp_poly[0])
         assert_power_of_2(inp_poly_len)
 
-        # Convert the input polynomial from Lagrange to monomial basis.
+        # Convert the input polynomial from Lagrange to monomial
+        # basis.
         inp_mon = field.inv_ntt(inp_poly[0], inp_poly_len)
         # Obtain n evaluations of the input polynomial I.
         inp_lag = field.ntt(inp_mon, self.n)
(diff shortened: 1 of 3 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -5409,14 +5518,18 @@ Elements of the inner field are encoded in little-endian byte order (as defined
 in {{field}}) and are represented as follows:
 
 ~~~ tls-presentation
+<CODE BEGINS>
 opaque Poplar1FieldInner[Fi];
+<CODE ENDS>
```
cjpatton, line 5523: What are &lt;CODE BEGINS&gt;..&lt;CODE ENDS&gt; used for? I'm wondering if we need this for pseudocode verbatim blocks or also ones for message specs.

cjpatton (approved):

draft-irtf-cfrg-vdaf.md
divergentdave: This is to fix warnings related to code comments. Version three's warning message is "Found something which looks like a code comment -- if you have code sections in the document, please surround them with '&lt;CODE BEGINS&gt;' and '&lt;CODE ENDS&gt;' lines." We use comments in both the Python and TLS presentation language embedded in the document, so I added those lines around all Python and TLS presentation language for consistency. I left diagrams and mathematical equations alone.

PR #621, opened 2026-08-14
Address preliminary shepherd comments

cjpatton:
1. Avoid framing VDAF as an alternative to DP in the intro
2. "NOT RECOMMENDED" is not an appropriate way to discourage use of DAFs in applications that would otherwise be insecure.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -310,34 +310,46 @@ fraction of people have experienced a given disease. Thus systems that provide
 aggregate statistics while protecting individual measurements can deliver the
 value of the measurements while protecting users' privacy.
 
-This problem is often formulated in terms of differential privacy (DP)
-{{Dwo06}}. Roughly speaking, a data aggregation system that is differentially
-private ensures that the degree to which any individual measurement influences
-the value of the aggregate result can be precisely controlled. For example, in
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -310,34 +310,46 @@ fraction of people have experienced a given disease. Thus systems that provide
...
+
+MPC and DP are therefore complementary, and an end-to-end private measurement
+system typically requires both. MPC bounds what the parties carrying out the
+computation learn, but not what the aggregate result reveals; DP bounds what the
+aggregate result reveals, but does not by itself prevent the party performing
+the aggregation from seeing each measurement. Indeed, a client that adds noise
+locally must add enough of it to protect itself from the data collector, which
+costs significantly more utility than is needed to protect the aggregate.
+Composing DP with MPC {{MPRV09}} allows the parties to calibrate the added
+noise more precisely.
```
divergentdave: The difference isn't in the precision of noise calibration, but the magnitude of the noise (and thus the precision of the aggregate result).
```suggestion
Composing DP with MPC {{MPRV09}} allows the parties to achieve similar privacy
guarantees with less noise, which improves the utility of the data analysis.
```

divergentdave (approved):

PR #624, opened 2026-08-15
Revise consensus-level statement

divergentdave:
This adopts a suggestion to make the consensus-level statement more detailed.

I also added a link to the CFRG Crypto Review Panel feedback reference in a separate commit, I'm not sure if we should add this or skip it. FWIW, note that once we remove the change log, all existing uses of this reference will be deleted.

### DIFF
--- a/draft-irtf-cfrg-vdaf.md
+++ b/draft-irtf-cfrg-vdaf.md
@@ -461,7 +461,10 @@ enumerates security considerations for DAFs and VDAFs in general and the
 Prio3 and Poplar1 constructions in particular.
 
 This document represents the consensus of the Crypto Forum Research Group
-(CFRG).
+(CFRG). It has received review from CFRG participants, including review during
+Research Group Last Call, and an additional review by the CFRG Crypto Review
+Panel. It is also informed by experience from multiple independent
+implementations
 
 ## Change Log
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-vdaf.md
```diff
@@ -461,7 +461,10 @@ enumerates security considerations for DAFs and VDAFs in general and the
 Prio3 and Poplar1 constructions in particular.
 
 This document represents the consensus of the Crypto Forum Research Group
-(CFRG).
+(CFRG). It has received review from CFRG participants, including review during
+Research Group Last Call, and an additional review by the CFRG Crypto Review
+Panel ({{PANEL-FEEDBACK}}). It is also informed by experience from multiple
+independent implementations
```
tgeoghegan:
```suggestion
independent implementations.
```

We might say "deployments", too, which I think is stronger than "implementation".

cjpatton (approved):

draft-irtf-cfrg-vdaf.md
divergentdave: I think multiple independent implementations might be more important than multiple independent deployments in this context. If we had multiple deployments of a single implementation of VDAF, that would not tell me as much about the quality and maturity of the specification.

